PR #2, opened 2022-09-12
Add Parser skeleton.

fernape:
Also, add definition for the `Part` interface to be implemented by the parts that will make the AST.

The parser contains two indexes to keep track of the current position and the last parsed position.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -1,0 +1,45 @@
+package parse
+
+// QueryPart defines a simple interface for all the different parts that
+// compose a ParsedExpr
+type QueryPart interface {
+	String() string         // print for debug
+	ToSql() (string, error) // print to sql format
+}
+
+// Parser defines the SDL parser
+type Parser struct {
+	input         string // statement to be parsed
+	lastParsedPos int    // position of the last parsed chunk
+	pos           int    // current position of the parser
+}
+
+// NewParser creates a new parser
+func NewParser() *Parser {
+	return &Parser{}
+}
+
+// init initializes the parser
+func (p *Parser) init(input string) {
+	p.input = input
+	p.lastParsedPos = 0
+	p.pos = 0
+}
+
+func (p *Parser) Parse(input string) (*ParsedExpr, error) {
+	p.init(input)
+	return nil, nil
+}
+
+// ParsedExpr represents a parsed expression.
+// It has a representation of the original SQL statement in terms of QueryParts
+// A SQL statement like this:
+//
+// Select p.* as &Person.* from person where p.name = $Boss.Name
+//
+// would be represented as:
+//
+// [stringPart outputPart stringPart inputPart]
+type ParsedExpr struct {
+	queryParts []QueryPart
+}
--- a/internal/parse/parser_test.go
+++ b/internal/parse/parser_test.go
@@ -1,0 +1,14 @@
+package parse
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInit(t *testing.T) {
+	p := NewParser()
+	expr, err := p.Parse("select foo from bar")
+	assert.Equal(t, nil, err)
+	assert.Equal(t, (*ParsedExpr)(nil), expr)
+}


internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
+package parser
+
+// Parser defines the SDL parser
+type Parser struct {
+	parts   []Part // Array of parts. Our simple AST
```
Aflynn50: The name 'parts' could be clearer, perhaps something like queryPart since it will often appear without it being super obvious its part of the Parser

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
...
+	input   string // statement to be parsed
+	parsed  int    // position of the last parsed chunk
+	skipped int    // current position of the parser
```
Aflynn50: Could this not be 'currentPos' instead of skipped since its the current position of the parser?

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
...
+	p.parsed = 0
+	p.skipped = 0
+}
+
+func (p *Parser) Parse(input string) (*ParsedExpr, error) {
+	p.init(input)
+	return nil, nil
+}
+
+// Part defines a simple interface for all the different parts that
```
Aflynn50: Again this could be renamed to QueryPart or something similar to make it a little more specific

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ParsedExpr represents a parsed expression.
+// It has a representation of the original SQL statement in terms of Parts
+// A SQL statement like this:
+//
+// Select p.* as &Person.* from person where p.name = $Boss.Name
+//
+// would be represented as:
+//
+// [stringPart outputPart stringPart inputPart]
```
Aflynn50: These implementations of the Part interface don't exist yet, but I guess it's still pretty clear

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
...
+type ParsedExpr struct {
+	parts []Part
```
Aflynn50: It's a bit confusing having a field with the same name as the Parser field 'parts'. This could be renamed to parsedQueryParts maybe, since that is what it contains.

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
+package parser
+
+// Parser defines the SDL parser
+type Parser struct {
+	parts   []Part // Array of parts. Our simple AST
+	input   string // statement to be parsed
+	parsed  int    // position of the last parsed chunk
```
Aflynn50: This name could also be expanded to be more descriptive

Aflynn50 (commented):
A lot of the design decisions made here are dependent on the future implementation of the parser and AST, they should make more sense once those have been added as well (e.g. 'parts' being a slice instead of a tree).

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
+package parser
```
manadart: Nit: I like "parse" for the package. There's a tendency towards using verb-er for everything.

internal/parser/parser.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (p *Parser) Parse(input string) (*ParsedExpr, error) {
+	p.init(input)
+	return nil, nil
+}
+
+// Part defines a simple interface for all the different parts that
+// compose a ParsedExpr
+type Part interface {
```
manadart: I'd declare this at the top ahead of it's first usage in `Parser`

internal/parser/parser.go
fernape: QueryPart is clearer, thanks.

internal/parser/parser.go
fernape: Changed!

internal/parser/parser.go
fernape: Yes. And names will probably change. `stringPart` might lead to confusion as in reality it has nothing to do with strings. It is more like a `verbatimPart`, `passthroughPart`, etc.

internal/parser/parser.go
fernape: I've removed the attribute from the parser. It was there in the experiment just to keep track of the parts and then it was copied to the returned `ParsedExpression`. It definitely does not belong in the `Parser`.

internal/parser/parser.go
fernape: Done, thanks!

internal/parser/parser.go
fernape: Done!

internal/parser/parser.go
fernape: This one is a bit more difficult to explain. We can of course change `parsed` and `skipped` to something else, but we should choose something that truly reflects the intention. `parsed`  marks the position up to which we have recognized the input stream and generated `QueryParts` for. `skipped` on the other hand represents the temporary cursor we use to [advance through the input and see if we recognize something else](https://github.com/fernape/sqlairexperiment/blob/main/README.md#parsing).

internal/parser/parser.go
fernape: `lastParsedPos` should be a better name.

internal/parser/parser.go
fernape: Changed this to `pos` as is very idiomatic and we keep it short.

manadart (approved):
Good start.

We need to set up some workflows to run tests for us automatically and gate landing, maybe a merge bot too.

Let's institute a policy now of using merge commits, not squash/rebase. Before merging this one, please squash it though.

PR #3, opened 2022-09-15
Add parser basic functions.

fernape:
Basic parser functions to be used as building blocks for higher level functionality.  Most of the functions return a boolean specifying if they were able to perform the action.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -1,5 +1,7 @@
 package parse
 
+import "strings"
+
 // QueryPart defines a simple interface for all the different parts that
 // compose a ParsedExpr
 type QueryPart interface {
@@ -42,4 +44,67 @@ func (p *Parser) Parse(input string) (*ParsedExpr, error) {
 // [stringPart outputPart stringPart inputPart]
 type ParsedExpr struct {
 	queryParts []QueryPart
+}
+
+// peekByte returns true if the current byte
+// equals the one passed as parameter.
+func (p *Parser) peekByte(b byte) bool {
+	return p.pos < len(p.input) && p.input[p.pos] == b
+}
+
+// skipByte jumps over the current byte if it matches
+// the byte passed as a parameter. Returns true in that case, false otherwise.
+func (p *Parser) skipByte(b byte) bool {
+	if p.pos < len(p.input) && p.input[p.pos] == b {
+		p.pos++
+		return true
+	}
+	return false
+}
+
+// skipByteFind advances the parser until it finds a byte that matches the one
+// passed as parameter and then jumps over it. In that case returns true. If the
+// end of the string is reached and no matching byte was found, it returns
+// false.
+func (p *Parser) skipByteFind(b byte) bool {
+	for i := p.pos; i < len(p.input); i++ {
+		if p.input[i] == b {
+			p.pos = i + 1
+			return true
+		}
+	}
+	return false
+}
+
+// skipSpaces advances the parser jumping over consecutive spaces. It stops when
+// finding a non-space character. Returns true if the parser position was
+// actually changed, false otherwise.
+func (p *Parser) skipSpaces() bool {
+	mark := p.pos
+	for p.pos < len(p.input) {
+		if p.input[p.pos] != ' ' {
+			break
+		}
+		p.pos++
+	}
+	return p.pos != mark
+}
+
+// skipString advances the parser and jumps over the string passed as parameter.
+// In that case returns true, false otherwise.
+// This function is case insensitive.
+func (p *Parser) skipString(s string) bool {
+	if p.pos+len(s) <= len(p.input) &&
+		strings.EqualFold(p.input[p.pos:p.pos+len(s)], s) {
+		p.pos += len(s)
+		return true
+	}
+	return false
+}
+
+// isNameByte returns true if the byte passed as parameter is considered to be
+// one that can be part of a name. It returns false otherwise
+func isNameByte(c byte) bool {
+	return 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' ||
+		'0' <= c && c <= '9' || c == '_'
 }
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -101,10 +101,3 @@ func (p *Parser) skipString(s string) bool {
 	}
 	return false
 }
-
-// isNameByte returns true if the byte passed as parameter is considered to be
-// one that can be part of a name. It returns false otherwise
-func isNameByte(c byte) bool {
-	return 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' ||
-		'0' <= c && c <= '9' || c == '_'
-}
--- a/internal/parse/parser_test.go
+++ b/internal/parse/parser_test.go
@@ -7,28 +7,58 @@ import (
 )
 
 type parseHelperTest struct {
-	bytef   func(byte) bool
-	stringf func(string) bool
-	result  bool
-	input   string
-	data    string
+	bytef    func(byte) bool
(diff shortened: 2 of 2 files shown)


internal/parse/parser.go
```diff
@@ -43,3 +45,66 @@ func (p *Parser) Parse(input string) (*ParsedExpr, error) {
...
+func (p *Parser) skipString(s string) bool {
+	if p.pos+len(s) <= len(p.input) &&
+		strings.EqualFold(p.input[p.pos:p.pos+len(s)], s) {
+		p.pos += len(s)
+		return true
+	}
+	return false
+}
+
+// isNameByte returns true if the byte passed as parameter is considered to be
```
Aflynn50: Does this match the full namespace of SQL column names (and is it meant to?)? This doc implies it could also be something of the form \`weirdname*$^%\` https://docs.microsoft.com/en-us/sql/odbc/microsoft/column-name-limitations?view=sql-server-ver16

fernape: Good point.
`isNameByte` is used (in the experiment) from `parseIdentifier` which is used to recognize things like `foo.bar` in things that come before the `AS` keyword *and* things that come after it.
Apparently SQLite [deviates a bit](https://www.sqlite.org/draft/tokenreq.html) from the standard too. Sybase IQ supports things like the pipe (`|`) in column names, postgres has its own rules, etc.
It's going to be difficult to be DB-agnostic here. Maybe we can allow for everything to be a name byte except for _blank separators_ (spaces, tabs, newlines...).

Aflynn50: That sounds reasonable, it'd probably be best to specify specifically what certainly isn't a name rather than what is. As long as the parser doesn't rely on the function too heavily

niemeyer: > it'd probably be best to specify specifically what certainly isn't a name

That creates double negatives which are harder to work with ("is this not-not-part-of-a-name?").

I think the main issue here is the lack of more context: a name of what?   I suspect we don't want to use the full SQL naming convention, but rather something that will be friendlier to our environment. But again, it depends on the "name of what?" idea.

internal/parse/parser_test.go
```diff
@@ -12,3 +12,73 @@ func TestInit(t *testing.T) {
...
+}
+
+func TestIsNameByte(t *testing.T) {
+	assert.Equal(t, true, isNameByte('f'))
+	assert.Equal(t, true, isNameByte('0'))
+	assert.Equal(t, true, isNameByte('5'))
+	assert.Equal(t, false, isNameByte('à'))
+	assert.Equal(t, false, isNameByte('\n'))
+	assert.Equal(t, false, isNameByte('\r'))
+	assert.Equal(t, false, isNameByte('\t'))
```
niemeyer: All of these tests feel quite repetitive. It seems worth organizing them in a table format. Something like:
```
type parseHelperTest struct {
        bytef   func(byte) bool
        stringf func(string) bool
        result  bool
        data    string
}

var parseTests = []parseHelperTest{{
        {bytef: isNameByte, result: true, data: "f"}
        {bytef: isNameByte, result: true, data: "0"}
        {bytef: isNameByte, result: true, data: "5"}
        ... etc ...
}
```
Then just have one single test function that for-loops the whole thing at once.

internal/parse/parser_test.go
```diff
@@ -12,3 +12,73 @@ func TestInit(t *testing.T) {
 	assert.Equal(t, nil, err)
 	assert.Equal(t, (*ParsedExpr)(nil), expr)
 }
+
+func TestPeekByte(t *testing.T) {
+	p := NewParser()
+	p.Parse("")
+	assert.Equal(t, false, p.peekByte('a'))
```
niemeyer: Might be worth using go-check, which allows us to have suites and is what we use the most at Canonical. I can provide some simple samples if needed.  Okay to wait a bit until it's more obvious, if you feel like it.

niemeyer (changes requested):
Looks good. Just a comment on the tests.

There are some potential issues in the base that was already merged, but we'll need to review it elsewhere. For example: QueryPart doesn't look like something that should be public.

internal/parse/parser.go
fernape: Here I think of `name` as something we can use to represent:
* table name
* column name
* Go type name
* Go struct field name

So we can use it to know when to stop parsing and say, "okay, this is where the name ends". So the function is really simple in the sense that it only accepts alphanumerics and underscore. This will be a limitation mainly for table/column names since [I don't expect](https://go.dev/ref/spec#identifier) any Go code to have weird characters in the type names or struct field names. But column names can go wild in some other databases. I'm ok limiting this to some sensible defaults.

internal/parse/parser_test.go
fernape: I can do like in the experiment. I couple of questions though:

1. Aren't this table-like tests more difficult to debug?
```
--- FAIL: TestRunTable (0.00s)
    parser_test.go:32: 
                Error Trace:    /parser_test.go:32
                Error:          Not equal: 
                                expected: false
                                actual  : true
                Test:           TestRunTable
FAIL
```
with parser_test.go:32 being the `assert` line of the test function. There is no indication of the test that is actually failing.

2. The parser keeps a state. Do you want table-like tests for those too? I'm thinking of tests like [this](https://github.com/canonical/sqlair/pull/3/commits/7e6558f0e038167f8f7a4608a955f9d4296ec5bc#diff-acf51ab254114fd61f15d384a41705347f24cf03efacf3447fd9c42f32b35633R37). Wouldn't this over complicate the function using the test table?

internal/parse/parser.go
niemeyer: Go names don't use underscores in general, and it may also be problematic to *stop parsing* based on that specifically, because as @Aflynn50 mentioned above, the syntax is richer than what we consider "good naming". To be clear, the constraint on good names itself is a good idea, but it's worth considering how to apply it. For the time being, I suggest dropping it and then bringing it back together with the PR that uses it, so we have a bit more context for the conversation.

internal/parse/parser_test.go
niemeyer: For (1), just print the test line being executed with %#v on the failure cases, so people can easily spot the failure.

For (2), fair enough, let's keep it as it is for now so we can move on. It's repetitive but it's not crazy yet, so let's wait for until/if it bothers us more.

internal/parse/parser.go
fernape: Removed.

internal/parse/parser_test.go
fernape: As of now, all the tests are in a table-like fashion.

internal/parse/parser_test.go
```diff
@@ -1,11 +1,68 @@
...
+
+	{stringf0: p.skipSpaces, result: []bool{false}, input: "", data: []string{}},
+	{stringf0: p.skipSpaces, result: []bool{false}, input: "abc    d", data: []string{}},
+	{stringf0: p.skipSpaces, result: []bool{true}, input: "     abcd", data: []string{}},
+	{stringf0: p.skipSpaces, result: []bool{true}, input: "  \t  abcd", data: []string{}},
+	{stringf0: p.skipSpaces, result: []bool{false}, input: "\t  abcd", data: []string{}},
+
+	{stringf: p.skipString, result: []bool{false}, input: "", data: []string{"a"}},
+	{stringf: p.skipString, result: []bool{true, true}, input: "helloworld", data: []string{"hElLo", "w"}},
+	{stringf: p.skipString, result: []bool{true, true}, input: "hello world", data: []string{"hello", " "}},
```
niemeyer, line 42: Nice, thanks! :+1:

niemeyer (approved):

PR #6, opened 2022-09-20
Remove reflection code.

fernape:
We will get to it in time and it will be reviewed.


niemeyer (changes requested):
This is removing go.mod / go.sum / .gitignore as well. Should these be changed instead of removed?

fernape:
> This is removing go.mod / go.sum / .gitignore as well. Should these be changed instead of removed?

Sorry I miss this, GH keeps notifying me in my personal mail. Restored .gitignored and updated go.mod and go.sum.

niemeyer (approved):
Thanks!

PR #7, opened 2022-09-22
Add parts types.

fernape:
The input SDL statement will be splitted in different `queryParts`. We support three classes:

 * PassthroughPart
 * InputPart
 * OutputPart

Input and Output parts provide constructors and accessors for convenience. The constructors check for the correctness of the parameters and returns an error if the expressions are malformed.

The definition of the `QueryPart` interface has been moved to the `parts.go` file and has been made private.

### DIFF
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -16,7 +16,7 @@ type FullName struct {
 	Prefix, Name string
 }
 
-// InputPart represents a named parameter that will be send to the database
+// InputPart represents a named parameter that will be sent to the database
 // while performing the query.
 type InputPart struct {
 	FullName
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -30,7 +30,8 @@ func (p *InputPart) ToSQL() string {
 	return ""
 }
 
-// OutputPart represents an expression to be used as output in our SDL.
+// OutputPart represents a named target output variable in the SQL expression,
+// as well as the source table and column where it will be read from.
 type OutputPart struct {
 	Source FullName
 	Target FullName
@@ -44,7 +45,7 @@ func (p *OutputPart) ToSQL() string {
 	return ""
 }
 
-// BypassPart represents a part of the SDL that we want to pass to the
+// BypassPart represents a part of the expression that we want to pass to the
 // backend database verbatim.
 type BypassPart struct {
 	Chunk string
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -19,7 +19,7 @@ type FullName struct {
 // InputPart represents a named parameter that will be sent to the database
 // while performing the query.
 type InputPart struct {
-	FullName
+	Source FullName
 }
 
 // String returns a textual representation of the InputPart meant for debugging
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -21,16 +21,6 @@ type InputPart struct {
 	FullName
 }
 
-// NewInputPart is a constructor that returns a pointer to an InputPart.
-func NewInputPart(typeName string, tagName string) (*InputPart, error) {
-	return &InputPart{
-		FullName{
-			Prefix: typeName,
-			Name:   tagName,
-		},
-	}, nil
-}
-
 func (p *InputPart) String() string {
 	return ""
 }
@@ -43,19 +33,6 @@ func (p *InputPart) ToSQL() string {
 type OutputPart struct {
 	Source FullName
 	Target FullName
-}
-
-// NewOutputPart is a constructor that returns a pointer to an OutputPart.
-func NewOutputPart(tableName string, colName string,
-	typeName string, tagName string) (*OutputPart, error) {
-	return &OutputPart{
-		FullName{Prefix: tableName,
-			Name: colName,
-		},
-		FullName{Prefix: typeName,
-			Name: tagName,
-		},
-	}, nil
 }
 
 func (p *OutputPart) String() string {
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -26,7 +26,6 @@ func (p *InputPart) String() string {
 	return ""
 }
 
-// ToSQL returns a string with the SQL translation for the InputPart.
 func (p *InputPart) ToSQL() string {
 	return ""
 }
@@ -41,7 +40,6 @@ func (p *OutputPart) String() string {
 	return ""
 }
 
-// ToSQL returns a string with the SQL translation for the OutputPart.
 func (p *OutputPart) ToSQL() string {
 	return ""
 }
@@ -56,7 +54,6 @@ func (p *BypassPart) String() string {
 	return ""
 }
 
-// ToSQL returns a string with the SQL translation for the BypassPart.
 func (p *BypassPart) ToSQL() string {
 	return ""
 }
--- a/internal/parse/parts.go
+++ b/internal/parse/parts.go
@@ -1,155 +1,81 @@
 package parse
 
-import "fmt"
-
-// queryPart is an interface to be implemented by all parts.
+// A queryPart represents a section of a parsed SQL statement, which forms
+// a complete query when processed together with its surrounding parts, in
+// their correct order.
 type queryPart interface {
-	// String-like output for debugging purposes
+	// String returns the part's representation for debugging purposes.
 	String() string
 
-	// Translate part to its Sql equivalent
-	ToSql() string
+	// ToSQL returns the SQL representation of the part.
+	ToSQL() string
 }
 
-// qualifiedExpression represents a qualified expression
-// qualified. Examples are:
-//	... column as &Person ... 	<--- "column" is unqualified
-//	... t.column as &Person.name	<--- "name" and "column" and qualified
-type qualifiedExpression struct {
-	qualifier string
-	name      string
+type FullName struct {
+	Prefix, Name string
 }
 
-// InputPart represents an expression used as input in the SDL query.
-// Examples:
-//	$Address
-//	$Address.postal_code
+// InputPart represents a named parameter that will be send to the database
+// while performing the query.
 type InputPart struct {
-	goType
+	FullName
 }
 
 // NewInputPart is a constructor that returns a pointer to an InputPart.
 func NewInputPart(typeName string, tagName string) (*InputPart, error) {
-	if typeName == "" && tagName != "" {
-		return nil, fmt.Errorf("malformed InputPart")
-	}
 	return &InputPart{
-		goType{
-			qualifiedExpression{qualifier: typeName,
-				name: tagName,
-			},
+		FullName{
+			Prefix: typeName,
+			Name:   tagName,
 		},
 	}, nil
 }
 
-func (ip *InputPart) String() string {
+func (p *InputPart) String() string {
 	return ""
 }
 
-func (ip *InputPart) ToSql() string {
+func (p *InputPart) ToSQL() string {
 	return ""
-}
-
-// TypeName returns the name of the type in the input part or empty string if
-// none is present.
-func (ip *InputPart) TypeName() string {
-	return ip.qualifier
-}
-
-// TagName returns the name of the tag in the input part or empty string if
-// none is present.
-func (ip *InputPart) TagName() string {
-	return ip.name
-}
-
-// A column is an internal type that represents the columns selected in the
-// query.
-type column struct {
-	qualifiedExpression
-}
-
-// TableName returns the name of the table that qualifies a column or empty
-// string if there is none. Examples:
-//
-// select t.column	<--- TableName() returns "t"
-// select column	<--- TableName() returns ""
-func (c *column) TableName() string {
-	return c.qualifier
-}
-
-// ColumnName returns the name of the column that is selected in a query or an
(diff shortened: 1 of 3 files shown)


internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
+package parse
+
+import "fmt"
+
+// queryPart is an interface to be implemented by all parts.
```
niemeyer: This is a redundant definition: _"Foo must be implemented by all foos."._  In such cases it's better to try to convey why this is helpful, etc. For example, something along the lines of:
```
// A queryPart represents a section of a parsed SQL statement, which forms
// a complete query when processed together with its surrounding parts, in
// their correct order.
```

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+type queryPart interface {
+	// String-like output for debugging purposes
```
niemeyer: There's a strong convention in Go for how to spell the beginning of method and function documentation:
```
String returns the part's representation for debugging purposes.
```

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+	String() string
+
+	// Translate part to its Sql equivalent
```
niemeyer: Similarly:
```
// ToSQL returns the SQL representation of the part.
```
Let's please use capitalized `SQL` across the board as this is its common form.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+	ToSql() string
+}
+
+// qualifiedExpression represents a qualified expression
```
niemeyer: Same as above, "fooBar represents a foo bar" is not helpful and it wastes the reader time without adding additional knowledge (if there's text we tend to read it regardless). Besides addressing the docs, though, maybe we can also improve a bit the naming here, to help people figure what this is, since neither "qualifier" nor "expression" are very indicative of what it really does for us (e.g. "1+1" is also an expression, but that's not it).

How about:
```
// fullName represents a column or field name together with its Go type or table name. 
type fullName struct {
        Prefix string
        Name string
}
```

Or perhaps, even better, can we split it up?
```
type fieldName struct {
        Type, Name string
}
type columnName struct {
        Table, Name string
}
```

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+// qualified. Examples are:
+//	... column as &Person ... 	<--- "column" is unqualified
+//	... t.column as &Person.name	<--- "name" and "column" and qualified
```
niemeyer: The examples were useful, but it's a symptom of the issues described above. In other words, if we get the type name and documentation to be clear, it becomes obvious and we don't need the examples anymore.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+type qualifiedExpression struct {
+	qualifier string
+	name      string
+}
+
+// InputPart represents an expression used as input in the SDL query.
```
niemeyer:
```
// InputPart represents a named parameter that will be sent to the database to
// while performing the query.
```

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+	if typeName == "" && tagName != "" {
+		return nil, fmt.Errorf("malformed InputPart")
+	}
+	return &InputPart{
+		goType{
+			qualifiedExpression{qualifier: typeName,
+				name: tagName,
+			},
+		},
+	}, nil
```
niemeyer: That's quite a bit of syntax to replace what seems trivial:
```
&InputPart{
        Type: "Foo",
        Name: "bar",
}
```
Why the additional complexity? We don't need the error checking here as we can do it when processing the query.

Also, this is smelling slightly a bit like a name soup, in the sense we have several different ways to refer to the same data. For example, a Go field is now a "name", a "tag", and an "expression", that sits in a "type", or a "qualifier". We'll likely benefit from simplifying this: it's a type name, with a field name.

just here we have three different ways to name the same thing: tagName,

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (ip *InputPart) String() string {
```
niemeyer: Nitpick: "ip" reads as "IP" :)

I suggest using just "p" for all parts.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+// TypeName returns the name of the type in the output part or empty string if
+// none is present.
+func (gt *goType) TypeName() string {
+	return gt.qualifier
+}
+
+// TagName returns the name of the type in the output part or empty string if
+// none is present.
+func (gt *goType) TagName() string {
+	return gt.name
```
niemeyer: I'm a bit lost by now. There's a lot of abstraction in the types and methods above that just feels like renaming data back and forth. I may be missing part of the picture, but I'm really hoping that we can dramatically simplify this since all of that back and forth is simply handling (type, name) with half a dozen different names.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+	return ""
+}
+
+func (op *OutputPart) ToSql() string {
+	return ""
+}
+
+// passthroughPart represents a part of the SDL that we want to pass to the
+// backend database verbatim.
+type PassthroughPart struct {
```
niemeyer: "Bypass" is a shorter synonym for "Passthrough". Might be nicer to code with.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+func (gt *goType) TagName() string {
+	return gt.name
+}
+
+// OutputPart represents an expression to be used as output in our SDL.
+// Examples:
+// 	&Person
+//	&Person.name
+type OutputPart struct {
+	Column column
```
niemeyer: For example, considering the point above:
```
type OutputPart struct {
        Source fullName
        Target fullName
}
```
This takes down two or three levels of indirection, and avoids a large number of methods.

internal/parse/parts.go
```diff
@@ -0,0 +1,155 @@
...
+	GoType goType
```
niemeyer: There's an additional issue that I'll mention here for the lack of a better place: it's not a good practice to have public fields with private types, as that makes them extremely inconvenient to work with. So we need to review what types from this package are supposed to be public, and in which cases the type data is an implementation detail or publicly exposed data.

niemeyer (changes requested):
Thanks, Fernando. Some early comments below. As you can see, quite a bit of this is actually about agreeing on basic conventions and patterns to use. These early few reviews will be a bit more extensive, even if the PR is small, but as we get past that early phase things will become easier and simpler to get in.

internal/parse/parts.go
fernape: Using `FullName` so the type is exported.

internal/parse/parts.go
fernape: Done.

internal/parse/parts.go
fernape: Done, using `FullName`

internal/parse/parts.go
fernape: The thing here was to use `qualifier` as a very common name in grammars, especially SQL ones:
https://go.dev/ref/spec#Qualified_identifiers
http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt
```
    <qualified name> ::=
              [ <schema name> <period> ] <qualified identifier>

File: internal/parse/parts.go (hunk 5 above)
Comment (fernape) [reply internal/parse/parts.go]: Removed since we changed the naming.

File: internal/parse/parts.go (hunk 8 above)
Comment (fernape) [reply internal/parse/parts.go]: Yep, you're right.

File: internal/parse/parts.go (hunk 10 above)
Comment (fernape) [reply internal/parse/parts.go]: Shorted and equally clear. Changed.

File: internal/parse/parts.go (hunk 12 above)
Comment (fernape) [reply internal/parse/parts.go]: Now the type is public.

The general idea here was to use the same underlying type for expressing input parts and output parts. The latter have columns and go types but the structure is identical: `something` `.` `something`. So we had the underlying `qualifierExpression` that was accessed as `TypeName()` or `TableName()` depending on the location: the former makes sense for go types, the latter only for column specifications. Types here were private (`column` and `goType`) since they had accessor methods. I didn't see any use cases for client code importing the `parse` package and wanting to deal with the inner structure of the parts.

Thanks for the review!

File: internal/parse/parts.go [hunk 13]
```diff
@@ -1,155 +1,81 @@
...
-//	&Person.name
 type OutputPart struct {
-	Column column
-	GoType goType
+	Source FullName
+	Target FullName
 }
 
 // NewOutputPart is a constructor that returns a pointer to an OutputPart.
 func NewOutputPart(tableName string, colName string,
```
Aflynn50: Same as for NewInputPart

internal/parse/parts.go
```diff
@@ -1,155 +1,81 @@
...
-//	$Address.postal_code
+// InputPart represents a named parameter that will be send to the database
+// while performing the query.
 type InputPart struct {
-	goType
+	FullName
 }
 
 // NewInputPart is a constructor that returns a pointer to an InputPart.
 func NewInputPart(typeName string, tagName string) (*InputPart, error) {
```
Aflynn50: These constructors are not as vital now we have a simplified structure, a new InputPart can be instantiated straight from a FullName, which in all likely-hood is what we will have in the natural flow of the parser

internal/parse/parts.go
fernape: See below

internal/parse/parts.go
fernape: Done.
Removed constructors. Simplified tests.

internal/parse/parts.go
niemeyer: I understand the intention, but have you noticed how the very documentation for this type was explaining what "qualified" and "unqualified" meant for the particular context?

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+	// ToSQL returns the SQL representation of the part.
+	ToSQL() string
+}
+
+// FullName represents a table column or a Go type identifier.
+type FullName struct {
+	Prefix, Name string
+}
+
+// InputPart represents a named parameter that will be send to the database
```
niemeyer: s/send/sent/

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+// while performing the query.
+type InputPart struct {
+	FullName
```
niemeyer: I'm lacking a bit of context on usage, but gut feeling is that I'd make this symmetrical with OutputPart, having a field name here instead of embedding. Maybe Source?

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// String returns a textual representation of the InputPart meant for debugging
```
niemeyer: We don't need these boilerplate docstrings on every one of these methods, as this will be pretty much a copy & paste from the queryPart interface.  As a guideline, we should only document something when the text is actually improving the understanding of the reader about the code at hand. Otherwise, instead of helping we're rather hindering the understanding of the code by forcing the reading of something obvious.

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+// purposes.
+func (p *InputPart) String() string {
+	return ""
+}
+
+// ToSQL returns a string with the SQL translation for the InputPart.
```
niemeyer: Clear example of the above: "ToSQL returns a SQL for this Type"

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+func (p *InputPart) ToSQL() string {
+	return ""
+}
+
+// OutputPart represents an expression to be used as output in our SDL.
```
niemeyer: This is one case where we can actually help the next person:
```
// OutputPart represents a named target output variable in the SQL expression, as well
// as the source table and column where it will be read from.
```
Again, this would help the person fill the gaps that are not obvious from the code below, rather than simply stating "Output part is an output part".

internal/parse/parts.go
```diff
@@ -0,0 +1,68 @@
...
+	return ""
+}
+
+// ToSQL returns a string with the SQL translation for the OutputPart.
+func (p *OutputPart) ToSQL() string {
+	return ""
+}
+
+// BypassPart represents a part of the SDL that we want to pass to the
+// backend database verbatim.
```
niemeyer, line 49: That's nice, but let's please replace SDL with "expression" in our documentation across the board to avoid unnecessary acronyms.

niemeyer (approved):
This is a lot better, thank you!

Last past with tiny nitpicks, and this is good to go.

internal/parse/parts.go
fernape: Removed comments from methods already in the interface.

internal/parse/parts.go
fernape: Done. It is clearer now.

internal/parse/parts.go
fernape: Done. `Source` and `Target` make more sense for the `OutputPart` but it is not confusing here either, so I suppose we can use it for symmetry.

PR #10, opened 2022-10-10
Add bypass parser

Aflynn50:
This branch adds in a complete parser that parses any statement as only BypassParts. There are some changes from the initial skeleton of the parser that existed in main already. There is a new struct, ParsedExprBuilder, that factors out the slice of parsed parts from the Parser itself. This means the checkpoint does not have to make copies of this slice when save() is called.

The behaviour of the add function is retained from the experiment. When add(p) is called the BypassPart preceding p is also added to the slice of parts. The ParsedExprBuilder keeps track of the relevant markers to make this possible.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -44,7 +44,6 @@ func (cp *checkpoint) restore() {
 
 // advance moves the parser's index forward by one element.
 func (p *Parser) advance() bool {
-	p.skipSpaces()
 	if p.pos == len(p.input) {
 		return false
 	}
@@ -117,7 +116,6 @@ func (p *Parser) Parse(input string) (*ParsedExpr, error) {
 	p.init(input)
 	var peb parsedExprBuilder
 	for {
-		p.skipSpaces()
 		peb.partStart = p.pos
 		if !p.advance() {
 			break
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -9,6 +9,12 @@ import (
 type Parser struct {
 	input string
 	pos   int
+	// prevPart is the value of pos when we last finished parsing a part.
+	prevPart int
+	// partStart is the value of pos just before we started parsing the part
+	// under pos. We maintain partStart >= prevPart.
+	partStart int
+	parts     []queryPart
 }
 
 func NewParser() *Parser {
@@ -19,22 +25,31 @@ func NewParser() *Parser {
 func (p *Parser) init(input string) {
 	p.input = input
 	p.pos = 0
+	p.prevPart = 0
+	p.partStart = 0
+	p.parts = []queryPart{}
 }
 
 // A checkpoint struct for saving parser state to restore later. We only use
 // a checkpoint within an attempted parsing of an part, not at a higher level
 // since we don't keep track of the parts in the checkpoint.
 type checkpoint struct {
-	parser *Parser
-	pos    int
+	parser    *Parser
+	pos       int
+	prevPart  int
+	partStart int
+	parts     []queryPart
 }
 
 // save takes a snapshot of the state of the parser and returns a pointer to a
 // checkpoint that represents it.
 func (p *Parser) save() *checkpoint {
 	return &checkpoint{
-		parser: p,
-		pos:    p.pos,
+		parser:    p,
+		pos:       p.pos,
+		prevPart:  p.prevPart,
+		partStart: p.partStart,
+		parts:     p.parts,
 	}
 }
 
@@ -42,6 +57,9 @@ func (p *Parser) save() *checkpoint {
 // checkpoint.
 func (cp *checkpoint) restore() {
 	cp.parser.pos = cp.pos
+	cp.parser.prevPart = cp.prevPart
+	cp.parser.partStart = cp.partStart
+	cp.parser.parts = cp.parts
 }
 
 // ParsedExpr is the AST representation of an SQL expression.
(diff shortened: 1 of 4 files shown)


niemeyer (changes requested):
I don't think I understand this PR.  It seems to change the agreements we have in the sketch while introducing something that is clearly non-functional. The particular base process that we ended up in the sketch took us a bit to agree on, IIRC, so while I'm of course happy to improve on it, it seems unwise to do that with a partial implementation that is clearly non-functional and suffers some of the original problems that we discussed back then.

Aflynn50:
Admittedly the rational behind these design changes should have been more clearly explained in this PR. The benefits are not apparent with only the Bypass parser present. 

The experiment did not include ParsedExprBuilder. That is the main change introduced in this PR. The ParsedExprBuilder separates the aggregation of parts from the checkpoint functionality, primarily to aid clarity. 

In the full parser we plan to have a separate, complete, parsing functions for input and output parts. This functionality will mirror the experiment. These functions attempt to parse an InputPart/OutputPart. If they fail they restore the checkpoint, thus consuming no input. If they succeed, then, with the current DSL specification, there is no proceeding input that could cause that parse to be rolled back. Therefore, there is no reason for the parts to be stored in the checkpoint. 

The ParsedExprBuilder is introduced to store this information about the parts desperately from the Parser. It is then not included in the checkpoint. We can continue to use the checkpoint as it has been used in the experiment. For the reasons in the paragraph above, it does not affect its functionality.

The ParsedExprBuilder also stores the information about the start and end of the previous chunk of text to bypass. And the checkpoint is reduced to only its essential parts.

This approach has been fully tested in a complete implementation here: https://github.com/Aflynn50/sqlair/tree/lazy-prepare .

internal/parse/parser.go
```diff
@@ -1,57 +1,140 @@
...
 
-// ParsedExpr represents a parsed expression.
-// It has a representation of the original SQL statement in terms of QueryParts
+// advance moves the parser's index forward by one element.
+func (p *Parser) advance() bool {
+	p.skipSpaces()
+	if p.pos == len(p.input) {
+		return false
+	}
+	p.pos++
```
niemeyer: Although we're not seeing the rest of the logic here, this feels a bit suspect.. we can't advance the pointer without knowing what we're advancing over. Every parse or skip function moves the point forward towards the next character, so by the time something is parsed in theory we should be at the next position to be parsed. If we advance blindly from there, we ignored something. That's double troubling if we're both ignoring spaces and blindly moving the pointer forward. Also, the skipping of spaces probably needs accounting, in the sense that we need to know if there was indeed a space there serving as a separator from the previous token, otherwise we'll need to know how the underlying SQL parser behaves (e.g. is "foo()" the same as "foo ()?"). Although we don't have control over the entire syntax, we have some control over ours and can enforce proper spacing around our own tokens. But we can't do that if we just optionally skip and blindly move forward, etc.  Makes sense?

internal/parse/parser.go
```diff
@@ -1,57 +1,140 @@
...
+		}
+		out = out + p.String()
+	}
+	out = out + "]"
+	return out
+}
+
+// parsedExprBuilder keeps track of the parts parsed so far, along with
+// information for parsing the BypassPart between input/output parts.
+type parsedExprBuilder struct {
```
niemeyer: It's also not clear what the benefit of this construct is, when we already have a parser that is already re-initialized at the same moment that this type is created. We might make a Parser that is stateless and holds the state in memory until the parsing is done, and then it might be safely used in parallel without races. But as it is we have neither, as half of the state of the parsing is in Parse, and half is in the builder.  Am I missing something?  Don't we get to the same place if these fields are in Parser itself?

internal/parse/parser_test.go
```diff
@@ -44,8 +42,8 @@ var parseTests = []parseHelperTest{
 
 func TestRunTable(t *testing.T) {
 	for _, v := range parseTests {
-		p.Parse(v.input)
-		for i, _ := range v.result {
+		p.init(v.input)
```
niemeyer: Blackbox tests are generally much nicer than whitebox ones, as they test the real external expectations of the implementation, while allowing the internal implementation to change freely as long as the public promises are respected.  Sometimes we may need to dig into the implementation to test certain things, but it doesn't look to be the case here.  Can we please rename this package (first line) to "parser_test"?  This will make us more honest about not touching the internal implementation unless that's really required.  Not sure if we have other test files in a similar situation, but if so can we please change them too?

niemeyer (changes requested):
Okay, given that we have a more complete sketch, let's see how far we can get with this approach.

A few comments on the details of the PR then:

internal/parse/parser.go
Aflynn50: You're certainly right in it feeling suspect, in my prototype it meant it was impossible to parse consecutive unseparated expressions (e.g. `&Person&Address`). I've in-placed the advance function so that in subsequent PRs the `p.pos++` will only be executed if nothing is parsed on in that iteration.

To keep track of the spaces we could instead:
Start and end the IO parser functions by calling skipSpaces and when we generate the SQL for them we put spaces in at each end. That way the Bypass expressions contain no spaces before/after an IO part but they end up in the generated SQL.

This approach will mostly require changes in future PRs and for now we only remove the calls to skipSpaces from the Parse function. Does this sound like a reasonable approach to you?

internal/parse/parser.go
Aflynn50: It doesn't add any further functionality. The benefits are only clarity, and the small optimisation that the list of parsed parts does not have to be copied every time we save a checkpoint. I can revert it back to the integrated parser that was used in the experiment if you think that would be better.

internal/parse/parser_test.go
Aflynn50: Done. The parse function tests that are necessarily whitebox are now in parse.go and the rest in separate test packages.

internal/parse/parser.go
niemeyer: The goal is noble, but per comment above the problem is that we cannot claim there is any clarity improvement for this particular case, because the Parser and the ParserExprBuilder are strictly defined in terms of each other's implementations. The Parser has some state which is re-initialized every time a parsing happens, and at that exact moment a new Builder is created. In other words, the Parser state and the Builder state are strictly connected and reset at once, and the Parser knows about the Builder's details, while the Builder knows about the exact implementation of the Parser and its private fields. So these don't really look like two different types.. they are indeed a single type with two names, with no clear abstraction or encapsulation across them.

We can clean this up in at least two ways:

1. Take the rest of parsing logic content outside of Parser itself. The only benefit I can see of doing this is that Parser becomes stateless, but then why even have a type for this?

2. Otherwise, merge the Builder back into the Parser itself, which also makes some sense as the "parts builder" is really the only point of having a Parser at all (it parses the expression).

niemeyer (approved):
Thanks!

PR #11, opened 2022-10-18
Add quote and input parser

Aflynn50:
Added the input and quote parser along with some extra tests


internal/parse/parts_test.go
```diff
@@ -27,14 +27,15 @@ func TestOutputPart(t *testing.T) {
 				Prefix: "mytable",
 				Name:   "mycolumn",
 			}},
-		parse.FullName{
-			Prefix: "mytype",
-			Name:   "mytag",
-		},
+		[]parse.FullName{
+			parse.FullName{
```
niemeyer: The double type name is not required in those cases. This should work:
```
[]parse.FullName{{
    Prefix: "mytype",
    Name: "mytag",
}},
```

internal/parse/parse_internal_test.go
```diff
@@ -1,61 +0,0 @@
...
-type parseHelperTest struct {
-	bytef    func(byte) bool
-	stringf  func(string) bool
-	stringf0 func() bool
-	result   []bool
-	input    string
-	data     []string
-}
-
-func TestRunTable(t *testing.T) {
```
niemeyer: Why was this dropped?  Is it coming back in a different shape?

internal/parse/parts.go
```diff
@@ -43,18 +43,25 @@ func (p *InputPart) ToSQL() string {
...
-	Target FullName
+	Target []FullName
 }
 
 func (p *OutputPart) String() string {
-	var colString string
+	var sourceString string
 	for _, col := range p.Source {
-		colString = colString + col.String() + " "
+		sourceString = sourceString + col.String() + " "
```
niemeyer: The best way to implement the logic in this function is with a bytes.Buffer, otherwise a lot of unnecessary garbage is generated due to the allocation and release of all these immutable strings. Alternatively, why aren't we just using the `%v` verb of fmt.Sprintf for all these strings?  It looks very close, and it's zero effort on our side.

niemeyer (changes requested):
Thanks. Just a few questions:

internal/parse/parse_internal_test.go
Aflynn50: It is! In a different file - see #13

internal/parse/parts.go
Aflynn50: Currently we don't use `fmt.Sprintf` so that we can make it more readable. For an expression &Person.\* we would get "FullName{Prefix=Person, Name=\*}" with Sprintf but instead we simply print "Person.\*".

I'll have a look into changing to use bytes.Buffer, that looks neat

niemeyer: > For an expression &Person.* we would get "FullName{Prefix=Person, Name=}"

Not really. Try it out. I think it'll help you, even if you use bytes.Buffer, so you don't need to do everything by hand.

Aflynn50: Though I've now just realised the Sprintf will use your custom string function so it does actually work very nicely! I've changed the String for ParsedExpr to use the bytes.Buffer pattern though

niemeyer: That wasn't actually the point.. have you tried to use the `%v` verb on a struct?  The output is not the one suggested above.

Aflynn50: I have, that removes the Field names from the output, that is actually the result for %+v that I wrote above

Aflynn50: though with %v we loose the "."

internal/parse/parser.go
```diff
@@ -113,14 +121,28 @@ func (p *Parser) add(part queryPart) {
 // a pointer to a ParsedExpr.
 func (p *Parser) Parse(input string) (*ParsedExpr, error) {
 	p.init(input)
+
 	for {
 		p.partStart = p.pos
-		if p.pos == len(p.input) {
+
+		if ip, ok, err := p.parseInputExpression(); err != nil {
+			return nil, fmt.Errorf("parser error: %s", err)
```
niemeyer: "cannot parse expression: " would be a better prefix here, as it avoids the idea that the parser itself is in error. But I suggest adding it at the top level with something like this as it can handle it generally:
```
func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
    defer func() {
        if err != nil {
            err = fmt.Errorf("cannot parse expression: %w", err)
        }
    }()
```
This way the rest of the logic can ignore the issue altogether.

internal/parse/parser.go
```diff
@@ -113,14 +121,28 @@ func (p *Parser) add(part queryPart) {
...
+
+		} else if sp, ok, err := p.parseStringLiteral(); err != nil {
+			return nil, fmt.Errorf("parser error: %s", err)
+		} else if ok {
+			p.add(sp)
+
+		} else if p.pos == len(p.input) {
 			break
+		} else {
+			p.pos++
```
niemeyer: The fact this ignores content arbitrarily doesn't feel entirely right as it'll start parsing in the middle of other expressions, but we can revisit this later as it's already in the existing code.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+// construct, if they n't parse they return false, restore the parser and leave
+// the default value in  other return type
+
+func (p *Parser) parseIdentifier() (string, bool) {
+	p.skipSpaces()
+	if p.pos >= len(p.input) {
+		return "", false
+	}
+	if p.peekByte('*') {
+		p.pos++
```
niemeyer, line 229: If peekByte is replaced by skipByte, the line above is not required. That's how these skip helpers are generally used.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
-	stringf0 func() bool
-	result   []bool
-	input    string
-	data     []string
+// These functions attempt to parse some construct, they return a bool and that
+// construct, if they n't parse they return false, restore the parser and leave
+// the default value in  other return type
+
+func (p *Parser) parseIdentifier() (string, bool) {
+	p.skipSpaces()
```
niemeyer: The logic of these functions is that when it returns false they are supposed to preserve the state they got when coming in. This function is manipulating the state, and returning false with a changed input, which could break the rest of the parser depending on how these are used. The comment right above this function is also documenting this.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+	if p.pos >= len(p.input) {
+		return "", false
+	}
+	if p.peekByte('*') {
+		p.pos++
+		return "*", true
+	}
+
+	idStart := p.pos
+	if !isNameByte(p.input[p.pos]) {
```
niemeyer: Looks like there's a skipName embedded here.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+	for i = p.pos; i < len(p.input); i++ {
+		if !isNameByte(p.input[i]) {
+			break
+		}
+	}
+	p.pos = i
+	return p.input[idStart:i], true
+}
+
+// Parses a column name or a Go type name. If parsing a Go type name then
```
niemeyer: The convention for comments in methods and functions in Go is similar to this:
```
// parseFullName parses a column name or a Go type name. ... etc ...
```
But this comment is also a bit misleading, because the Go type is akin to the table name, and not the column name, which would be analogous to the field name. What follows is also somewhat tricky to read for an operation that seems relatively simple.

How about something like this:
```
// parseFullName parses a column name or Go field name, optionally dot-prefixed by
// its table or type name respectively.
```
That's probably all we need. The fields in the returned type are literally `Name` and `Prefix`, so we shouldn't have to explain further.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+// When parsing a column the table name (if extant) is in FullName.Prefix and
+// the column name is in FullName.Name func (p *Parser)
+func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+	cp := p.save()
+	var fn FullName
+	p.skipSpaces()
+	if id, ok := p.parseIdentifier(); ok {
+		fn.Prefix = id
+		if p.skipByte('.') {
+			if id, ok := p.parseIdentifier(); ok {
```
niemeyer, line 268: This makes it look like the skipSpaces at the beginning of parseIdentifier is wrong. We probably don't want spaces here?

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
-		{bytef: p.skipByteFind, result: []bool{false, true, true}, input: "abcde", data: []string{"x", "b", "c"}},
-		{bytef: p.skipByteFind, result: []bool{true, false}, input: "abcde ", data: []string{" ", " "}},
+		// Case 2: Multiple columns.
+	} else if p.skipByte('(') {
+		col, ok := p.parseFullName(columnId)
+		cols = append(cols, col)
+		p.skipSpaces()
+		// If the column names are not formated in a recognisable way then give
+		// up trying to parse.
+		if !ok {
```
niemeyer: Why do we wait until here to give up?

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+			return cols, false
```
niemeyer: Same comment about saving the state from earlier above.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+		}
+		for p.skipByte(',') {
+			p.skipSpaces()
+			col, ok := p.parseFullName(columnId)
+			p.skipSpaces()
+			if !ok {
```
niemeyer: Same about waiting to bail.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
+				return cols, false
```
niemeyer: ... and state.

internal/parse/parser.go
```diff
@@ -187,57 +209,143 @@ func isNameByte(c byte) bool {
...
-				log.Printf("Test %#v failed. Expected: '%t', got '%t'\n", v, result, v.result[i])
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseInputExpression parses an SDL input go-defined type to be used as a
+// query argument.
+func (p *Parser) parseStringLiteral() (*BypassPart, bool, error) {
+	cp := p.save()
+	p.skipSpaces()
```
niemeyer: Per above, these skipSpaces seems a bit problematic as it prevents detecting cases where spaces are not present. The outer layer is probably the right place to be doing the skipping.

niemeyer (changes requested):
Thanks for the PR, Alastair. Some additional comments here.  Some of this has been previously discussed with Fernando. Did you discuss this with him previously?

internal/parse/parser.go
Aflynn50: This is a nice approach. Though would this not mean we ignore errors until we have consumed the whole input string? It would wait until the Parse function has finished before dealing with errors so if there was a problem right at the start of the input we would still try and parse the rest rather than giving up when the error occurred.

Parsing should be relatively low cost compared to the other things we are doing though so this could be fine.

Aflynn50: I've changed the errors to use wrapping and the %w. I've held off adding the defer function - waiting on your reply to the previous comment

internal/parse/parser.go
Aflynn50: This else is only triggered when theres nothing else starting at that position that can be parsed. The parser then moves on to the next position and tries there. When something is successfully parsed, all the input that was skipped over before getting to it is then added as a BypassPart in `p.add`. Even though its ignoring content, it will be eventually parsed into a BypassPart.

internal/parse/parser.go
Aflynn50: Changed to skipByte

internal/parse/parser.go
Aflynn50: removed skipSpaces and amended tests

internal/parse/parser.go
Aflynn50: factored out the skipName function

internal/parse/parser.go
Aflynn50: Changed comment to suggested.

internal/parse/parser.go
Aflynn50: True

internal/parse/parser.go
Aflynn50: I've realised that the parseColumns function is not even meant to be in this PR, it should be in the next one. It has been removed

internal/parse/parser.go
Aflynn50: parseColumns removed from this PR

internal/parse/parser.go
Aflynn50: Same as above.

internal/parse/parser.go
Aflynn50: Same as above.

internal/parse/parser.go
Aflynn50: skipSpaces removed.

internal/parse/parser.go
niemeyer: Just for the record, we discussed the problem of this approach in a call. It should do a more controlled skip towards the next start point.

internal/parse/parser.go
```diff
@@ -111,16 +119,36 @@ func (p *Parser) add(part queryPart) {
...
+		} else if p.pos == len(p.input) {
 			break
+		} else {
+			// If nothing above can be parsed we advance the parser.
+			p.pos++
 		}
-		p.pos++
 	}
-	// Add any remaining unparsed string input to the parser.
+	// Add any remaining uparsed string input to the parser
```
niemeyer: Looks like the comment was corrupted here. The former line was right.

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
+	}
+	if p.skipByte('*') {
+		return "*", true
+	}
 
-		{bytef: p.skipByte, result: []bool{false}, input: "", data: []string{"a"}},
-		{bytef: p.skipByte, result: []bool{false}, input: "abc", data: []string{"b"}},
-		{bytef: p.skipByte, result: []bool{true, true}, input: "abc", data: []string{"a", "b"}},
+	idStart := p.pos
+	if p.skipName() {
```
niemeyer, line 254: This function looks much better now, thanks.

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
+	return fn, false
+}
 
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "abc    d", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "     abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "  \t  abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "\t  abcd", data: []string{}},
+// parseInputExpression parses an SDL input go-defined type to be used as a
+// query argument.
```
niemeyer: This sentence is reading a bit curiously. How about:
```
// parseInputExpression parses an input expression of the form $Type.name.
```

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
-		{stringf: p.skipString, result: []bool{true, true}, input: "helloworld", data: []string{"hElLo", "w"}},
-		{stringf: p.skipString, result: []bool{true, true}, input: "hello world", data: []string{"hello", " "}},
+	p.skipSpaces()
+	if p.skipByte('$') {
+		fn, ok = p.parseFullName(typeId)
+		if !ok {
+			err = fmt.Errorf("malformed input type")
+		}
+		p.skipSpaces()
+		return &InputPart{fn}, true, err
```
niemeyer: This function is returning an error with `ok=true`, which doesn't look right. The `!ok` case above should be immediately bailing to prevent this, and this should return `... , true, nil` instead. 

On a more general note, it's a good practice to return nil/false/zero values in case of errors, except in the rare cases where the intermediate values are actually desired.

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
-				result = v.stringf0()
-			}
-			if v.result[i] != result {
-				log.Printf("Test %#v failed. Expected: '%t', got '%t'\n", v, result, v.result[i])
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseInputExpression parses an SDL input go-defined type to be used as a
+// query argument.
```
niemeyer: Similar to the above case, it would be nice to have this a bit more readable.

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
+func (p *Parser) parseStringLiteral() (*BypassPart, bool, error) {
+	cp := p.save()
+
+	var err error
+
+	if p.pos < len(p.input) {
+		c := p.input[p.pos]
+		if c == '"' || c == '\'' {
+			p.skipByte(c)
+			if !p.skipByteFind(c) {
```
niemeyer, line 312: It's okay to ignore this for now, but we need at the very least a `// TODO Handle escaping` here.

internal/parse/parser.go
```diff
@@ -187,57 +215,105 @@ func isNameByte(c byte) bool {
...
+				// Reached end of string and didn't find the closing quote
+				err = fmt.Errorf("missing right quote in string literal")
 			}
+			return &BypassPart{p.input[cp.pos:p.pos]}, true, err
```
niemeyer: Same issue as above.

niemeyer (changes requested):
Thanks. Here is another round:

internal/parse/parser.go
Aflynn50: fixed

internal/parse/parser.go
Aflynn50: changed.

internal/parse/parser.go
Aflynn50: I've changed the returns so we return nil in all cases except where an error is expected.

I've also added a comment explaining the way the parse functions work with return values. There are three different return states that a parse function can be in as explained in the comment in the code.

internal/parse/parser.go
Aflynn50: changed.

internal/parse/parser.go
Aflynn50: changed.

internal/parse/parser.go
Aflynn50: I'll add it to our Jira board too

internal/parse/parser.go
```diff
@@ -187,57 +214,108 @@ func isNameByte(c byte) bool {
...
 
-		{bytef: p.skipByteFind, result: []bool{false}, input: "", data: []string{"a"}},
-		{bytef: p.skipByteFind, result: []bool{false, true, true}, input: "abcde", data: []string{"x", "b", "c"}},
-		{bytef: p.skipByteFind, result: []bool{true, false}, input: "abcde ", data: []string{" ", " "}},
+// parseFullName parses a column name or Go field name, optionally dot-prefixed by
+// its table or type name respectively.
+func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+	cp := p.save()
+	var fn FullName
+	p.skipSpaces()
```
niemeyer: This is still skipping spaces. I thought we had agreed this was a job for the call site on the previous review / call?

internal/parse/parser.go
```diff
@@ -187,57 +214,108 @@ func isNameByte(c byte) bool {
...
+// parseInputExpression parses an input expression of the form $Type.name.
+func (p *Parser) parseInputExpression() (*InputPart, bool, error) {
+	cp := p.save()
+	var fn FullName
+	var ok bool
 
-		{stringf: p.skipString, result: []bool{false}, input: "", data: []string{"a"}},
-		{stringf: p.skipString, result: []bool{true, true}, input: "helloworld", data: []string{"hElLo", "w"}},
-		{stringf: p.skipString, result: []bool{true, true}, input: "hello world", data: []string{"hello", " "}},
+	p.skipSpaces()
```
niemeyer: Same here.

internal/parse/parser.go
```diff
@@ -111,14 +119,33 @@ func (p *Parser) add(part queryPart) {
...
 	for {
 		p.partStart = p.pos
-		if p.pos == len(p.input) {
+
+		if ip, ok, err := p.parseInputExpression(); err != nil {
+			return nil, err
+		} else if ok {
+			p.add(ip)
+
+		} else if sp, ok, err := p.parseStringLiteral(); err != nil {
```
niemeyer: The formatting here feels slightly hard to read and follow. Have you tried breaking this down into independent blocks (no else) with their own continue/break/return?  Half of them are already returning or breaking, which means they'll interrupt the flow regardless of the else.

internal/parse/parser.go
```diff
@@ -187,57 +214,108 @@ func isNameByte(c byte) bool {
...
-				log.Printf("Test %#v failed. Expected: '%t', got '%t'\n", v, result, v.result[i])
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseStringLiteral parses quoted expressions and ignores their content.
+func (p *Parser) parseStringLiteral() (*BypassPart, bool, error) {
+	cp := p.save()
+
+	if p.pos < len(p.input) {
```
niemeyer, line 307: ... and no skipping here, which means these cases are indeed unusual.

internal/parse/parser_test.go
```diff
@@ -27,142 +29,99 @@ type M map[string]any
...
 	var tests = []struct {
-		input             string
-		expectedParsed    string
-		prepArgs          []any
-		completeArgs      []any
-		expectedCompleted string
+		input          string
+		expectedParsed string
 	}{
 		{
```
niemeyer: Per chat online, can we please do these in the standard compact "cuddled" form:
```
{{
    ...
}, {
    ...
}}
```

internal/parse/parser_test.go
```diff
@@ -171,106 +130,187 @@ func TestRound(t *testing.T) {
...
+			t.Errorf("test %d failed (Parse): input: %s\nexpected: %s\nerr: %s\n", i, test.input, test.expectedParsed, err)
+		} else if parsedExpr.String() != test.expectedParsed {
+			t.Errorf("test %d failed (Parse): input: %s\nexpected: %s\nactual:   %s\n", i, test.input, test.expectedParsed, parsedExpr.String())
 		}
 	}
 }
+
+// We return a proper error when we find an unbound string literal
+func TestUnfinishedStringLiteral(t *testing.T) {
+	sql := "select foo from t where x = 'dddd"
```
niemeyer: Per conversation in the sprint, can we please standardize on using casing such as `SELECT foo FROM` to make the statements more readable?  This is already the case in the tests above.

niemeyer (changes requested):
Thanks. Maybe last one:

internal/parse/parser.go
Aflynn50: Ah maybe I changed the wrong version before. Its fixed now.

internal/parse/parser.go
Aflynn50: Removed

internal/parse/parser.go
Aflynn50: I think the idea for doing it like that was to avoid parsing two different things in one loop but on reflect there seems to be nothing wrong with doing that. Changed to separate into different blocks

internal/parse/parser_test.go
Aflynn50: Ah I thought you were referring in your comment to instances such as []FullName{FullName{.... where the type was specified unnecessary in a list. Changed.

internal/parse/parser_test.go
Aflynn50: Standardised.

internal/parse/parser.go
```diff
@@ -134,13 +134,15 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 			return nil, err
 		} else if ok {
 			p.add(ip)
+		}
 
-		} else if sp, ok, err := p.parseStringLiteral(); err != nil {
+		if sp, ok, err := p.parseStringLiteral(); err != nil {
 			return nil, err
 		} else if ok {
 			p.add(sp)
```
niemeyer, line 143: The new pattern here lead to a different outcome. For the logic to be equivalent, this branch needs to continue/break/return. Maybe that's intended, but my suggestion was to just improve the readability of the code instead of changing the behavior.

internal/parse/parser_test.go
niemeyer: Doesn't look like it changed. Indeed, formatting looks misaligned as well now?

niemeyer (changes requested):

internal/parse/parser.go
Aflynn50: I've added in the continues so it retains the original, more stable, behaviour

internal/parse/parser_test.go
Aflynn50: Ah, I misunderstood the comment, I've changed it and it looks a lot better now

internal/parse/parser.go
niemeyer: Thanks!

niemeyer (approved):
Thanks for the changes, and patience, Alastair.

PR #12, opened 2022-11-03
Create a new reflection code.

fernape:
* Use `typeinfo`  package name to avoid clashing with Go's reflection package.
* Use a simple map-based cache to store reflection information.
* `Info struct` holds the relevant reflection information.
* Implement reflection for simple types, structs and `sqlair.M`.

  A more complete version that allows for setting and getting values is in https://github.com/fernape/sqlairexperiment/tree/minimal-reflection


typeinfo/typeinfo.go
```diff
@@ -0,0 +1,137 @@
...
+		cmutex.Lock()
+		cache[tname] = info
+		cmutex.Unlock()
+	}
+
+	return info, nil
+}
+
+// generate produces and returns reflection information for the input
+// reflect.Value that is specifically required for Sqlair operation.
```
niemeyer: Either SQLair, or SQLAir (probably the former?), but definitely not Sqlair. :)

Please do a global search as there are other instances.

typeinfo/typeinfo.go
```diff
@@ -0,0 +1,137 @@
+package typeinfo
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+)
+
+var cmutex sync.RWMutex
```
niemeyer: This would look better as `cacheMutex`, as this is a global and it's nice to know what it's talking about when reading code that references it.

typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// Index of this field in the structure
+	Index int
+
+	// OmitEmpty is true when "omitempty" is
+	// a property of the field's "db" tag.
+	OmitEmpty bool
+}
+
+// Struct represents reflected information about a struct type.
```
niemeyer: s/Struct/Info/

typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
+package typeinfo
```
niemeyer: The suggestion at the sprint was to make this into private types and functions, but it's okay to have it in a package if you'd prefer that. That said, if it's a package it definitely needs to be inside an intenal/typeinfo package, as this shouldn't be exposed or used by anyone else but ourselves inside the implementation.

typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
...
+	OmitEmpty bool
+}
+
+// Struct represents reflected information about a struct type.
+type Info struct {
+	Type reflect.Type
+
+	// TagsToFields  maps "db" tags to struct fields.
+	// Sqlair does not care about fields without a "db" tag.
+	TagsToFields map[string]Field
```
niemeyer: Such fields are better named in the singular, as that'll read better when using it:
```
field := TagToField[tag]
```

typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// FieldsToTags maps field names to tags
```
niemeyer: The comment from both of these fields may be dropped as they are pretty much reading the code, and obvious from context.

typeinfo/typeinfo.go
```diff
@@ -0,0 +1,137 @@
...
+var cache = make(map[string]Info)
+
+// Return the Info of a type name passed as parameter
+// Returns error if the type has not been reflected yet.
+func GetInfoFromName(typeName string) (Info, error) {
```
niemeyer: The "Get" prefix is most often avoided in Go, and the established convention in Go is to start comments as:
```
// InfoFromName returns blah blah blah
```
With that said, we can't use the type name here, so all this logic needs changing. What happens when two types have the same name?  The code from go-yaml that we discussed does that right, and can be used as a reference still.

typeinfo/typeinfo.go
```diff
@@ -0,0 +1,137 @@
...
+		return info, nil
+	}
+
+	info, err := generate(v)
+	if err != nil {
+		return Info{}, err
+	}
+
+	// Do not cache for "M" types
+	if !(info.Type.Kind() == reflect.Map && info.Type.Name() == "M") {
```
niemeyer: The M logic from go-yaml doesn't seem relevant for SQLair. What's the idea here?

typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
...
+	// Index of this field in the structure
+	Index int
+
+	// OmitEmpty is true when "omitempty" is
+	// a property of the field's "db" tag.
+	OmitEmpty bool
+}
+
+// Struct represents reflected information about a struct type.
+type Info struct {
```
niemeyer: The info type seems somewhat short. I suspect we'll need more data, but it's okay to start small.

typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,144 @@
...
+	var i int
+	var s string
+	var mymap map[string]string
+
+	{
+		info, err := GetTypeInfo(i)
+		assert.NotEqual(t, info, Info{})
+		assert.Nil(t, err)
+	}
+	{
```
niemeyer: This is unconventional. It's okay to just use spaces to separate the the different cases. With that said, the test above doesn't seem super helpful as it's just ignoring the result altogether.  Why is it not testing the actual results?

niemeyer (changes requested):
Some preliminar notes here:

typeinfo/typeinfo.go
fernape: This function is for use in a future `prepare()` function that is already [prototyped](https://github.com/Aflynn50/sqlair/blob/prepare/internal/prepare/prepare.go#L22). When preparing the query, at some point we have the textual name of the type (as in `Person` if the query contains `SELECT col_foo AS &Person.Foo ...`)

Aflynn50: To address your point about two types having the same name:

In the prepare prototype the flow is:
	- generate reflect information for all the (empty instantiations of) types and store it in the cache
	- cycle through the IO expressions in the parsed query and for each Go object specified (`&Person.name`), look up the reflection information using GetInfoFromName
	- use this reflection info to run relevant checks (such as the tag written in the IO expression being a member of the Go struct, or generating the list of columns for an asterisk expression

In the [specification](https://docs.google.com/document/d/1FbRZisjK0z7shy8m1cHzzEMDpxGwsK-54xDpf9HY4Xo/edit#) we have the condition that two types should not have the same name. There should be a type alias declared for instances where the user wants to scan two separate objects of the same type. This code was designed with that condition in mind.

It is also worth noting that these cached objects are only being used for their type information; not their actual value. So for instances such as certain types of Join where the user might write the same object in the query twice, they still only have to pass the a single object to prepare. Also the order in which they pass the objects does not matter, it can be any order.

typeinfo/typeinfo.go
fernape: Using SQLAir as in the README.

typeinfo/info.go
fernape: Moved to internal/typeinfo.

typeinfo/typeinfo.go
fernape: `s/GetInfoForName/InfoForName/`

typeinfo/typeinfo_test.go
fernape: Improved tests.

typeinfo/info.go
fernape: If you are OK with it, we can add things as they are needed.

typeinfo/typeinfo.go
fernape: The `sqlair.M` was in the [original specification](https://docs.google.com/document/d/1FbRZisjK0z7shy8m1cHzzEMDpxGwsK-54xDpf9HY4Xo/edit#) and we have kept it that way. It was also mentioned in the sprint during a meeting with the juju team.

typeinfo/typeinfo.go
niemeyer: > In specification we have the condition that two types should not have the same name.

The API we designed in the specification is able to handle such cases properly. The constraint for type naming is per expression, not globally, otherwise imagine what happens if we have 5 different `DataInfo` objects, created by 5 different packages written by 5 different authors, in the overall application. How would SQLair pick up the right data for each type? Insanity, right?

typeinfo/typeinfo.go
niemeyer: Again, what's the plan about `M` handling here?

fernape: Sorry, I'm missing something here. We are caching by type. So if we have something like:
```
       var mymap M
       mymap = make(M)
       mymap["foo"] = 7 
       mymap["bar"] = "baz"
...
...
       var mysecondmap M
       mysecondmap = make(M)
       info, err := GetTypeInfo(mysecondmap)
       assert.Nil(t, err)
       assert.Len(t, info.TagToField, 0)
```
If we retrieve data from the cache, the last `assert.Len(t, info.TagToField, 0)` will fail. It will report length = 2, instead of the expected 0.

typeinfo/typeinfo.go
Aflynn50: Ah ok that makes sense. I had a different idea in my head of how the cache would be used (I was thinking it would be limited to a single query for caching info from prepare to use in scan, rather than global). For this use case a local map in prepare would make more sense.

niemeyer: The cache can't be limited to a single query if it sits in a global variable as is being done in this PR. The original idea was to have the type reference in the prepared statement, as you describe, and the actual type information cached globally, since types are unique.

typeinfo/typeinfo.go
niemeyer: What tags and field names will be cached for `M`? What tags and field names will be cached for other maps?

fernape: The plan for `M` is not to cache anything. We don't support other maps (as in generic maps) AFAIK. There is no mention to that in the specification. Either `structs` or `sqlair.M` with `M` being of underlying type `map[string]any`. Is the counter example above ill formed?

typeinfo/typeinfo.go
Aflynn50: Nice, makes sense.

typeinfo/typeinfo.go
fernape: I changed it so we don't try to store key information whatsoever. I would not be necessary according to a [concept](https://github.com/fernape/sqlairexperiment/blob/minimal-reflection/typeinfo/info.go#L40) we've been working on.

typeinfo/typeinfo.go
fernape: No `InfoFromName` function and the cache is back to use `reflect.Type` as key.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+)
+
+var cacheMutex sync.RWMutex
+var cache = make(map[reflect.Type]Info)
```
niemeyer: I believe this should be a pointer, as in the go-yaml package that we're basing this logic off of, so that we're not copying this data around on every access of the map.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Reflect will return the Info of a given type,
+// generating and caching as required.
+func GetTypeInfo(value any) (Info, error) {
```
niemeyer: Same here: the result should be a pointer.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+// reflect.Value that is specifically required for SQLAir operation.
+func generate(value reflect.Value) (Info, error) {
+	// Dereference the pointer if it is one.
+	value = reflect.Indirect(value)
+
+	// Reflection information is generated for structs, typeinfo.M
+	// and plain types only.
+	if value.Kind() != reflect.Struct {
+		if value.Kind() == reflect.Map && value.Type().Name() != "M" {
+			return Info{}, fmt.Errorf("Can't reflect map type")
```
niemeyer: Couple of nitpicks from former reviews and conversations:

- Our errors are always lower case, so that they can be composed nicely
- We try to use "cannot" instead of "can't", "could not", "won't", "unable to", etc

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+		} else {
+			return Info{Type: value.Type()}, nil
```
niemeyer: Per prior review on this PR, I don't think we should be handling maps here, M or otherwise.  This functionality is about extracting tag information from structs, so that we can tell which fields map to which columns and vice versa. The map handling by definition needs to be done in a very different way. If we use the struct type info in a map, we have a bug.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+
+	info := Info{
+		TagToField: make(map[string]Field),
+		FieldToTag: make(map[string]string),
+		Type:       value.Type(),
+	}
+
+	// If we reach this point, this is a reflect.Struct:
```
niemeyer: This whole function only makes sense for structs, I believe.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+			Index:     i,
+			OmitEmpty: omitEmpty,
+			Type:      reflect.TypeOf(value.Field(i).Interface()),
+		}
+		info.FieldToTag[field.Name] = tag
+	}
+
+	return info, nil
+
+	return info, nil
```
niemeyer: Some left over here. Btw, it's a very nice practice to do a self-review on changes before proposing them for merging onto the upstream, as this sort of thing is normal when we go back and forth on code for a bit.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,105 @@
...
+
+	var omitEmpty bool
+	if len(options) > 1 {
+		if strings.ToLower(options[1]) != "omitempty" {
+			return "", false, fmt.Errorf("unexpected tag value %q", options[1])
+		}
+		omitEmpty = true
+	}
+
+	return options[0], omitEmpty, nil
```
niemeyer: This function seems to ignore empty and invalid names in general.

niemeyer (changes requested):
Thanks. Another pass:

internal/typeinfo/typeinfo.go
fernape: Changed.

internal/typeinfo/typeinfo.go
fernape: Done. This code has been removed.

internal/typeinfo/typeinfo.go
fernape: Changed as with the previous comment. The function now only handles structs. Tests were also changed accordingly.

internal/typeinfo/typeinfo.go
fernape: Removed. Sorry for that.

internal/typeinfo/typeinfo.go
fernape: Strengthen the `parseTag`  function. Check for empty and invalid tags. Added some tests accordingly.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+		omitEmpty = true
+	}
+
+	name := options[0]
+	if len(name) == 0 {
+		return "", false, fmt.Errorf("empty db tag")
+	}
+
+	if !validColNameRx.MatchString(name) {
+		return "", false, fmt.Errorf("invalid column name in 'db' tag")
```
niemeyer: In such errors it's useful to know what went wrong. Something along these lines:
```
fmt.Errorf("field %s.%s defines an invalid 'db' column name: %q",  typeName, columName, name)
``` 
Same is true for most of the errors in the PR, so it's worth doing a pass over these.

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,161 @@
...
+		NotInDB string
+	}
+
+	s := something{
+		ID:      99,
+		Name:    "Chainheart Machine",
+		NotInDB: "doesn't matter",
+	}
+
+	info, err := GetTypeInfo(s)
```
niemeyer: In Go in general we avoid the 'Get' in such APIs. In go-yaml I added 'get' to the method, likely as a cheap way to avoid the conflict with the type name, but that API is private and somewhat irrelevant. Here it doesn't look like the conflict exists.

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,161 @@
...
+	assert.Nil(t, err)
+
+	assert.Equal(t, reflect.Struct, info.Type.Kind())
+	assert.Equal(t, "something", info.Type.Name())
```
niemeyer: The info.Type may be compared to reflect.TypeOf(s) directly as these are unique and comparable.

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,161 @@
...
+func TestReflectBadTagError(t *testing.T) {
+	{
+		type s1 struct {
+			ID int64 `db:"id,bad-juju"`
+		}
+		ss := s1{ID: 99}
+		_, err := GetTypeInfo(ss)
+		assert.Equal(t, fmt.Errorf(`unexpected tag value "bad-juju"`), err)
+	}
+	{
```
niemeyer: I think we talked about avoiding these sub-branches in a different review point. Can we please avoid them as they artificially space-out the testing without adding much value?  If these are independent tests, then they shoudl be different functions. If they are the same test, then they can just be normal code, spaced out appropriately.

For the particular case at hand, this sort of test is better organized in a table-styled test. For example, pretty much all of them try to get the type info of some value, and return an error. So you may have something like this:
```
type tagErrorTest struct {
    value any
    error string
}

var tagErrorTests = []tagErrorTest{{
    value: &struct{ ID int64 `db:"id,bad-juju"` } {99}
    error: "error message,
}, {
    ...
}}

func ... {
    for _, test := range tagErrorTests {
    }
}
```

In other words, the repetitive part of the test is extracted out from the key logic, so the latter may be done only once.

These points also apply for the former test case.

internal/typeinfo/info.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// Relate tag names to fields.
+	TagToField map[string]Field
+
+	// Relate field names to tags.
+	FieldToTag map[string]string
+}
+
+// SQLAir's specific map type.
+type M map[string]any
```
niemeyer: This can go now, I believe. There's no use for it inside the package for the time being.

niemeyer (changes requested):
Thanks for the improvements, Fernando.  The logic looks good. Just some general conventional sync around the tests and this should be ready to go in. This is not the kind of thing I'll be bothering you much about in the future, but it's nice to get into the same page early on so we can be on the same page and tend to create code that looks the same across projects.

One side point that we should talk about later is to migrate the code to go-check instead of using the assert package, but let's keep that out for now so we can get things merged and moving on.

internal/typeinfo/typeinfo.go
fernape: Error messages changed to be more informative.

internal/typeinfo/typeinfo_test.go
fernape: Changed.

internal/typeinfo/typeinfo_test.go
fernape: Done.

internal/typeinfo/typeinfo_test.go
fernape: Changed where applicable.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+var cacheMutex sync.RWMutex
+var cache = make(map[reflect.Type]*Info)
+
+// Reflect will return the Info of a given type,
+// generating and caching as required.
+func TypeInfo(value any) (*Info, error) {
+	if value == (any)(nil) {
+		return &Info{}, fmt.Errorf("cannot reflect nil value")
```
niemeyer: On errors we pretty much always return the zero value, nil for pointers, unless there are good reasons not to. This is one more way to prevent the error from being ignored, and also prevents a needless allocation as that value is broken and really shouldn't be used.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+	cacheMutex.Unlock()
+
+	return info, nil
+}
+
+// generate produces and returns reflection information for the input
+// reflect.Value that is specifically required for SQLAir operation.
+func generate(value reflect.Value) (*Info, error) {
+	// Dereference the value if it is a pointer.
+	value = reflect.Indirect(value)
```
niemeyer: This was already done by the function that calls this function. In go-yaml, we don't do either of those because the method expects an actual struct, since there's already handling for type-related information before reaching the point where we decide to call this function, and as we covered a few times, this is only meant to be used in structs, so whatever is calling it needs to know somehow.  It's okay to use a different convention, of course, but as it is right now it feels like there's no clear behavior defined, and we just call it everywhere just in case. What's the overall idea?

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Reflection information is only generated for structs.
+	if value.Kind() != reflect.Struct {
+		return &Info{}, fmt.Errorf("cannot reflect type %q, only struct", value.Kind())
```
niemeyer: Same note about zero values on error above, and the error message is also not very clear, as we certainly can reflect on any type we want in Go.  This error should say something like: "internal error: attempted to obtain struct information for something that is not a struct: %T.   Also, the parameter is a Kind, not a Type, and it would indeed be more useful to have the type name here so that it's easier for the developer to find the offending code.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		// Fields without a "db" tag are outside of SQLAir's remit.
+		tag := field.Tag.Get("db")
+		if tag == "" {
+			continue
+		}
+		tag, omitEmpty, err := parseTag(tag)
+		if err != nil {
+			return &Info{}, fmt.Errorf("cannot parse tag for field %s.%s: %s", typ.Name(), field.Name, err)
```
niemeyer: Same about zero value on errors.

internal/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,115 @@
...
+	// Refuse to parse if there are more than 2 items.
+	if len(options) > 2 {
+		return "", false, fmt.Errorf("too many options in 'db' tag: %s", strings.Join(options, ", "))
+	}
+	if len(options) == 2 {
+		if strings.ToLower(options[1]) != "omitempty" {
+			return "", false, fmt.Errorf("unexpected tag value %q", options[1])
+		}
+		omitEmpty = true
+	}
```
niemeyer, line 98: Let's please use logic similar to go-yaml here, as we don't really care about how many options we have, but rather about whether the particular flag is supported or not.
```
		inline := false
		fields := strings.Split(tag, ",")
		if len(fields) > 1 {
			for _, flag := range fields[1:] {
				switch flag {
				case "omitempty":
					info.OmitEmpty = true
				case "flow":
					info.Flow = true
				case "inline":
					inline = true
				default:
					return nil, errors.New(fmt.Sprintf("unsupported flag %q in tag %q of type %s", flag, tag, st))
				}
			}
			tag = fields[0]
		}
```

niemeyer (changes requested):
Thanks. Another pass, and probably the last one.

internal/typeinfo/typeinfo.go
fernape: Done everywhere. Thanks.

internal/typeinfo/typeinfo.go
fernape: That's a remnant of when we allowed other types including references (we had `&struct{...}`) in the code for instance. Changed. It is the callers responsibility to know what to pass to the function.

internal/typeinfo/typeinfo.go
fernape: Using type name now.

internal/typeinfo/typeinfo.go
fernape: Done. Notice that something like this `id_,omitempty,omitempty,omitempty` would be allowed though.

niemeyer: That looks awful, but otherwise innocuous. The outcome does exactly what it says on the tin. On the plus side, the suggested logic will tell what flag is wrong, and is extensible to any number of flags.

niemeyer (approved):
Thanks for the changes for keeping up with it.

PR #13, opened 2022-11-04
Move internal tests to seperate file

Aflynn50:
Cleaning up the main branch by moving internal tests to a separate file.


niemeyer (changes requested):
This PR doesn't seem to do what says on the tin.

Maybe something got it unintendedly with it being hanging for a few weeks?

PR #14, opened 2022-11-14
Add fuzzing function for Parse()

fernape:
This function will receive inputs with a high variability. The fuzzer should help us catch any problems early as we complete the parser.

Run with `go test --fuzz=Fuzz`

### DIFF
--- a/internal/parse/parser_test.go
+++ b/internal/parse/parser_test.go
@@ -195,6 +195,7 @@ func TestRound(t *testing.T) {
 	}
 }
 
+// We return a proper error when we find an unbound string literal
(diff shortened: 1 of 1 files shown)


niemeyer (approved):
For other reviewers, using "Hide whitespace" will show that most of the PR is just moving the tests around without changing them.

LGTM, thanks for the fuzzing.

manadart:
@Aflynn50 Prefer rebasing in-flight branches over merging the mainline back in. It makes for a cleaner commit history.

niemeyer:
Thanks for fixing it, Alastair.

PR #15, opened 2022-11-28
Make QueryPart public

Aflynn50:
QueryPart is used in ParsedExpr, which is exported. QueryPart is the only field in the struct and will be used in the subsequent stages so should be public too.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -14,7 +14,7 @@ type Parser struct {
 	// partStart is the value of pos just before we started parsing the part
 	// under pos. We maintain partStart >= prevPart.
 	partStart int
-	parts     []queryPart
+	parts     []QueryPart
 }
 
 func NewParser() *Parser {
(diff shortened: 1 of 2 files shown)


fernape:
LGTM

internal/parse/parser.go
```diff
@@ -69,25 +69,24 @@ const (
...
 )
 
-// ParsedExpr is the AST representation of an SQL expression.
-// It has a representation of the original SQL statement in terms of queryParts
-// A SQL statement like this:
+// ParsedExpr is the AST representation of an SQL expression. The AST is made up
+// of QueryParts. For example, a SQL statement like this:
 //
-// Select p.* as &Person.* from person where p.name = $Boss.Name
+// Select p.* as &Person.* from person where p.name = $Boss.col_name
```
TheSignPainter98, line 68: Should this `s/Name/col_name` be in a separate PR?

Aflynn50: Its a comment that was missed in a previous PR that pertains to the QueryParts so here is probably the best place for it

niemeyer (approved):
As we chatted, this sounds okay for now, but it's a good practice to have a nice separation between the public interface of a pacakge (or library in general) and its internal APIs which is how it does it work to expose that public API. It's gradually feeling like quite a bit of the internal packages are becoming public details, which while okay because it's all under an internal module and thus protected from external use, makes me wonder if we have a clear notion of what is the purpose of these packages and how that interacts with others.

In either case, per our conversation, I'm just trying to plant a seed here so that such decisions become conscious.

PR #16, opened 2022-11-28
Add advance function to Parse method

Aflynn50:
This PR removed the p.pos++ from the Parse function and replaces it with a function advance that skips input until it finds the next character of interest. These are the characters that appear at the start of IO expressions and quoted string literals.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -44,15 +44,11 @@ var tests = []struct {
 }, {
 	"multiple-quoted bypass expression",
 	`SELECT '''' AS &Person.*`,
-	`[Bypass[SELECT ] Bypass[''''] Bypass[ AS &Person.*]]`,
+	`[Bypass[SELECT ] Bypass[''''] Bypass[ AS ] Output[[] [Person.*]]]`,
 }, {
 	"single quote in double quotes",
 	`SELECT "'" AS &Person.*`,
-	`[Bypass[SELECT ] Bypass["'"] Bypass[ AS &Person.*]]`,
-}, {
-	"dollar without preceding space",
-	"SELECT p.* AS&Person.*",
-	"[Bypass[SELECT p.* AS&Person.*]]",
+	`[Bypass[SELECT ] Bypass["'"] Bypass[ AS ] Output[[] [Person.*]]]`,
 }, {
 	"quoted output expression",
 	"SELECT p.* AS &Person.*, '&notAnOutputExpresion.*' AS literal FROM t",
@@ -288,8 +284,7 @@ var tests = []struct {
 	"input with no space",
 	"SELECT p.*, a.district " +
 		"FROM person AS p WHERE p.name=$Person.name",
-	"[Bypass[SELECT p.*, a.district FROM person AS p WHERE p.name=] " +
-		"Input[Person.name]]",
+	"[Bypass[SELECT p.*, a.district FROM person AS p WHERE p.name=$Person.name]]",
 }, {
 	"escaped double quote",
 	`SELECT foo FROM t WHERE t.p = "Jimmy ""Quickfingers"" Jones"`,
@@ -304,7 +299,7 @@ var tests = []struct {
 	"complex escaped quotes",
 	`SELECT * AS &Person.* FROM person WHERE ` +
 		`name IN ('Lorn', 'Onos T''oolan', '', ''' ''');`,
-	`[Bypass[SELECT * AS &Person.* FROM person WHERE name IN (] ` +
+	`[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[ FROM person WHERE name IN (] ` +
 		`Bypass['Lorn'] Bypass[, ] Bypass['Onos T''oolan'] ` +
 		`Bypass[, ] Bypass[''] Bypass[, ] Bypass[''' '''] ` +
 		`Bypass[);]]`,
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -160,14 +160,14 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 func (p *Parser) advance() {
 
 	// These bytes might be the start of an expression.
-	initialBytes := map[byte]bool{
+	stopBytes := map[byte]bool{
 		'$':  true,
 		'"':  true,
 		'\'': true,
 		'(':  true,
 	}
 
-	// The byte following these bytes might the start of an expression.
+	// The byte following these bytes might be the start of an expression.
 	delimiterBytes := map[byte]bool{
 		' ':  true,
 		'\t': true,
@@ -178,7 +178,7 @@ func (p *Parser) advance() {
 	}
 
 	p.pos++
-	for p.pos < len(p.input) && !initialBytes[p.input[p.pos]] &&
+	for p.pos < len(p.input) && !stopBytes[p.input[p.pos]] &&
 		!delimiterBytes[p.input[p.pos]] {
 		p.pos++
 	}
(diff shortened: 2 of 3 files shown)


internal/parse/parser.go
```diff
@@ -118,15 +119,15 @@ func (p *Parser) add(part queryPart) {
 }
 
 // Parse takes an input string and parses the input and output parts. It returns
-// a pointer to a ParsedExpr.
+// a pointer to a ParsedExpr. If the parser encounters an error then ParsedExpr
+// is nil.
```
fernape: We could write this the same way everywhere, like:
```
// Returns this and that on success.
// Returns this and nil otherwise.
```

internal/parse/parser.go
```diff
@@ -118,15 +119,15 @@ func (p *Parser) add(part queryPart) {
...
 func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("cannot parse expression: %s", err)
 		}
 	}()
 	p.init(input)
-
```
TheSignPainter98: A little breathing room is good, perhaps a blank line above instead would look nice?

internal/parse/parser.go
Aflynn50: Sounds good, though might be clearer to write:
// On success returns this and that
// Otherwise returns this and nil
To make it slightly faster to parse when you're looking for what the function should return

internal/parse/parser.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 )
 
+// Parser keeps track of the current parsing state.
```
niemeyer: This is not so helpful, in the sense that every struct type is there to keep track of its internal state, by definition.  As a rule I use for myself and recommend in general, we should only write comments when they are not obvious from context. This has two advantages: first, we don't waste people's time reading text that is not going to add any useful information, and second, we become more critical of our own comments so that they are indeed relevant and helpful. For example, there are probably good things to be said about what is the role of the Parser in this module.

internal/parse/parser.go
```diff
@@ -111,15 +112,16 @@ func (p *Parser) add(part queryPart) {
 }
 
 // Parse takes an input string and parses the input and output parts. It returns
-// a pointer to a ParsedExpr.
+// a pointer to a ParsedExpr. If the parser encounters an error then ParsedExpr
```
niemeyer: The change here also violates that rule: it's a standard practice in Go to return zero values every time there's an error. We don't need to add that information to every comment.

internal/parse/parser.go
```diff
@@ -142,13 +144,38 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+		p.advance()
 	}
 	// Add any remaining unparsed string input to the parser.
 	p.add(nil)
 	return &ParsedExpr{p.parts}, nil
 }
 
+// advance increments p.pos until we reach a space, a $, a " or a '. If we find
+// a space we advance to the rightmost adjacent space so the p.pos points to the
+// last space before the next non-space character.
```
niemeyer: Similarly here, the documentation is copying the code in English form, which is also an anti-pattern. Instead, the best comments provide context without repeating the code. For example:
```
// advance moves p.pos forward until it reaches content that may be the start of known tokens
// that must be taken into account by the parser.
```
That will indeed save the next reader the trouble of implying why this function was created, while allowing them to read the code if they would like to know details of that process.

To be clear, I'm nitpicking more than usual here to attempt to establish some shared baselines. I won't be bothering you for too long about this. :)

internal/parse/parser.go
```diff
@@ -142,13 +144,38 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+func (p *Parser) advance() {
+	noteableBytes := map[byte]bool{
+		'$':  true,
+		'"':  true,
+		'\'': true,
+		' ':  true,
+	}
+	// Advance the parser to the next char.
```
niemeyer: Same.. that's literally what `p.pos++` means. The English version of that is much longer. :-)

I'd argue the same about the remaining ones, but please feel free to decide whether to keep or not.

internal/parse/parser.go
```diff
@@ -142,13 +144,38 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+	for p.pos < len(p.input) && !noteableBytes[p.input[p.pos]] {
+		p.pos++
+	}
+	// If we are on a space, advance to the rightmost in the sequence.
+	if p.peekByte(' ') {
+		for p.pos+1 < len(p.input) && p.input[p.pos+1] == ' ' {
+			p.pos++
+		}
+	}
+	return
```
niemeyer: We conventinoally avoid bare returns in Go except in very niche scenarios.

internal/parse/parser.go
```diff
@@ -141,13 +143,38 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+	}
+	// Advance the parser to the next char.
+	p.pos++
+	// Keep incrementing pos until we find a character of interest.
+	for p.pos < len(p.input) && !noteableBytes[p.input[p.pos]] {
+		p.pos++
+	}
+	// If we are on a space, advance to the rightmost in the sequence.
+	if p.peekByte(' ') {
+		for p.pos+1 < len(p.input) && p.input[p.pos+1] == ' ' {
```
niemeyer: So, about the actual behavior: this skipping towards any "notable" char, such as `$`, even if it hasn't been preceded by something that is a real separator. For example, this would advance to the `$`  in `asd$asd`, even though this isn't valid syntax. Also, the opposite case is interesting: the current logic stops at spaces, even if the next character isn't something that could be relevant to the expression language. So, I guess the question is: what logic should we implement here?  This seems close, but doesn't seem to be quite there yet.

niemeyer (changes requested):
This looks clean, and like a good direction. I have one relevant question about the actual algorithm, and many tiny nitpicks about comment conventions.

internal/parse/parser.go
Aflynn50: As the [spec](https://docs.google.com/document/d/11WUBE5Ik5t113pWzMq78H8nXT_Rv_YIawDZpziSV0_k/edit#heading=h.2bm4jvbrvoef) stands, the part of `asd$asd` following the `$` is considered as an input expression (though not a well formed one). We don't require that there is a space before the `$`. The thinking being that a user may well write `p=$Person.name` or `($Person.name)`.

We could change the input parser so that we only recognise it as an input expression if `$` is followed by a capital letter and/or the type is qualified by a column (though this would be a separate PR).

We stop at spaces because we have to watch out for the columns which come before the expression e.g. `col1 AS &P.col1` or `(col1, col2) AS &P.*. We could skip to the next space that is followed by an alphanumeric char but since that includes nearly everything likely to follow a space it would not be much of a speed up.

An alternative might be to skip the next `&` and somehow look backwards for the columns, though this would require a rewrite of #23.

internal/parse/parser.go
Aflynn50: Noted, comment removed.

internal/parse/parser.go
Aflynn50: Removed part of comment about error.

internal/parse/parser.go

internal/parse/parser.go

internal/parse/parser.go
Aflynn50: Removed.

internal/parse/parser.go
niemeyer: For completeness so we can close this point, since we discussed this a while ago in our calls: the specification doesn't have a precise syntax definition for the syntax, and thus is prone to interpretation and discussion during implementation. With that said, assuming that "foo$asd" is a valid expression when the spec was talking about $Type.field seems like an unresonable stretch.

PR #17, opened 2022-11-28
Handle escaped quotes

Aflynn50:
Modified parseStringLiteral to ignore escaped quotes and added testing for it.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -153,6 +153,12 @@ func (p *Parser) peekByte(b byte) bool {
 	return p.pos < len(p.input) && p.input[p.pos] == b
 }
 
+// peekByteAt returns true if the byte at i equals the one passed as a
+// parameter.
+func (p *Parser) peekByteAt(b byte, i int) bool {
+	return i >= 0 && i < len(p.input) && p.input[i] == b
+}
+
 // skipByte jumps over the current byte if it matches the byte passed as a
 // parameter. Returns true in that case, false otherwise.
 func (p *Parser) skipByte(b byte) bool {
@@ -230,6 +236,39 @@ func (p *Parser) skipName() bool {
 		}
 	}
 	return p.pos > mark
+}
+
+// isEscaped checks if the byte at p.pos is escaped.
+func (p *Parser) isEscaped() bool {
+	return p.peekByteAt('\\', p.pos-1) && !p.peekByteAt('\\', p.pos-2)
+}
+
+// skipCharFind advances the parser until it finds the char passed as a
+// parameter then jumps over it and returns true. If the parameter is escaped
+// the it is ignored.  If the end of the string is reached and no matching byte
+// was found, it returns false.
+func (p *Parser) skipCharFind(c byte) bool {
+	mark := p.pos
+	for p.pos < len(p.input) {
+		if p.peekByte(c) && !p.isEscaped() {
+			p.pos++
+			return true
+		}
+		p.pos++
+	}
+	p.pos = mark
+	return false
+}
+
+// skipChar jumps over the current char if it matches the byte passed as a
+// parameter. If the char is escaped it is ignored. Returns true in that case,
+// false otherwise.
+func (p *Parser) skipChar(c byte) bool {
+	if (p.pos < len(p.input)) && (p.input[p.pos] == c) && (!p.isEscaped()) {
+		p.pos++
+		return true
+	}
+	return false
 }
 
 // Functions with the prefix parse attempt to parse some construct. They return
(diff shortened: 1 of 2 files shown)


internal/parse/parser.go
```diff
@@ -300,20 +300,22 @@ func (p *Parser) parseInputExpression() (*InputPart, bool, error) {
 	return nil, false, nil
 }
 
-// parseStringLiteral parses quoted expressions and ignores their content.
+// parseStringLiteral parses quoted expressions and ignores their content
+// including escaped quotes.
```
fernape: Maybe add this:
```
+// Returns a pointer to a BypassPart, true and nil on success.
+// Returns nil, false and an error if an error was found.
+// Returns nil, false and nil if no string literal was found.
```

Aflynn50: I think for changing the comments to this form we could do it all at once in a separate PR so that everything remains  consistent.

internal/parse/parser.go
```diff
@@ -231,6 +237,39 @@ func (p *Parser) skipName() bool {
...
+		p.pos++
+	}
+	p.pos = mark
+	return false
+}
+
+// skipChar jumps over the current char if it matches the byte passed as a
+// parameter. If the char is escaped it is ignored. Returns true in that case,
+// false otherwise.
+func (p *Parser) skipChar(c byte) bool {
```
solmyrcan: Maybe we could call `peekByte()` here, like so:

`
if p.peekByte(c) && (!p.isEscaped()) { `

could also do this at the `skipByte()` method, like so:

```
func (p *Parser) skipByte(b byte) bool {
	if p.peekByte(b) { 
```

Aflynn50: Nice, I've changed skipChar. I'll leave skipByte for now since its not in this PRs remit but its a good point.

manadart:
Double quoted strings are not generally used in SQL.

One would prefer 
```sql 
SELECT 'something' FROM dual
```
over
```sql 
SELECT "something" FROM dual
```

And apostrophes are repeated in order to escape. So "Onos T'oolan" would be retrieved thusly:
```sql
SELECT 'Onos T''oolan' FROM dual
```

In turn, selecting a single quote would be:
```sql
SELECT '''' FROM dual
```

Are we covering all of these?

Aflynn50:
> Double quoted strings are not generally used in SQL.
> 
> One would prefer
> 
> ```sql
> SELECT 'something' FROM dual
> ```
> 
> over
> 
> ```sql
> SELECT "something" FROM dual
> ```
> 
> And apostrophes are repeated in order to escape. So "Onos T'oolan" would be retrieved thusly:
> 
> ```sql
> SELECT 'Onos T''oolan' FROM dual
> ```
> 
> In turn, selecting a single quote would be:
> 
> ```sql
> SELECT '''' FROM dual
> ```
> 
> Are we covering all of these?

> Double quoted strings are not generally used in SQL.
> 
> One would prefer
> 
> ```sql
> SELECT 'something' FROM dual
> ```
> 
> over
> 
> ```sql
> SELECT "something" FROM dual
> ```
> 
> And apostrophes are repeated in order to escape. So "Onos T'oolan" would be retrieved thusly:
> 
> ```sql
> SELECT 'Onos T''oolan' FROM dual
> ```
> 
> In turn, selecting a single quote would be:
> 
> ```sql
> SELECT '''' FROM dual
> ```
> 
> Are we covering all of these?

PR changed to use repeated quote escaping rather than backslash escaping.

internal/parse/parser.go
```diff
@@ -305,23 +305,35 @@ func (p *Parser) parseInputExpression() (*InputPart, bool, error) {
...
+
 		if c == '"' || c == '\'' {
-			p.skipByte(c)
-			// TODO Handle escaping
-			if !p.skipByteFind(c) {
-				// Reached end of string and didn't find the closing quote.
-				return nil, false, fmt.Errorf("missing right quote in string literal")
+			mark := p.pos
+
+			if p.skipByte(c) {
```
niemeyer: Isn't all of the above logic, starting at the very first `if` block, a manual version of `if p.skipByte('"') || p.skipByte('\'')`?

niemeyer (changes requested):
Just one comment about the implementation. Idea looks nice again.

internal/parse/parser.go
Aflynn50: You're right, I think I misunderstood the flow of the code when we talked in the call. I've changed it.

internal/expr/parser.go
```diff
@@ -305,23 +305,29 @@ func (p *Parser) parseInputExpression() (*inputPart, bool, error) {
...
+			if maybeCloser && !p.peekByte(c) {
+				return &bypassPart{p.input[mark:p.pos]}, true, nil
 			}
-			return &bypassPart{p.input[cp.pos:p.pos]}, true, nil
+			maybeCloser = !maybeCloser
 		}
-	}
 
+		// Reached end of string and didn't find the closing quote
+		return nil, false, fmt.Errorf("missing right quote for char %d in string literal", cp.pos)
```
niemeyer: Thanks for the changes. Only one trivial: can we please update this to use the same `column %d:` prefix style we discussed elsewhere?

niemeyer (approved):

internal/expr/parser.go

PR #18, opened 2022-11-28
Modify skipName to not accept numerals as the start of a name

Aflynn50:
This PR modifies skipName to not accept numerals as the start of a name.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -203,24 +203,33 @@ func (p *Parser) skipString(s string) bool {
 	return false
 }
 
-// isNameByte returns true if the byte passed as parameter is considered to be
-// one that can be part of a name. It returns false otherwise.
+// isNameByte returns true if the byte c can be part of a name. It returns false
+// otherwise.
 func isNameByte(c byte) bool {
 	return 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' ||
 		'0' <= c && c <= '9' || c == '_'
 }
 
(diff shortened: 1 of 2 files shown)


internal/parse/parser.go
```diff
@@ -211,24 +211,33 @@ func (p *Parser) skipString(s string) bool {
 	return false
 }
 
-// isNameByte returns true if the byte passed as parameter is considered to be
-// one that can be part of a name. It returns false otherwise.
+// isNameByte returns true if the byte c can be part of a name. It returns false
+// otherwise.
```
TheSignPainter98: How about 'the given byte'?

internal/parse/parser.go
```diff
@@ -211,24 +211,33 @@ func (p *Parser) skipString(s string) bool {
...
 func isNameByte(c byte) bool {
 	return 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' ||
 		'0' <= c && c <= '9' || c == '_'
 }
 
-// skipName returns false if the parser is not on a name. Otherwise it advances
-// the parser until it is on the first non name byte and returns true.
+// isNameInitialByte returns true if the byte c can appear at the start of a
+// name. It returns false otherwise.
```
TheSignPainter98, line 214: Same here perhaps

internal/parse/parser.go

internal/parse/parser.go
Aflynn50: Changed.

niemeyer (approved):
LGTM, thanks.

PR #19, opened 2022-11-28
Replace parseFullName with parseGoObject

Aflynn50:
This PR replaces parseFullName with the more specific parseGoObject. Previously, parseFullName was intended for use parsing columns and go objects but we decided it made more sense to have two separate functions.

By using parseGoObject we can also be more specific with how we parse input expressions. We will parse an input expression as long as the character after the $ is a name byte. We ignore the character before the $.

The errors returned by parseInputExpression also now include an approximate position of the error. We intend to provide an approximate position for all errors in future PRs.

### DIFF
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -244,7 +244,8 @@ const (
 )
 
 // parseIdentifier parses either a name made up only of nameBytes (or an
-// asterisk if asteriskF=allowAsterisk)
+// asterisk if asteriskF=allowAsterisk). On success it returns the parsed string
+// and true. Otherwise, it returns the empty string and false.
 func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
 	if asteriskF == allowAsterisk {
 		if p.skipByte('*') {
@@ -260,14 +261,16 @@ func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
 
 // parseGoObject parses a source or target go object of the form Prefix.Name
 // where the type is the Prefix and the field is the Name (this applies to maps
-// and structs).
+// and structs). On sccuess it returns the parsed FullName, true and nil.
+// If it cannot parse the Go object it returns false. If the construct was
+// recognised as a Go object there will be an error. Otherwise err=nil.
 func (p *Parser) parseGoObject() (FullName, bool, error) {
 	cp := p.save()
 	var fn FullName
 	if id, ok := p.parseIdentifier(disallowAsterisk); ok {
 		if p.skipByte('.') {
 			if idField, ok := p.parseIdentifier(allowAsterisk); ok {
-				return FullName{Prefix: id, Name: idField}, true, nil
+				return FullName{id, idField}, true, nil
 			}
 			return fn, false, fmt.Errorf("invalid identifier near char %d", p.pos)
 		}
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -243,9 +243,9 @@ const (
 	disallowAsterisk
 )
 
-// parseIdentifier parses either a name made up only of nameBytes (or an
-// asterisk if asteriskF=allowAsterisk). On success it returns the parsed string
-// and true. Otherwise, it returns the empty string and false.
+// parseIdentifier parses a name made up of only nameBytes, or alternatively a
+// single asterisk if `allowAsterisk` is passed. On success it returns the
+// parsed string and true. Otherwise, it returns the empty string and false.
 func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
 	if asteriskF == allowAsterisk {
 		if p.skipByte('*') {
@@ -261,9 +261,9 @@ func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
 
 // parseGoObject parses a source or target go object of the form Prefix.Name
 // where the type is the Prefix and the field is the Name (this applies to maps
-// and structs). On sccuess it returns the parsed FullName, true and nil.
-// If it cannot parse the Go object it returns false. If the construct was
-// recognised as a Go object there will be an error. Otherwise err=nil.
+// and structs). On success it returns the parsed FullName, true and nil. If a
+// Go object is found, but not formatted correctly, false and an error are
+// returned. Otherwise the error is nil.
 func (p *Parser) parseGoObject() (FullName, bool, error) {
 	cp := p.save()
 	var fn FullName
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -236,12 +236,12 @@ func (p *Parser) skipName() bool {
 //  - bool == false, err == nil
 //		The construct was not the one we are looking for
 
-// parseIdentifier parses either a name made up only of nameBytes or an
-// asterisk.
-func (p *Parser) parseIdentifier() (string, bool) {
-	if p.skipByte('*') {
-		return "*", true
-	}
+type asteriskFlag int
+
+const (
+	allowAsterisk asteriskFlag = iota
+	disallowAsterisk
+)
 
 // parseIdentifier parses either a name made up only of nameBytes (or an
 // asterisk if asteriskF=allowAsterisk)
@@ -275,18 +275,6 @@ func (p *Parser) parseGoObject() (FullName, bool, error) {
 	}
 	cp.restore()
 	return fn, false, nil
-}
-
-// asteriskCount returns the number of FullNames in the argument with a asterisk in
-// the Name field.
-func asteriskCount(fns []FullName) int {
-	s := 0
-	for _, fn := range fns {
-		if fn.Name == "*" {
-			s++
-		}
-	}
-	return s
 }
 
 // parseInputExpression parses an input expression of the form $Type.name.
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -281,12 +281,11 @@ func (p *Parser) parseGoObject() (FullName, bool, error) {
 }
 
 // parseInputExpression parses an input expression of the form $Type.name.
-func (p *Parser) parseInputExpression() (ip *InputPart, ok bool, err error) {
+func (p *Parser) parseInputExpression() (*InputPart, bool, error) {
 	cp := p.save()
 
 	if p.skipByte('$') {
-		var fn FullName
-		if fn, ok, err = p.parseGoObject(); ok {
+		if fn, ok, err := p.parseGoObject(); ok {
 			if fn.Name == "*" {
 				return nil, false, fmt.Errorf("asterisk not allowed "+
 					"in expression near %d", p.pos)
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -257,11 +257,11 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
-// parseGoTarget parses a go type name qualified by a tag name (or asterisk) of
-// the form "TypeName.col_name". On success it returns the parsed FullName, true
-// and nil. If a target is found, but not formatted correctly, false and an
-// error are returned. Otherwise the error is nil.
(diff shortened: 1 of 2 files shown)


internal/parse/parser.go
```diff
@@ -250,51 +243,67 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
-				return fn, true
-			}
-		} else {
-			// A column name specified without a table prefix is a name not a
-			// prefix.
-			if idc == columnId {
-				fn.Name = fn.Prefix
-				fn.Prefix = ""
+			if idField, ok := p.parseIdentifier(allowAsterisk); ok {
+				return FullName{Prefix: id, Name: idField}, true, nil
```
fernape: We can use the short form here: `return FullName{id, idField}, true, nil`

internal/parse/parser.go
```diff
@@ -250,51 +243,67 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		if fn.Name == "*" {
+			s++
+		}
+	}
+	return s
 }
 
 // parseInputExpression parses an input expression of the form $Type.name.
-func (p *Parser) parseInputExpression() (*InputPart, bool, error) {
+func (p *Parser) parseInputExpression() (ip *InputPart, ok bool, err error) {
```
fernape: I am not sure about the named return values. `ip` and `ok` are never "used" since the function always composes the `return` statement with hardcoded first and second parameters. Using `:=` in `if fn, ok, err = p.parseGoObject(); ok {` would have the same result. In addition we'd need to remove `var fn FullName`. I don't know out policy about this actually.

Aflynn50: Removed.

internal/parse/parser.go

internal/parse/parser.go
```diff
@@ -243,58 +236,64 @@ func (p *Parser) skipName() bool {
...
-func (p *Parser) parseIdentifier() (string, bool) {
-	if p.skipByte('*') {
-		return "*", true
-	}
+type asteriskFlag int
+
+const (
+	allowAsterisk asteriskFlag = iota
+	disallowAsterisk
+)
```
TheSignPainter98: Instead of `asteriskFlag` being a separate type, could you use a `bool`?

internal/parse/parser.go
```diff
@@ -243,58 +236,64 @@ func (p *Parser) skipName() bool {
...
 
-	idStart := p.pos
+// parseIdentifier parses either a name made up only of nameBytes (or an
+// asterisk if asteriskF=allowAsterisk). On success it returns the parsed string
+// and true. Otherwise, it returns the empty string and false.
```
TheSignPainter98: The mention of `asteriskF` is surprising, also the 'either' dangles if you ignore the brackets. Perhaps something like:
```
parseIdentifier parses a name made up of only nameBytes, or alternatively
a single asterisk if `allowAsterisk` is passed.
```
would be cleaner?

internal/parse/parser.go
```diff
@@ -243,58 +236,64 @@ func (p *Parser) skipName() bool {
...
+func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
+	if asteriskF == allowAsterisk {
+		if p.skipByte('*') {
+			return "*", true
+		}
+	}
+	mark := p.pos
 	if p.skipName() {
-		return p.input[idStart:p.pos], true
+		return p.input[mark:p.pos], true
```
TheSignPainter98, line 255: `idStart` seems clearer to me

internal/parse/parser.go
```diff
@@ -243,58 +236,64 @@ func (p *Parser) skipName() bool {
...
 }
 
-// parseFullName parses a column name or Go field name, optionally dot-prefixed by
-// its table or type name respectively.
-func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+// parseGoObject parses a source or target go object of the form Prefix.Name
+// where the type is the Prefix and the field is the Name (this applies to maps
+// and structs). On sccuess it returns the parsed FullName, true and nil.
+// If it cannot parse the Go object it returns false. If the construct was
+// recognised as a Go object there will be an error. Otherwise err=nil.
```
TheSignPainter98: `s/sccuess/success/`

Also 'if the construct _was_ recognised...' should this be _was not?_ :)
More generally though, I think the last two lines can be simplified to:
```
If it cannot parse the Go object it returns false and an error.
```

Aflynn50: I think this comment wasn't very clear, I've reworded it.
```If a Go object is found, but not formatted correctly, false and an error are returned. Otherwise the error is nil.```

File: internal/parse/parser.go (hunk 5 above)
Comment (Aflynn50) [reply internal/parse/parser.go:255]: It was changed to be consistent with the other skip functions. Hopefully it's still easy to understand.

File: internal/parse/parser.go (hunk 4 above)

File: internal/parse/parser.go (hunk 3 above)
Comment (Aflynn50) [reply internal/parse/parser.go]: Using a bool was considered, but we opted to go for a const to make it clearer in functions where `parseIdentifier` is used. This way it is clear to see that the part before the `.` in a Go object cannot be an asterisk but the part after can. (It would be so nice to have a Haskell style two element type with named members but alas)

Comment (TheSignPainter98) [reply internal/parse/parser.go]: Ah okay I understand, thanks

File: internal/parse/parser.go (hunk 5 above)
Comment (TheSignPainter98) [reply internal/parse/parser.go:255]: I wasn't aware of this context, conforming to the pattern seems like a good shout though

File: internal/parse/parser.go [hunk 7]
```diff
@@ -243,58 +236,64 @@ func (p *Parser) skipName() bool {
...
-				fn.Prefix = ""
-			}
-			return fn, true
+			return fn, false, fmt.Errorf("invalid identifier near char %d", p.pos)
 		}
+		return fn, false, fmt.Errorf("go object near char %d not qualified", p.pos)
 	}
 	cp.restore()
-	return fn, false
+	return fn, false, nil
```
solmyrcan: Could this method return `nil, false, nil` in failure cases instead of `fn, false, nil` because with this change, `fn` seems to be no longer used. We could delete the declaration for `fn` at the top as well :)

Aflynn50: Thats a good point. Go doesnt actually allow us to use nil there because its not a zero value for structs, but I have just changed it to return `FullName{}` instead so I can remove the var declaration.

internal/parse/parser.go
```diff
@@ -243,58 +236,65 @@ func (p *Parser) skipName() bool {
...
+const (
+	allowAsterisk asteriskFlag = iota
+	disallowAsterisk
+)
+
+// parseIdentifier parses a name made up of only nameBytes, or alternatively a
+// single asterisk if `allowAsterisk` is passed. On success it returns the
+// parsed string and true. Otherwise, it returns the empty string and false.
+func (p *Parser) parseIdentifier(asteriskF asteriskFlag) (string, bool) {
+	if asteriskF == allowAsterisk {
```
niemeyer: Instead of having a flag, why not just having two functions:
```
func parseAsteriskIdentifier() (string, bool) {
     if p.skipByte(...) {
         return "*", ...
     }
    return parseIdentifier()
}

func parseIdentifier() (string, bool) {
    return
}
```

Up to you, though.

internal/parse/parser.go
```diff
@@ -243,58 +236,65 @@ func (p *Parser) skipName() bool {
...
 
-// parseFullName parses a column name or Go field name, optionally dot-prefixed by
-// its table or type name respectively.
-func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+// parseGoObject parses a source or target go object of the form Prefix.Name
+// where the type is the Prefix and the field is the Name (this applies to maps
+// and structs). On success it returns the parsed FullName, true and nil. If a
+// Go object is found, but not formatted correctly, false and an error are
+// returned. Otherwise the error is nil.
+func (p *Parser) parseGoObject() (FullName, bool, error) {
```
niemeyer: This looks fine, but wasn't the point of having a parseFullName that "database.column" and "Type.db_field" look the same?

internal/parse/parser.go
```diff
@@ -243,58 +236,65 @@ func (p *Parser) skipName() bool {
...
-		return p.input[idStart:p.pos], true
+		return p.input[mark:p.pos], true
 	}
 	return "", false
 }
 
-// parseFullName parses a column name or Go field name, optionally dot-prefixed by
-// its table or type name respectively.
-func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+// parseGoObject parses a source or target go object of the form Prefix.Name
```
niemeyer: It says Prefix.Name, but assuming this remains as parseGoObject, isn't this "TypeName.db_tag_name"?  In other words, the db_tag_name will typically be lower case and not TitleCased because it's supposed to match a column name in the database. If that's true, then the name parseGoObject might also require some thinking as this isn't a "Go object", strictly speaking.

niemeyer (changes requested):
Logic looks reasonable. Question is mainly about details and the background behind these functions.

internal/parse/parser.go
Aflynn50: Changed. We were trying to work out the most idiomatic way to do this but I think separate functions is nicer than what we had.

Aflynn50: Changed. We were trying to work out the most idiomatic way to do this but I think separate functions is nicer than what we had.

internal/parse/parser.go
Aflynn50: If we want to support function columns and columns without table prefixes it makes more sense to have separate functions. It also allows us to give better errors. It would be possible to parse them in one function but it would be hard to follow and rely on a lot of conditionals, even if it was more compact.

internal/parse/parser.go

internal/parse/parser.go
```diff
@@ -243,58 +236,62 @@ func (p *Parser) skipName() bool {
...
 }
 
-// parseFullName parses a column name or Go field name, optionally dot-prefixed by
-// its table or type name respectively.
-func (p *Parser) parseFullName(idc idClass) (FullName, bool) {
+// parseGoTarget parses a go type name qualified by a tag name (or asterisk) of
+// the form "TypeName.col_name". On success it returns the parsed FullName, true
+// and nil. If a target is found, but not formatted correctly, false and an
+// error are returned. Otherwise the error is nil.
+func (p *Parser) parseGoTarget() (FullName, bool, error) {
```
niemeyer: Isn't this both Source and Target?  If so, parseGoFullName might be more appropriate. If not, then it's fine.

internal/parse/parser_test.go
```diff
@@ -191,28 +195,34 @@ func TestRound(t *testing.T) {
...
-	parser := parse.NewParser()
-	_, err := parser.Parse(sql)
-	assert.Equal(t, fmt.Errorf("cannot parse expression: missing right quote in string literal"), err)
-}
+	var testTable = []struct {
+		name string
+		sql  string
+		err  string
+	}{
+		{
```
niemeyer: The two parenthesis above can be merged together. That's below as well on the ending (line 217).

This is the same style used right above in this same diff.

internal/parse/parser_test.go
```diff
@@ -231,58 +241,36 @@ func TestBadEscaped(t *testing.T) {
...
-	parser := parse.NewParser()
-	_, err := parser.Parse(sql)
-	assert.Equal(t, fmt.Errorf("cannot parse expression: malformed input type"), err)
-}
+	var testTable = []struct {
+		name string
+		sql  string
+		err  string
+	}{
+		{
```
niemeyer: Same here.

niemeyer (approved):
Only nitpicks now. LGTM, thanks.

internal/parse/parser.go

internal/parse/parser_test.go
Aflynn50: Changed.

internal/parse/parser_test.go
Aflynn50: Changed.

PR #20, opened 2022-11-29
Add names to tests

Aflynn50:
Add names to table tests to make debugging easier. Also make tags in the output parts of the SQL lower case and capitalise IN where it was missed before.

### DIFF
--- a/internal/parse/parser_test.go
+++ b/internal/parse/parser_test.go
@@ -31,68 +31,92 @@ type District struct {
 type M map[string]any
 
 var tests = []struct {
+	summary        string
 	input          string
 	expectedParsed string
 }{{
+	"star table as output",
 	"SELECT p.* AS &Person.*",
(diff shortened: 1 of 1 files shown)


internal/parse/parser_test.go
```diff
@@ -29,113 +29,125 @@ type M map[string]any
 
 func TestRound(t *testing.T) {
 	var tests = []struct {
+		name           string
```
niemeyer: The examples below look more like a "summary" than a "name". I frequently do something similar as well,btw... it's very convenient for the organization, so thanks for doing it.

niemeyer (approved):
Nice, thanks. Just one trivial suggestion. LGTM regardless.

internal/parse/parser_test.go
Aflynn50: Changed and rebased :)

PR #21, opened 2022-11-30
Convert to table-driven tests and use go-check.

fernape:
Convert similar tests to table-driven ones.
Use [go-check](http://labix.org/gocheck) as testing framework.

This will likely affect canonical/sqlair#17, canonical/sqlair#19 and canonical/sqlair#20


internal/parse/parts_test.go
```diff
@@ -4,30 +4,37 @@ import (
...
 	i := parse.InputPart{
 		parse.FullName{
 			Prefix: "mytype",
 			Name:   "mytag",
 		},
 	}
-	assert.NotEqual(t, nil, i)
-	assert.Equal(t, "mytype", i.Source.Prefix)
-	assert.Equal(t, "mytag", i.Source.Name)
+	c.Assert(i, Not(IsNil))
```
niemeyer: This is impossible to be nil unless the compiler is broken. Indeed, it's not even a pointer.

As an aside, IIRC we have NotNil, but typically we don't do this because we always test more details about the value than simply it being nil or not. If it was nil, the logic below will break, for instance.

internal/parse/parts_test.go
```diff
@@ -4,30 +4,37 @@ import (
...
 	p := parse.OutputPart{
 		[]parse.FullName{{"mytable", "mycolumn"}},
 		[]parse.FullName{{"mytype", "mytag"}},
 	}
-	assert.NotEqual(t, nil, p)
-	assert.Equal(t, "mytable", p.Source[0].Prefix)
-	assert.Equal(t, "mycolumn", p.Source[0].Name)
-	assert.Equal(t, "mytype", p.Target[0].Prefix)
-	assert.Equal(t, "mytag", p.Target[0].Name)
+	c.Assert(p, Not(IsNil))
```
niemeyer: Same here. It's worth reviewing the rest of the PR for those cases.

niemeyer (changes requested):
Looks good, with only a couple of trivials. THanks for converting the tests over.

internal/parse/parts_test.go
fernape: I think the original test didn't make sense either since this was not a pointer anyway. The rest of the tests looks good to me here and in `parser_test.go`

internal/parse/parts_test.go
fernape: Done.

niemeyer (approved):
Thanks for tuning, and thanks for de-conflicting, Alastair.

PR #22, opened 2022-12-02
Add back internal tests for the skip functions

Aflynn50:
These were previously removed by accident. They have been updated for adding back in.


niemeyer (approved):
Thanks.

PR #23, opened 2022-12-05
Add output parser

Aflynn50:
Add the functions for parsing outputs and updated the tests.
Some of the functions are somewhat similar to each other and could possibly be merged into one however we have elected to keep them separate with the view that with future updates their definitions will diverge.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -327,11 +327,10 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 // parseList takes a parsing function that returns a fullName and parses a
 // bracketed, comma seperated, list.
 func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+	cp := p.save()
 	if !p.skipByte('(') {
 		return nil, false, nil
 	}
-
-	cp := p.save()
 
 	parenPos := p.pos
 	p.skipSpaces()
@@ -366,7 +365,6 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
 // parseColumns parses a list of columns. For lists of more than one column the
 // columns must be enclosed in brackets e.g. "(col1, col2) AS &Person.*".
 func (p *Parser) parseColumns() ([]fullName, bool) {
-	cp := p.save()
 
 	// Case 1: A single column e.g. "p.name".
 	if col, ok, _ := p.parseColumn(); ok {
@@ -378,7 +376,6 @@ func (p *Parser) parseColumns() ([]fullName, bool) {
 		return cols, true
 	}
 
-	cp.restore()
 	return nil, false
 }
 
@@ -387,8 +384,6 @@ func (p *Parser) parseColumns() ([]fullName, bool) {
 // If the ampersand is not preceded by a space and succeeded by a name or
 // opening bracket the targets are not parsed.
 func (p *Parser) parseTargets() ([]fullName, bool, error) {
-	cp := p.save()
-
 	// Case 1: A single target e.g. "&Person.name".
 	if target, ok, err := p.parseTarget(); err != nil {
 		return nil, false, err
@@ -403,7 +398,6 @@ func (p *Parser) parseTargets() ([]fullName, bool, error) {
 		return targets, true, nil
 	}
 
-	cp.restore()
 	return nil, false, nil
 }
 
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -122,17 +122,24 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 	for {
 		p.partStart = p.pos
 
-		if ip, ok, err := p.parseInputExpression(); err != nil {
+		if out, ok, err := p.parseOutputExpression(); err != nil {
 			return nil, err
 		} else if ok {
-			p.add(ip)
+			p.add(out)
 			continue
 		}
 
-		if sp, ok, err := p.parseStringLiteral(); err != nil {
+		if in, ok, err := p.parseInputExpression(); err != nil {
 			return nil, err
 		} else if ok {
-			p.add(sp)
+			p.add(in)
+			continue
+		}
+
+		if bypass, ok, err := p.parseStringLiteral(); err != nil {
+			return nil, err
+		} else if ok {
+			p.add(bypass)
 			continue
 		}
 
@@ -265,12 +272,45 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
+// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// its table name. parseColumn returns an error so that it can be used with
+// parseList.
+func (p *Parser) parseColumn() (fullName, bool, error) {
+	cp := p.save()
+
+	if id, ok := p.parseIdentifierAsterisk(); ok {
+		if id != "*" && p.skipByte('.') {
+			if idCol, ok := p.parseIdentifierAsterisk(); ok {
+				return fullName{prefix: id, name: idCol}, true, nil
+			}
+		} else {
+			// A column name specified without a table prefix should be in name.
+			return fullName{name: id}, true, nil
+		}
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
+func (p *Parser) parseTarget() (fullName, bool, error) {
+	cp := p.save()
+
+	if p.skipByte('&') {
+		return p.parseGoFullName()
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
 // parseGoFullName parses a Go type name qualified by a tag name (or asterisk)
-// of the form "TypeName.col_name". On success it returns the parsed fullName,
+// of the form "&TypeName.col_name". On success it returns the parsed FullName,
 // true and nil. If a Go full name is found, but not formatted correctly, false
 // and an error are returned. Otherwise the error is nil.
 func (p *Parser) parseGoFullName() (fullName, bool, error) {
 	cp := p.save()
+
 	if id, ok := p.parseIdentifier(); ok {
 		if p.skipByte('.') {
 			if idField, ok := p.parseIdentifierAsterisk(); ok {
@@ -282,11 +322,148 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 		return fullName{}, false,
 			fmt.Errorf("go object near char %d not qualified", p.pos)
 	}
+
 	cp.restore()
 	return fullName{}, false, nil
 }
 
-// parseInputExpression parses an input expression of the form $Type.name.
+// parseList takes a parsing function that returns a fullName and parses a
+// bracketed, comma seperated, list.
+func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+	cp := p.save()
+
+	if !p.skipByte('(') {
+		return nil, false, nil
+	}
+
+	parenPos := p.pos
+	p.skipSpaces()
+
+	nextItem := true
+	var objs []fullName
+	for i := 0; nextItem; i++ {
+		if obj, ok, err := parseFn(p); ok {
+			objs = append(objs, obj)
+		} else if err != nil {
+			return nil, false, err
+		} else if i == 0 {
+			// If the first item is not what we are looking for we exit.
+			cp.restore()
+			return nil, false, nil
+		} else {
+			// On subsequent items we return an error.
+			return nil, false, fmt.Errorf("column %d: invalid expression", p.pos)
+		}
+
+		p.skipSpaces()
+		if p.skipByte(')') {
+			return objs, true, nil
+		}
+
+		nextItem = p.skipByte(',')
+		p.skipSpaces()
+	}
+	return nil, false, fmt.Errorf("column %d: missing closing parentheses", parenPos)
+}
+
+// parseColumns parses a list of columns. For lists of more than one column the
+// columns must be enclosed in brackets e.g. "(col1, col2) AS &Person.*".
+func (p *Parser) parseColumns() ([]fullName, bool) {
+	cp := p.save()
+
+	// Case 1: A single column e.g. "p.name".
+	if col, ok, _ := p.parseColumn(); ok {
+		return []fullName{col}, true
+	}
+
+	// Case 2: Multiple columns e.g. "(p.name, p.id)".
+	if cols, ok, _ := p.parseList((*Parser).parseColumn); ok {
+		return cols, true
+	}
+
+	cp.restore()
+	return nil, false
+}
+
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more references Go objects.
+// If the ampersand is not preceded by a space and succeeded by a name or
+// opening bracket the targets are not parsed.
+func (p *Parser) parseTargets() ([]fullName, bool, error) {
+	cp := p.save()
+
+	// Case 1: A single target e.g. "&Person.name".
+	if target, ok, err := p.parseTarget(); ok {
+		return []fullName{target}, true, nil
+	} else if err != nil {
+		return nil, false, err
+	}
+
+	// Case 2: Multiple targets e.g. "(&Person.name, &Person.id)".
+	if targets, ok, err := p.parseList((*Parser).parseTarget); ok {
+		if starCount(targets) > 1 {
+			return nil, false, fmt.Errorf("column %d: more than one asterisk in expression", p.pos)
+		}
+		return targets, true, nil
+	} else if err != nil {
+		return nil, false, err
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+func starCount(fns []fullName) int {
+	s := 0
+	for _, fn := range fns {
+		if fn.name == "*" {
+			s++
+		}
+	}
+	return s
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
+	cp := p.save()
+
+	// Case 1: There are no columns e.g. "&Person.*".
+	if targets, ok, err := p.parseTargets(); ok {
+		return &outputPart{[]fullName{}, targets}, true, nil
+	} else if err != nil {
+		return nil, false, err
+	}
+
+	// Case 2: There are columns e.g. "p.col1 AS &Person.*".
+	if cols, ok := p.parseColumns(); ok {
+		numCols := len(cols)
+		p.skipSpaces()
+		if p.skipString("AS") {
+			p.skipSpaces()
+			if targets, ok, err := p.parseTargets(); ok {
+				numTargets := len(targets)
+				// If the target is not * then check there are equal columns
+				// and targets.
+				if !(numTargets == 1 && targets[0].name == "*") &&
+					numCols != numTargets {
+					return nil, false, fmt.Errorf("mismatched number of "+
+						"cols and targets in expression: %s",
+						p.input[cp.pos:p.pos])
+				}
+
+				return &outputPart{cols, targets}, true, nil
+			} else if err != nil {
+				return nil, false, err
+			}
+		}
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseInputExpression parses an input expression of the form "$Type.name".
 func (p *Parser) parseInputExpression() (*inputPart, bool, error) {
 	cp := p.save()
 
@@ -301,6 +478,7 @@ func (p *Parser) parseInputExpression() (*inputPart, bool, error) {
 			return nil, false, err
 		}
 	}
+
 	cp.restore()
 	return nil, false, nil
 }
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -331,11 +331,11 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
 	}
 
 	parenPos := p.pos
-	p.skipSpaces()
 
 	nextItem := true
 	var objs []fullName
 	for i := 0; nextItem; i++ {
+		p.skipSpaces()
 		if obj, ok, err := parseFn(p); ok {
 			objs = append(objs, obj)
 		} else if err != nil {
@@ -355,7 +355,6 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
 		}
 
 		nextItem = p.skipByte(',')
-		p.skipSpaces()
 	}
 	return nil, false, fmt.Errorf("column %d: missing closing parentheses", parenPos)
 }
@@ -379,8 +378,6 @@ func (p *Parser) parseColumns() ([]fullName, bool) {
 
 // parseTargets parses the part of the output expression following the
 // ampersand. This can be one or more references to Go types.
-// If the ampersand is not preceded by a space or opening bracket and succeeded
-// by a name the targets are not parsed.
 func (p *Parser) parseTargets() ([]fullName, bool, error) {
 	// Case 1: A single target e.g. "&Person.name".
 	if target, ok, err := p.parseTarget(); err != nil {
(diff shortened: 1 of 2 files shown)


internal/parse/parser.go
```diff
@@ -123,6 +123,13 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 	for {
 		p.partStart = p.pos
 
+		if op, ok, err := p.parseOutputExpression(); err != nil {
```
TheSignPainter98: Naming the return value `op` here says 'operation' to me, and given that the function is called 'parseOutputExpression,' perhaps naming this `out` or `outPart` would be more intuitive.

I realise `op` is consistent with `ip`, but I think both names could benefit from change.

internal/parse/parser.go
```diff
@@ -278,6 +304,138 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+			} else if err != nil {
+				return nil, false, err
+			} else {
+				return nil, false, fmt.Errorf("invalid identifier near char %d", p.pos)
+			}
+			p.skipSpaces()
+			if p.skipByte(')') {
+				return objs, true, nil
+			}
+			nextItem = p.skipByte(',')
```
TheSignPainter98: Should you skip spaces before the `','`? Having a space there would be bad formatting, but might not be considered a syntax error

internal/parse/parser.go
```diff
@@ -278,6 +304,138 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more Go objects. If the ampersand is not found
+// or is not preceded by a space and succeeded by a name or opening bracket the
+// targets are not parsed.
+func (p *Parser) parseTargets() ([]FullName, bool, error) {
+	cp := p.save()
+
+	// An '&' must be preceded by a space and succeeded by a name or opening
+	// bracket.
+	if p.skipString(" &") {
```
TheSignPainter98: Same note here about the number of spaces before the `&`, here this would be consistent with `AS`

internal/parse/parser.go
```diff
@@ -278,6 +304,138 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+}
+
+// parseColumns parses text in the SQL query of the form "table.colname". If
+// there is more than one column then the columns must be enclosed in brackets
+// e.g.  "(col1, col2) AS &Person.*".
+func (p *Parser) parseColumns() ([]FullName, bool) {
+	cp := p.save()
+	// We skip a space here to keep consistent with parseTargets which also
+	// consumes one space before the start of the expression.
+	p.skipByte(' ')
```
TheSignPainter98: Should it always be preceded by a single space or by a positive number of them? (As in `skipSpaces` returns `true`)

internal/parse/parser.go
```diff
@@ -257,6 +264,25 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
+// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// its table name. parseColumn returns an error so that it can be used with
+// parseList.
+func (p *Parser) parseColumn() (FullName, bool, error) {
+	cp := p.save()
+	if id, ok := p.parseIdentifierAsterisk(); ok {
```
solmyrcan: I might be wrong, so please check: `parseIdentifierAsterisk()` seems to return `true` if the position is at an asterisk (`*`) character.

_**Case 1:**_ `FullName: *somename`

`id` becomes `*`. The inner if block execution falls over to the `else`, and we get `Fullname: someone`

_**Case 2:**_ `FullName: *.somename`

`skipByte()` in the inner `if` returns `true`, and then we execute `parseIdentifierAsterisk()` again, which would in turn call `parseIdentifier()`.

We get: `Prefix: *, Name:somename` for the `FullName` output. 

**_My question is:_** 
At the inner `if`, can we instead call `parseIdentifier()` directly,rather than `parseIdentifierAsterisk()`? 

Or do we call `parseIdentifierAsterisk()`, because we may get a possible inner `*` as well (something like `FullName: *.*somename`) ?

However, if it is indeed possible to have an an inner `*`, then the second `parseIdentifierAsterisk()` would return `*` as well, skipping the remaining `somename`, and we get `FullName = Prefix: *, Name: *`

This suggests that we probably don't expect an inner `*` - is my interpretation correct? :)

internal/parse/parser.go
```diff
@@ -278,6 +304,138 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+		if fn.Name == "*" {
+			s++
+		}
+	}
+	return s
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (op *OutputPart, ok bool, err error) {
```
TheSignPainter98: If you choose to change `op` to something else, it should be updated here also :)

internal/parse/parser.go
Aflynn50: We do, on line 327

internal/parse/parser.go
Aflynn50: That is true, ip is almost as confusing as op

internal/parse/parser.go
```diff
@@ -278,6 +304,138 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+		for nextItem {
+			if obj, ok, err := parseFn(p); ok {
+				objs = append(objs, obj)
+				p.skipSpaces()
+			} else if err != nil {
+				return []FullName{}, false, err
+			} else {
+				return []FullName{}, false, fmt.Errorf("invalid identifier near char %d", p.pos)
+			}
+			p.skipSpaces()
```
solmyrcan: Here, `p.skipSpaces()` executes for a second time I think, is this intentional?

internal/parse/parser.go
Aflynn50: Ah yea, so we have in the spec that & needs to be preceded by a space. If there is more than one space that isnt a problem since we advance the parse one step at a time (i.e. the char before that space can also be a space so there can be as many as the user wants) and when PR #16 is added it will respect this as well.

internal/parse/parser.go
Aflynn50: This is just to keep consistant with parseTargets, since it is at the start of the expression it can have any number of spaces before it, they are just not captured as part of the expression.

internal/parse/parser.go
Aflynn50: Done.

internal/parse/parser.go
Aflynn50: So, what we are trying to capture with this function is that a column can be: `*`, `col`, `t.col` or `t.*`. If we don't encounter a `.` then we want to put the name in the `Name` field. If we counter a `.` we want to put the first thing in the `Prefix` field and the second in the `Name` field.

There is a problem in that it is possible to have `*.*` which wouldn't be valid. But I as far as I know in SQL (I may very well be wrong) you don't get columns of the form `*col` so they shouldn't be an issue.

internal/parse/parser.go
Aflynn50: It is, thats in case the user does `col1 , col2` which is weird but valid.

internal/parse/parser.go
TheSignPainter98: 327 checks for a space before the bracket (` ),`), this code doesn't allow for whitespace between the bracket and comma (`) ,`) :P

internal/parse/parser.go
solmyrcan: Actually what I meant was that after skipSpaces at line 321 executes (the case for err=nil and ok is OK :P ), execution falls to line 327, which again skips spaces. The error cases there do return(), so they cannot run line 327 either, so it feels like line 327 is redundant

internal/parse/parser.go
Aflynn50: I'm not sure I understand, once we parse a closing bracket we return. When the bracket is closed we finish parsing, we want to ignore everything after the bracket in this function and leave it up to the caller how they want to handle it.

TheSignPainter98: Ah right my bad I misread, this is fine

internal/parse/parser.go
Aflynn50: Ah yea, good point, so it is.

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
 	return FullName{}, false, nil
 }
 
+// parseList takes a parsing function that returns a FullName and parses a
+// bracketed, comma seperated, list. On success it returns an array of FullName
+// objects. Otherwise it returns an empty array, false and an error.
```
fernape: _Otherwise it returns  a nil array..._
A `nil` array is always empty but an empty array is not necessarily `nil`. Since we explicitly return `nil`, I think the comment should reflect that.

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+	cp := p.save()
+	// We skip a space here to keep consistent with parseTargets which also
+	// consumes one space before the start of the expression.
+	p.skipByte(' ')
+	// Case 1: A single column e.g. p.name
+	if col, ok, _ := p.parseColumn(); ok {
+		return []FullName{col}, true
+	} else if cols, ok, _ := p.parseList((*Parser).parseColumn); ok {
+		return cols, true
+	}
```
fernape: This would be equivalent and a bit more readable maybe? We could easily add a comment for the list case.
```       
        // Case 1: A single column e.g. p.name
        if col, ok, _ := p.parseColumn(); ok {
                return []FullName{col}, true
        }

        // Case 2: Multiple columns
        if cols, ok, _ := p.parseList((*Parser).parseColumn); ok {
                return cols, true
        }
```

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+	// An '&' must be preceded by a space and succeeded by a name or opening
+	// bracket.
+	if p.skipString(" &") {
+		// Case 1: A single target e.g. &Person.name
+		if target, ok, err := p.parseGoFullName(); ok {
+			return []FullName{target}, true, nil
+		} else if err != nil {
+			return nil, false, err
+			// Case 2: Multiple targets e.g. &(Person.name, Person.id)
+		} else if targets, ok, err := p.parseList((*Parser).parseGoFullName); ok {
```
fernape: Similarly as [above](https://github.com/canonical/sqlair/pull/23/files#r1044390489) we could remove the last `} else if...` of the first branch and make them two separate branches. The logic would be the same. In fact, that seems the pattern we have in `parseOutputExpression()`

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+		if fn.Name == "*" {
+			s++
+		}
+	}
+	return s
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (out *OutputPart, ok bool, err error) {
```
fernape: We are not using output named parameters in other places. I'm not sure if they add much value here. For instance, line 405:
```
if targets, ok, err = p.parseTargets(); ok {
```
It seems the intention is to use `ok` and `err` as local variables and not overwrite the output parameters, since all the `return` statements use specific values.

Maybe we could remove the output names.

```
diff --git a/internal/parse/parser.go b/internal/parse/parser.go
index 6ad084e..d9c14d6 100644
--- a/internal/parse/parser.go
+++ b/internal/parse/parser.go
@@ -396,24 +396,23 @@ func starCount(fns []FullName) int {
 
 // parseOutputExpression parses all output expressions. The ampersand must be
 // preceded by a space and followed by a name byte.
-func (p *Parser) parseOutputExpression() (out *OutputPart, ok bool, err error) {
+func (p *Parser) parseOutputExpression() (*OutputPart, bool, error) {
        cp := p.save()
        var cols []FullName
-       var targets []FullName
 
        // Case 1: simple case with no columns e.g. &Person.*
-       if targets, ok, err = p.parseTargets(); ok {
+       if targets, ok, err := p.parseTargets(); ok {
                return &OutputPart{cols, targets}, true, nil
        } else if err != nil {
                return nil, false, err
        }
-       if cols, ok = p.parseColumns(); ok {
+       if cols, ok := p.parseColumns(); ok {
                // Case 2: The expression contains an AS
                // e.g. "p.col1 AS &Person.*".
                numCols := len(cols)
                p.skipSpaces()
                if p.skipString("AS") {
-                       if targets, ok, err = p.parseTargets(); ok {
+                       if targets, ok, err := p.parseTargets(); ok {
                                numTargets := len(targets)
                                // If the target is not * then check there are equal columns
                                // and targets.
 ```

File: internal/parse/parser.go (hunk 8 above)

File: internal/parse/parser.go (hunk 9 above)
Comment (Aflynn50) [reply internal/parse/parser.go]: Changed, and make the examples in the Case comments consistent.

File: internal/parse/parser.go (hunk 10 above)
Comment (Aflynn50) [reply internal/parse/parser.go]: Changed, and also the one in parseOutputExpression

File: internal/parse/parser.go (hunk 11 above)

File: internal/parse/parser.go [hunk 12]
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
 	return FullName{}, false, nil
 }
 
-// parseInputExpression parses an input expression of the form $Type.name.
+// parseList takes a parsing function that returns a FullName and parses a
+// bracketed, comma seperated, list.
+func (p *Parser) parseList(parseFn func(p *Parser) (FullName, bool, error)) ([]FullName, bool, error) {
+	cp := p.save()
+
+	if p.skipByte('(') {
```
niemeyer: Trivial: this might be silghtly simplified / optimized if reorganized as:
```
if !p.skipByte('(') {
    return ...
}

cp := p.save()

File: internal/parse/parser.go [hunk 13]
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+
+		nextItem := true
+		var objs []FullName
+		for nextItem {
+			if obj, ok, err := parseFn(p); ok {
+				objs = append(objs, obj)
+			} else if err != nil {
+				return nil, false, err
+			} else {
+				return nil, false, fmt.Errorf("invalid identifier near char %d", p.pos)
```
niemeyer: Can't quite tell if this will work at this point, so just FYI for now: this may create issues if the expression being parsed is a normal SQL expression that is not supposed to be interpreted by SQLAir. For this not to be an error, this point must only be reached once we're sure this is a SQLAir expression.

niemeyer: Also, maybe instead of "near char %d", we can use it as a prefix such as "column %d: invalid identifier"

internal/parse/parser.go
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+
+			p.skipSpaces()
+			if p.skipByte(')') {
+				return objs, true, nil
+			}
+
+			nextItem = p.skipByte(',')
+			p.skipSpaces()
+		}
+		return nil, false, fmt.Errorf("missing closing parentheses for char %d", parenPos)
```
niemeyer: Same as above about the "column %d: " prefix.

internal/parse/parser.go
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseColumns parses a list of columns. For lists of more than one column the
+// columns must be enclosed in brackets e.g. "(col1, col2) AS &Person.*".
+func (p *Parser) parseColumns() ([]FullName, bool) {
+	cp := p.save()
+
+	// We skip a space here to keep consistent with parseTargets which also
```
niemeyer: That seems a bit awkward. I'd review both to not skip spaces, but rather expect the preceding logic to make sure that this is the boundary of an expression and not the middle of a word. Otherwise, we'll find awkward bugs such as not supporting parenthesis when they are innocuous for instance.

internal/parse/parser.go
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+}
+
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more references Go objects.
+// If the ampersand is not preceded by a space and succeeded by a name or
+// opening bracket the targets are not parsed.
+func (p *Parser) parseTargets() ([]FullName, bool, error) {
+	cp := p.save()
+
+	// An '&' must be preceded by a space and succeeded by a name or opening
```
niemeyer: Same note as above about the space: ideally we can organize it so that the call site is responsible for identifying whether this is a proper place to have targets or not, only call this function at the right places.

internal/parse/parser.go
```diff
@@ -273,11 +302,152 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+	if p.skipString(" &") {
+
+		// Case 1: A single target e.g. "&Person.name".
+		if target, ok, err := p.parseGoFullName(); ok {
+			return []FullName{target}, true, nil
+		} else if err != nil {
+			return nil, false, err
+		}
+
+		// Case 2: Multiple targets e.g. "&(Person.name, Person.id)".
```
niemeyer: I believe the agreement about the syntax was:
```
(&Person.name, &Person.id)
```

niemeyer (changes requested):
Thanks, here is a first pass:

internal/parse/parser.go

internal/parse/parser.go
Aflynn50: The way it currently works is that the caller of parseList decides weather to heed the error or not. If they are trying to parse something they know is an expression they can pass the error on, otherwise they will ignore it.

Changed the comment.

internal/parse/parser.go

internal/parse/parser.go
Aflynn50: I have removed this, though we will only have the desired behaviour (only recognising & when it is preceded by a space) once #16 is merged.

internal/parse/parser.go
Aflynn50: Same reply as above.

internal/parse/parser.go
Aflynn50: With the new logic for parsing `(&P.a, &P.b)` this did need changing. It now works as follows:
- If the first item in the list doesn't parsed we return `ok=false, err=nil`
- If the first item did parse and any subsequent items dont/the closing bracket is missing we throw an error

internal/parse/parser.go
Aflynn50: I've changed the parser to recognise this. It's working all fine so it seems good to switch to this syntax.

internal/parse/parser.go
niemeyer: Nice, that sounds like helpful behavior.

internal/parse/parser.go
niemeyer: Ack.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+			return fullName{name: id}, true, nil
+		}
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
+func (p *Parser) parseTarget() (fullName, bool, error) {
+	cp := p.save()
```
niemeyer: Save and restore are not needed in this method given the logic below, and if the input ($) method looks the same, it would be the same.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		return p.parseGoFullName()
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
 // parseGoFullName parses a Go type name qualified by a tag name (or asterisk)
-// of the form "TypeName.col_name". On success it returns the parsed fullName,
+// of the form "&TypeName.col_name". On success it returns the parsed FullName,
```
niemeyer: Curious.. this one was already lowercased, and should probably stay that way.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+
 	if id, ok := p.parseIdentifier(); ok {
-		if p.skipByte('.') {
-			if idField, ok := p.parseIdentifierAsterisk(); ok {
-				return fullName{id, idField}, true, nil
-			}
-			return fullName{}, false,
-				fmt.Errorf("invalid identifier near char %d", p.pos)
+		if !p.skipByte('.') {
+			return fullName{}, false, fmt.Errorf("go object near char %d not qualified", p.pos)
```
niemeyer: In a previous review in this PR there was a suggestion to tune these error messages using a `column %d:` sort of prefix, but some of them seem to have left behind. "go" should also be capitalized, but actually we can probably avoid using the term "go" here and instead just talk about types and fields when referring to Go, and tables and columns when talking about the database.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+
 	cp.restore()
 	return fullName{}, false, nil
 }
 
-// parseInputExpression parses an input expression of the form $Type.name.
+// parseList takes a parsing function that returns a fullName and parses a
+// bracketed, comma seperated, list.
+func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+	cp := p.save()
```
niemeyer, line 328: The save can be moved past the first skip below, as a small optimization that prevents doing anything unless this is indeed looking like a proper start at all.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		nextItem = p.skipByte(',')
+		p.skipSpaces()
+	}
+	return nil, false, fmt.Errorf("column %d: missing closing parentheses", parenPos)
+}
+
+// parseColumns parses a list of columns. For lists of more than one column the
+// columns must be enclosed in brackets e.g. "(col1, col2) AS &Person.*".
+func (p *Parser) parseColumns() ([]fullName, bool) {
+	cp := p.save()
```
niemeyer: I believe we don't need to save/restore here either, because the entire functionality is implemented inside the two functions below, and only one of them can possibly work.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	cp.restore()
+	return nil, false
+}
+
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more references Go objects.
+// If the ampersand is not preceded by a space and succeeded by a name or
+// opening bracket the targets are not parsed.
+func (p *Parser) parseTargets() ([]fullName, bool, error) {
+	cp := p.save()
```
niemeyer: Same here.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
+	cp := p.save()
```
niemeyer: Similar here, but here we actually need the save/restore before the parseColumsn branch, since we're moving the pointer forward there and may still fail for other reasons.

niemeyer (changes requested):
Good changes, thanks. Second pass, and probably last one.

internal/expr/parser.go
Aflynn50: I've just realised we can't do this. Because the skipByte is negated, if we **don't** exit the function at the if then the state will have been changed, we will have skipped the `(` byte. So we would be saving the state after the byte has been skipped, thus when we restore it it would not be a true restore.

internal/expr/parser.go
Aflynn50: Done.

internal/expr/parser.go
Aflynn50: Done.

internal/expr/parser.go
Aflynn50: Done.

internal/expr/parser.go

internal/expr/parser.go

internal/expr/parser.go

internal/expr/parser.go
niemeyer: Good catch. :+1:

niemeyer (approved):
Thanks for the changes!

PR #24, opened 2022-12-14
Add assemble input

fernape:
Create first PR for the Assemble (or prepared) stage. This one just assembles the input parts, which is simple.


internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/canonical/sqlair/internal/parse"
+	"github.com/canonical/sqlair/internal/typeinfo"
+)
+
+// AssembledExpr represents an SQL expression after the input and output parts
+// have been replaced but their corresponding expressions.
```
TheSignPainter98: `s/but/by/`

internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+type AssembledExpr struct {
+	ParsedExpr *parse.ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*typeinfo.Info
```
TheSignPainter98, line 19: Does this need a separate type? It's only currently used twice

internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+
+// assembleInput prepares an input part.
+// Since input parts do not need any special transformations, this function just
+// checks that the used name corresponds to a tag in the struct.
+// Returns an error otherwise.
+func assembleInput(ti typeNameToInfo, p *parse.InputPart) error {
```
TheSignPainter98, line 22: The middle two lines describe what feels like an implementation detail, is this necessary? If not, I'd also remove the word 'just' here too

internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Assemble prepares a parsed expression.
+// Returns a pointer to AssembledExpr and nil on success.
+// Returns nil and an error otherwise.
+func Assemble(pe *parse.ParsedExpr, args ...any) (expr *AssembledExpr, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot assemble expression: %s", err)
+		}
+	}()
```
TheSignPainter98, line 41: I believe `Errorf` is never called here as each time a variable called `err` is used, it is a new one which shadows the return value, so the `err` here is never assigned.

Also the `defer` feels unnecessary to me, as there are only two places where an error can happen, I think it's clearer to avoid the defer here by copying the `fmt.Errorf` call. Also if the message were to change in future, the strings aren't too far apart

internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+			err := assembleInput(ti, p)
+			if err != nil {
+				return nil, err
+			}
+			sql.WriteString(p.ToSQL([]string{}))
+			continue
+		}
+		if p, ok := part.(*parse.OutputPart); ok {
+			// Do nothing for now.
+			sql.WriteString(p.ToSQL([]string{"DUMMY_OUTPUT"}))
```
TheSignPainter98, line 67: Is this a placeholder to be replaced in a future PR?

internal/assemble/assemble_test.go
```diff
@@ -0,0 +1,45 @@
...
+	PostalCode int    `db:"address_id"`
+}
+
+type Manager struct {
+	Name string `db:"manager_name"`
+}
+
+type District struct {
+}
+
```
TheSignPainter98, line 36: It would be good to also test valid input as well as invalid ones :)

internal/parse/parser.go
```diff
@@ -122,17 +122,24 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
-		if sp, ok, err := p.parseStringLiteral(); err != nil {
+		if in, ok, err := p.parseInputExpression(); err != nil {
 			return nil, err
 		} else if ok {
-			p.add(sp)
+			p.add(in)
+			continue
+		}
+
+		if bypass, ok, err := p.parseStringLiteral(); err != nil {
```
TheSignPainter98: Perhaps this should be called `str` or `lit` to match the above?

internal/parse/parser.go
```diff
@@ -256,6 +263,25 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
+// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// its table name. parseColumn returns an error so that it can be used with
+// parseList.
```
TheSignPainter98: Do you need to mention `parseList` specifically? This seems like a detail which could easily get out-dated in future

internal/assemble/assemble.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf("there is no tag with name %s in %s",
+			p.Source.Name, inf.Type.Name())
+	}
+	return fmt.Errorf("unknown type: %s", p.Source.Prefix)
+}
+
+// Assemble prepares a parsed expression.
+// Returns a pointer to AssembledExpr and nil on success.
+// Returns nil and an error otherwise.
+func Assemble(pe *parse.ParsedExpr, args ...any) (expr *AssembledExpr, err error) {
```
TheSignPainter98, line 36: This doesn't really tell me why I should use `Assemble`. Also as I think it's quite common for Go functions to return `(SomeValue, error)`, describing the error in the doc comment can be avoided.

Perhaps something like this?
```go
// Assemble prepares a parsed expression for execution in SQL.
```

internal/parse/parser.go
```diff
@@ -256,6 +263,25 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+func (p *Parser) parseColumn() (FullName, bool, error) {
+	cp := p.save()
+	if id, ok := p.parseIdentifierAsterisk(); ok {
+		if id != "*" && p.skipByte('.') {
+			if idCol, ok := p.parseIdentifierAsterisk(); ok {
+				return FullName{Prefix: id, Name: idCol}, true, nil
+			}
+		} else {
+			// A column name specified without a table prefix should be in Name.
+			return FullName{Name: id}, true, nil
```
TheSignPainter98: 'Should' sounds unsure, also I'm not sure this comment is necessary?

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more Go objects. If the ampersand is not found
+// or is not preceded by a space and succeeded by a name or opening bracket the
+// targets are not parsed.
+func (p *Parser) parseTargets() ([]FullName, bool, error) {
+	cp := p.save()
+
+	// An '&' must be preceded by a space and succeeded by a name or opening
+	// bracket.
+	if p.skipString(" &") {
```
TheSignPainter98: Should this skip exactly one space or one-or-more spaces? Perhaps something like:
```go
if p.skipSpaces() && p.skipByte('&') {
```
This would match the parsing of `"AS"`

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+
+	// Case 1: simple case with no columns e.g. &Person.*
+	if targets, ok, err = p.parseTargets(); ok {
+		return &OutputPart{cols, targets}, true, nil
+	} else if err != nil {
+		return nil, false, err
+	}
+	if cols, ok = p.parseColumns(); ok {
+		// Case 2: The expression contains an AS
+		// e.g. "p.col1 AS &Person.*".
```
TheSignPainter98: The case 1 and case 2 comments appear in different scopes, perhaps the case 2 comment should be immediately before its `if`.

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+		p.skipSpaces()
+		if p.skipString("AS") {
+			if targets, ok, err = p.parseTargets(); ok {
+				numTargets := len(targets)
+				// If the target is not * then check there are equal columns
+				// and targets.
+				if !(numTargets == 1 && targets[0].Name == "*") &&
+					numCols != numTargets {
+					return nil, false, fmt.Errorf("number of cols = %d "+
+						"but number of targets = %d in expression near %d",
```
TheSignPainter98: "...in expression near _char_ %d"

internal/parse/parser.go
```diff
@@ -277,6 +303,137 @@ func (p *Parser) parseGoFullName() (FullName, bool, error) {
...
+				return nil, false, fmt.Errorf("invalid identifier near char %d", p.pos)
+			}
+			p.skipSpaces()
+			if p.skipByte(')') {
+				return objs, true, nil
+			}
+			nextItem = p.skipByte(',')
+			p.skipSpaces()
+		}
+		return nil, false, fmt.Errorf("missing closing parentheses for char %d", parenPos)
```
TheSignPainter98: I think either "at char" is slightly clearer, or perhaps "near char" is more consistent with the other error messages here

internal/assemble/assemble.go
fernape: Done. Thanks!

internal/assemble/assemble.go
Aflynn50: I believe this is callable. We have named return values so the `err` shouldn't be a shadow. We have the same pattern in `Parse` (unless there is a difference I haven't caught).

It is arguably somewhat unnecessary but there will soon be three places the error can happen (one for the output part as well) and it's nice to have consistency across the `Parse` and `Assemble` functions

internal/assemble/assemble_test.go
Aflynn50: Ah yea, so the thinking was that all the valid tests would be done in the complete whitebox testing for the whole package, but on reflection there's no reason not to also have some here, so they have been added.

internal/parse/parser.go
Aflynn50: So this should actually be part of a seperate PR, it seems it ended up here by mistake.

internal/parse/parser.go
Aflynn50: Same as above.

internal/assemble/assemble.go
Aflynn50: Changed.

internal/parse/parser.go
Aflynn50: Also meant to be part of a different PR.

internal/parse/parser.go
Aflynn50: Same as above.

internal/assemble/assemble.go
TheSignPainter98: Okay, I was incorrect, this code does function as when returning, the named return values are implicitly assigned (I was unaware of this). There is shadowing here though as every time `err` is assigned, it is on the left of a `:=`, so the named `err` return value isn't actually assigned within the body of the function, only when `return` is called.

This behaviour within the defer seems well-defined but still a little magical to me... the code looks quite misleading.

internal/parse/parts.go
```diff
@@ -52,8 +55,16 @@ func (p *OutputPart) String() string {
 	return fmt.Sprintf("OutputPart[%+v %+v]", p.Source, p.Target)
 }
 
-func (p *OutputPart) ToSQL() string {
-	return ""
+func (p *OutputPart) ToSQL(cs []string) string {
+	var out bytes.Buffer
+	out.WriteString(" ")
+	for i, c := range cs {
+		if i > 0 {
```
solmyrcan, line 62: Is this method meant to return a string as "`cs[0], cs[1], ... , cs[len-1]`" ? (i.e. a "`, `"-separated sequence of words). If so, maybe we could then do `if i != len(s)-1` instead of `if i>0`, and move the `if` block after `out.WriteString(c)` (so for `i=0,...len-2`), so that no comma is added after the last element. Is the initial `out.WriteString(" ")` there to prevent an empty string output if cs is empty? If so, would it make sense to do `if (len(s)==0)` right before `out.WriteString(" ")`

internal/assemble/assemble.go
```diff
@@ -0,0 +1,75 @@
+package assemble
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/canonical/sqlair/internal/parse"
+	"github.com/canonical/sqlair/internal/typeinfo"
```
solmyrcan, line 9: Not in this PR, but in `typeinfo.go`, we might need to change the docstring from:
`// Reflect will return the Info of a given type,` to:
`// TypeInfo will return the Info of a given type,`

internal/assemble/assemble.go
```diff
@@ -0,0 +1,75 @@
...
+// AssembledExpr represents an SQL expression after the input and output parts
+// have been replaced by their corresponding expressions.
+type AssembledExpr struct {
+	ParsedExpr *parse.ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*typeinfo.Info
+
+// assembleInput prepares an input part.
```
solmyrcan, line 21: If I understand it correctly, this method checks whether  the `Source` in `p` has a valid Prefix first, and if so then checks whether there's a `Field` entry with name `Source.Name`. It doesn't return the Field. Feels like it's a validation method, working on the `Source`. That makes me think, should the method name be altered accordingly? (and the docstring?)

internal/assemble/assemble.go
Aflynn50: True, though I think the shadowing doesn't cause any problems. The assignment of the error to the `err` declared in the function header is done when return is called and the `defer` func always runs after that. It is a bit weird, though I guess its just the magic of named return values, I can't think of another language that has them.

fernape:
Closing because it's been superseded by https://github.com/canonical/sqlair/pull/28

PR #25, opened 2022-12-14
Expand skip spaces

fernape:
Expand `skipSpaces` so it handles new lines and tabs. This will be useful since queries in code tend to be indented for clarity. Since it now handles more than spaces, rename the function to `skipBlanks`

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -42,6 +42,28 @@ var tests = []struct {
 	"SELECT p.* AS &Person.*",
 	"[Bypass[SELECT p.* AS &Person.*]]",
 }, {
+	"spaces and tabs",
+	"SELECT p.* 	AS 		   &Person.*",
+	"ParsedExpr[BypassPart[SELECT p.* 	AS 		   &Person.*]]",
+}, {
+	"new lines",
+	`SELECT
+		p.* AS &Person.*,
+		foo
+	 FROM t
+	 WHERE
+		foo = bar
+		and
+		x = y`,
+	`ParsedExpr[BypassPart[SELECT
+		p.* AS &Person.*,
+		foo
+	 FROM t
+	 WHERE
+		foo = bar
+		and
+		x = y]]`,
+}, {
 	"quoted output expression",
 	"SELECT p.* AS &Person.*, '&notAnOutputExpresion.*' AS literal FROM t",
 	"[Bypass[SELECT p.* AS &Person.*, ] " +
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -177,9 +177,8 @@ func (p *Parser) skipByteFind(b byte) bool {
 	return false
 }
 
-// skipBlanks advances the parser jumping over consecutive spaces, tabs or new
-// lines. It stops when finding a non-blank character. Returns true if the
-// parser position was actually changed, false otherwise.
+// skipBlanks advances the parser past spaces, tabs and newlines. Returns
+// whether the parser position was changed.
 func (p *Parser) skipBlanks() bool {
 	mark := p.pos
 	blanks := map[byte]bool{
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -177,13 +177,14 @@ func (p *Parser) skipByteFind(b byte) bool {
 	return false
 }
 
-// skipSpaces advances the parser jumping over consecutive spaces. It stops when
-// finding a non-space character. Returns true if the parser position was
-// actually changed, false otherwise.
-func (p *Parser) skipSpaces() bool {
+// skipBlanks advances the parser jumping over consecutive spaces, tabs or new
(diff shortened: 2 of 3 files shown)


internal/parse/parser_internal_test.go
```diff
@@ -29,11 +29,12 @@ func TestRunTable(t *testing.T) {
...
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "abc    d", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "     abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "  \t  abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "\t  abcd", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{false}, input: "", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{false}, input: "abc    d", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{true}, input: "     abcd", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{true}, input: "  \t  abcd", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{true}, input: "\t  abcd", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{true}, input: "\n\t  abcd", data: []string{}},
```
TheSignPainter98: There's no test for `\r` here :)

internal/parse/parser.go
```diff
@@ -177,15 +177,18 @@ func (p *Parser) skipByteFind(b byte) bool {
...
 }
 
-// skipSpaces advances the parser jumping over consecutive spaces. It stops when
-// finding a non-space character. Returns true if the parser position was
-// actually changed, false otherwise.
-func (p *Parser) skipSpaces() bool {
+// skipBlanks advances the parser jumping over consecutive spaces, tabs or new
+// lines. It stops when finding a non-blank character. Returns true if the
+// parser position was actually changed, false otherwise.
+func (p *Parser) skipBlanks() bool {
```
TheSignPainter98: The middle sentence seems redundant to me. Perhaps something like this?
```go
// skipBlanks advances the parser past spaces, tabs and newlines. Returns whether
// the parser position was changed.
```

internal/parse/parser_internal_test.go
fernape: Done.

Thanks!

internal/parse/parser.go
fernape: Simpler and clearer. Thanks!

internal/parse/parser_internal_test.go
```diff
@@ -29,11 +29,16 @@ func TestRunTable(t *testing.T) {
 		{bytef: p.skipByteFind, result: []bool{false, true, true}, input: "abcde", data: []string{"x", "b", "c"}},
 		{bytef: p.skipByteFind, result: []bool{true, false}, input: "abcde ", data: []string{" ", " "}},
 
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "abc    d", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "     abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{true}, input: "  \t  abcd", data: []string{}},
-		{stringf0: p.skipSpaces, result: []bool{false}, input: "\t  abcd", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{false}, input: "", data: []string{}},
+		{stringf0: p.skipBlanks, result: []bool{false}, input: "abc    d", data: []string{}},
```
solmyrcan: Would it be relevant to have test cases for `\n \r \t` also mixed up in the whitespaces as in here:
`input: "abc  \t  d"`
Just used `\t` here, but maybe others as well?

fernape: Those should be covered with the tests around line 37 and such.

internal/parse/parser.go
```diff
@@ -177,15 +177,17 @@ func (p *Parser) skipByteFind(b byte) bool {
...
-		if p.input[p.pos] != ' ' {
-			break
-		}
+	blanks := map[byte]bool{
+		' ':  true,
+		'\t': true,
+		'\r': true,
+		'\n': true,
+	}
+	for p.pos < len(p.input) && blanks[p.input[p.pos]] {
```
niemeyer: This is a longer and much more expensive version of this:
```
for ... {
    switch p.input[p.pos] {
    case ' ', '\t', ...:
        ...
    }
}
```

niemeyer (changes requested):
Idea is good, but the implementation could be tuned a bit per notes below.

internal/parse/parser.go
fernape: Done. Thanks.

PR #26, opened 2022-12-15
Move black box tests

Aflynn50:
This PR is to move the black box tests of the whole library into a seperate test package. Currently they are in parse_test.


niemeyer (changes requested):
As we spoke in the call, this change has a bit of a bad smell. Tests are being taken away from the parsing code and into an empty test package that doesn't have a clearly defined context. I understand that the goal was to move tests to a location that could be used when testing other stages of the parsing, but that also means that now the parsing code has no clear tests next to it, and that the refactoring of these tests may end up dropping parsing tests by mistake. I would suggest keeping the parser tests next to the parser still, and having further tests next to the further logic, or alternatively to rethink the package organization so that more stages are part of the same properly encapsulated expression package.

PR #28, opened 2023-01-05
Add Prepare stage for inputs

Aflynn50:
This PR contains the Prepare stage for input parts. In the prepare stage the user passes instantiations of every Go struct that is mentioned in an input or output expression. These objects are used only for information about their type, the actual values inside the structs are irrelevant. 

The `prepareInput` function checks that the structs and tags mentioned in IO expressions in the query correspond to the types of the objects passed as arguments to `Prepare`.

This PR is an updated version of #24 which was a draft PR. However since the name of the branch needed changing this PR has been created.

### DIFF
--- a/internal/expr/prepare.go
+++ b/internal/expr/prepare.go
@@ -3,6 +3,7 @@ package expr
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"strings"
 )
 
@@ -21,6 +22,7 @@ func getKeys(m map[string]*info) []string {
 		keys[i] = k
 		i++
 	}
+	sort.Strings(keys)
 	return keys
 }
 
@@ -78,7 +80,7 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 		case *bypassPart:
 			sql.WriteString(p.chunk)
 		default:
-			return nil, fmt.Errorf("internal error")
+			return nil, fmt.Errorf("internal error: unknown query part type %T", part)
 		}
 	}
 
--- a/internal/expr/prepare.go
+++ b/internal/expr/prepare.go
@@ -3,26 +3,37 @@ package expr
 import (
 	"bytes"
 	"fmt"
+	"strings"
 )
 
 // PreparedExpr contains an SQL expression that is ready for execution.
 type PreparedExpr struct {
-	ParsedExpr *ParsedExpr
-	SQL        string
+	inputs []*inputPart
+	SQL    string
 }
 
 type typeNameToInfo map[string]*info
 
+func getKeys(m map[string]*info) []string {
+	i := 0
+	keys := make([]string, len(m))
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	return keys
+}
+
 // prepareInput checks that the input expression corresponds to a known type.
 func prepareInput(ti typeNameToInfo, p *inputPart) error {
-	inf, ok := ti[p.source.prefix]
+	info, ok := ti[p.source.prefix]
 	if !ok {
-		return fmt.Errorf(`unknown type: "%s"`, p.source.prefix)
+		return fmt.Errorf(`type %s unknown, have: %s`, p.source.prefix, strings.Join(getKeys(ti), ", "))
 	}
 
-	if _, ok = inf.tagToField[p.source.name]; !ok {
-		return fmt.Errorf(`no tag with name "%s" in "%s"`,
-			p.source.name, inf.structType.Name())
+	if _, ok = info.tagToField[p.source.name]; !ok {
+		return fmt.Errorf(`type %s has no %q db tag`,
+			info.structType.Name(), p.source.name)
 	}
 
 	return nil
@@ -42,32 +53,34 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 
 	// Generate and save reflection info.
 	for _, arg := range args {
-		inf, err := typeInfo(arg)
+		info, err := typeInfo(arg)
 		if err != nil {
 			return nil, err
 		}
-		ti[inf.structType.Name()] = inf
+		ti[info.structType.Name()] = info
 	}
 
 	var sql bytes.Buffer
 	// Check and expand each query part.
+
+	ins := []*inputPart{}
+
 	for _, part := range pe.queryParts {
-		if p, ok := part.(*inputPart); ok {
+		switch p := part.(type) {
+		case *inputPart:
 			err := prepareInput(ti, p)
 			if err != nil {
 				return nil, err
 			}
-			sql.WriteString(p.toSQL())
-			continue
+			sql.WriteString("?")
+			ins = append(ins, p)
+		case *outputPart:
+		case *bypassPart:
+			sql.WriteString(p.chunk)
+		default:
+			return nil, fmt.Errorf("internal error")
 		}
-		if p, ok := part.(*outputPart); ok {
-			// Do nothing for now.
-			sql.WriteString(p.toSQL())
-			continue
-		}
-		p := part.(*bypassPart)
-		sql.WriteString(p.toSQL())
 	}
 
-	return &PreparedExpr{ParsedExpr: pe, SQL: sql.String()}, nil
+	return &PreparedExpr{inputs: ins, SQL: sql.String()}, nil
 }
--- a/internal/expr/prepare.go
+++ b/internal/expr/prepare.go
@@ -1,0 +1,73 @@
+package expr
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// PreparedExpr contains an SQL expression that is ready for execution.
+type PreparedExpr struct {
+	ParsedExpr *ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*info
+
+// prepareInput checks that the input expression corresponds to a known type.
+func prepareInput(ti typeNameToInfo, p *inputPart) error {
+	inf, ok := ti[p.source.prefix]
+	if !ok {
+		return fmt.Errorf(`unknown type: "%s"`, p.source.prefix)
+	}
+
+	if _, ok = inf.tagToField[p.source.name]; !ok {
+		return fmt.Errorf(`no tag with name "%s" in "%s"`,
+			p.source.name, inf.structType.Name())
+	}
+
+	return nil
+}
+
+// Prepare takes a parsed expression and all the Go objects mentioned in it.
+// The IO parts of the statement are checked for validity against the types
+// and expanded if necessary.
+func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot prepare expression: %s", err)
+		}
+	}()
+
+	var ti = make(typeNameToInfo)
+
+	// Generate and save reflection info.
+	for _, arg := range args {
+		inf, err := typeInfo(arg)
+		if err != nil {
+			return nil, err
+		}
+		ti[inf.structType.Name()] = inf
+	}
+
+	var sql bytes.Buffer
+	// Check and expand each query part.
+	for _, part := range pe.queryParts {
+		if p, ok := part.(*inputPart); ok {
+			err := prepareInput(ti, p)
+			if err != nil {
+				return nil, err
+			}
+			sql.WriteString(p.toSQL())
+			continue
+		}
+		if p, ok := part.(*outputPart); ok {
+			// Do nothing for now.
+			sql.WriteString(p.toSQL())
+			continue
+		}
+		p := part.(*bypassPart)
+		sql.WriteString(p.toSQL())
+	}
+
+	return &PreparedExpr{ParsedExpr: pe, SQL: sql.String()}, nil
+}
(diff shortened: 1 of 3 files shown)


internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+// have been replaced by their corresponding expansions.
+type PreparedExpr struct {
+	ParsedExpr *parse.ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*typeinfo.Info
+
+// prepareInput checks that the input expression corrosponds to a Go struct
+// passed to Prepare.
```
TheSignPainter98: s/corrosponds/corresponds/

internal/prepare/prepare_test.go
```diff
@@ -0,0 +1,67 @@
...
+	sql := "SELECT street FROM t WHERE x = $Address.street"
+	parser := parse.NewParser()
+	parsedExpr, err := parser.Parse(sql)
+	_, err = prepare.Prepare(parsedExpr, Person{ID: 1})
+	c.Assert(err, ErrorMatches, "cannot prepare expression: unknown type: Address")
+}
+
+func (s *ParserSuite) TestMissingTagInput(c *C) {
+	sql := "SELECT street FROM t WHERE x = $Address.number"
+	parser := parse.NewParser()
```
TheSignPainter98: This isn't within the scope of this PR, but `parse.NewParser()` feels slightly off to me, perhaps renaming the `parse` module to `syntax` would be clearer (this would become `syntax.NewParser()`). Being a noun, this module name would match the existing `typeinfo` module, but be at odds with the existing `prepare` module (verb).

internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("there is no tag with name %s in %s",
+			p.Source.Name, inf.Type.Name())
+	}
+	return fmt.Errorf("unknown type: %s", p.Source.Prefix)
+}
+
+// Prepare takes a parsed expression and all the go objects mentioned in it.
+// The IO parts of the statement are checked for validity against the Go objects
+// and expanded if necessary.
+func Prepare(pe *parse.ParsedExpr, args ...any) (expr *PreparedExpr, err error) {
```
TheSignPainter98: The capitalisation on 'go' is inconsistent here :P 

Also this comment doesn't really describe _why_ `Prepare` would be used, it seems to be a re-phrasing of what the code is doing. Following a [previous conversation] on #24, perhaps something like this would be good?
```go
// Prepare prepares a parsed expression for execution in SQL.
```
[previous conversation]: https://github.com/canonical/sqlair/pull/24#discussion_r1048450990

internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+		if p, ok := part.(*parse.OutputPart); ok {
+			// Do nothing for now.
+			sql.WriteString(p.ToSQL())
+			continue
+		}
+		p := part.(*parse.BypassPart)
+		sql.WriteString(p.ToSQL())
+	}
+
+	return &PreparedExpr{ParsedExpr: pe, SQL: strings.TrimSpace(sql.String())}, nil
```
TheSignPainter98: Is this `TrimSpace` necessary? If it is, it sounds like `sql.String()` is putting in too many spaces for some reason,  so should probably be fixed elsewhere.

Aflynn50: True, this can actually be removed. Previously we had been enforcing correct spacing around queries, but now its been changed to accept whatever was there before. The `TrimSpace` took care of spaces round IO expressions at the start of end of the string.

internal/prepare/prepare_test.go
Aflynn50: It does feel off, I'm currently reading my way through the google style guide and it suggests using something like `parse.New` to avoid repetition. At some point we're going to do a big PR that makes it conform to the  The `typeinfo` module was originally `reflect` which made more sense, but clashed with the standard library.

internal/prepare/prepare.go
Aflynn50: Done.

internal/prepare/prepare.go
Aflynn50: Fixed, and comment added.

internal/prepare/prepare_test.go
TheSignPainter98: Ah right, I understand.

I actually really like the idea of using `parse.New` (or something similar), it feels very nice and concise.

internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+	ParsedExpr *parse.ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*typeinfo.Info
+
+// prepareInput checks that the input expression corresponds to a Go struct
+// passed to Prepare.
+func prepareInput(ti typeNameToInfo, p *parse.InputPart) error {
+	if inf, ok := ti[p.Source.Prefix]; ok {
```
solmyrcan: I read about this recently - dealing with error cases first and return, and then leave the "success" case and its treatment after those checks are done, rather than success being within an if block. Perhaps this approach could be adopted here? :) 

```
if inf, ok := ti[p.Source.Prefix]; !ok {
    return fmt.Errorf("unknown type: %s", p.Source.Prefix)
}
if _, ok := inf.TagToField[p.Source.Name]; !ok {
    return fmt.Errorf("there is no tag with name %s in %s", p.Source.Name, inf.Type.Name())
}
// success
return nil
```

internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			err = fmt.Errorf("cannot prepare expression: %s", err)
+		}
+	}()
+
+	var ti = make(typeNameToInfo)
+
+	// Generate and save reflection info.
+	for _, arg := range args {
+		i, err := typeinfo.TypeInfo(arg)
```
solmyrcan: Minor comment, but would it make sense to use something like "info" rather than "i" here, as I guess traditionally "i" is used as an index variable name.

internal/prepare/prepare.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+			sql.WriteString(p.ToSQL())
+			continue
+		}
+		if p, ok := part.(*parse.OutputPart); ok {
+			// Do nothing for now.
+			sql.WriteString(p.ToSQL())
+			continue
+		}
+		p := part.(*parse.BypassPart)
```
solmyrcan: Would we need to check the outcome of the type assertion here as well? If I understand it correctly, the type could only be one of `*parse.InputPart`, `*parse.OutputPart`, and `*parse.BypassPart`, but from an error handling point of view, would `Prepare` need to handle any case in which the input is actually not one of these, for reliability.

internal/prepare/prepare_test.go
```diff
@@ -0,0 +1,67 @@
...
+		c.Assert(preparedExpr.SQL, Equals, test.expectedPrepared)
+	}
+}
+
+func (s *ParserSuite) TestMismatchedInputStructName(c *C) {
+	sql := "SELECT street FROM t WHERE x = $Address.street"
+	parser := parse.NewParser()
+	parsedExpr, err := parser.Parse(sql)
+	_, err = prepare.Prepare(parsedExpr, Person{ID: 1})
+	c.Assert(err, ErrorMatches, "cannot prepare expression: unknown type: Address")
```
solmyrcan: I may not have understood the test here, so please ignore if not relevant :) In line, 57, we deliberately pass a `Person` to `Prepare()`, so should it complain about the `sql` statement having an Address or about the second argument being a `Person`? Does this test have the expectation that `Prepare()` must complain when it receives an expression that does not contain the provided type (`Person`, in this case). I guess the error is produced by `prepareInput()`. Could that method equivalently complain about the provided argument type itself? or do we always blame the sql expression :P

internal/prepare/prepare.go
Aflynn50: True, it does seem unlikely that another would be added, but since in Go its pretty easy to accidentally make something part of an interface its probably worth adding this anyway.

internal/prepare/prepare.go
Aflynn50: Maybe, but I think since its scope is so small its probably ok there, and then its consistant with `ti` which is Type Info.

internal/prepare/prepare.go
Aflynn50: Very true! I also read this recently - I've changed it.

internal/prepare/prepare_test.go
Aflynn50: Prepare basically takes all the type arguments it gets into a bag and looks in the bag for the ones it finds in the query. So if extra arguments are given then they are ignored.

Admittedly this behaviour isn't really documented, so I will add that!

internal/prepare/prepare.go
Aflynn50: I've changed my mind, I think `inf` is actually better lol.

internal/expr/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// PreparedExpr contains an SQL expression that is ready for execution.
+type PreparedExpr struct {
+	ParsedExpr *ParsedExpr
+	SQL        string
```
niemeyer: I would make those details private, as one of them is the input to the query, and the other one is the actual "prepared" expression. If we expose it, we'll lose the opportunity to do anything else with this type without breaking its promises.

internal/expr/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type typeNameToInfo map[string]*info
+
+// prepareInput checks that the input expression corresponds to a known type.
+func prepareInput(ti typeNameToInfo, p *inputPart) error {
+	inf, ok := ti[p.source.prefix]
+	if !ok {
+		return fmt.Errorf(`unknown type: "%s"`, p.source.prefix)
```
niemeyer: Instead of ˋ"%s"ˋ you can use %q. The latter will handle escaping.

That said, we should not need the quoting here as these names shoud be validated elsewhere, and perhaps we could make the debugging easier with something along the lines of:
```
cannot prepare expression: type Fooz unknown, have: Foo, Bar, Baz
```

internal/expr/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if _, ok = inf.tagToField[p.source.name]; !ok {
+		return fmt.Errorf(`no tag with name "%s" in "%s"`,
```
niemeyer: Similarly, maybe:
```
cannot prepare expression: type Fooz has no "field_name" db tag
```
The field name here needs to be quoted because being a word it could be mixed with the message and misinterpreted.

internal/expr/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			err = fmt.Errorf("cannot prepare expression: %s", err)
+		}
+	}()
+
+	var ti = make(typeNameToInfo)
+
+	// Generate and save reflection info.
+	for _, arg := range args {
+		inf, err := typeInfo(arg)
```
niemeyer: Can we please use "info" instead of "inf" for the shorthand? I read it as infinite every time. :)

internal/expr/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			return nil, err
+		}
+		ti[inf.structType.Name()] = inf
+	}
+
+	var sql bytes.Buffer
+	// Check and expand each query part.
+	for _, part := range pe.queryParts {
+		if p, ok := part.(*inputPart); ok {
```
niemeyer: This would look a bit nicer with a type switch statement. It also needs a default clause so we get an "internal error" message instead of a panic if we ever do something silly.

niemeyer (changes requested):
Probably the only pass. Looks very nice, thanks!

internal/expr/prepare.go

internal/expr/prepare.go
Aflynn50: I've changed `ParsedExpr` to `inputs` and made it private. I've left SQL as public because we use it in testing and it makes sense that the `PreparedExpr` should have the prepared SQL.

internal/expr/prepare.go

internal/expr/prepare.go

internal/expr/prepare.go

internal/expr/prepare.go
```diff
@@ -0,0 +1,86 @@
...
+type typeNameToInfo map[string]*info
+
+func getKeys(m map[string]*info) []string {
+	i := 0
+	keys := make([]string, len(m))
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	return keys
```
niemeyer, line 26: Are tests passing consistently?  Iterating over the map is unstable, so this needs a `sort.Strings` call before return.

internal/expr/prepare.go
```diff
@@ -0,0 +1,86 @@
...
+			if err != nil {
+				return nil, err
+			}
+			sql.WriteString("?")
+			ins = append(ins, p)
+		case *outputPart:
+		case *bypassPart:
+			sql.WriteString(p.chunk)
+		default:
+			return nil, fmt.Errorf("internal error")
```
niemeyer: This should say what the internal error is: ("internal error: unknown query part type: %T", part)

niemeyer (approved):
A couple of trivials and LGTM, thanks!

internal/expr/prepare.go
Aflynn50: I clearly hadn't run it enough times! Changed.

internal/expr/prepare.go

PR #29, opened 2023-01-10
Refector packages

Aflynn50:
This moves the contents of the `parse` and `typeinfo` packages to a new package called `expr`. All the functions changed are to be not exported so that they may be exported when needed.

The odd variable name has been changed here and there to avoid clashes with the new lowercase type.


internal/expr/parts.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (p *outputPart) toString() string {
+	return fmt.Sprintf("OutputPart[%+v %+v]", p.source, p.target)
+}
+
+func (p *outputPart) toSQL() string {
+	return ""
+}
+
+// BypassPart represents a part of the expression that we want to pass to the
```
TheSignPainter98: `s/BypassPart/bypassPart/`

internal/expr/parts.go
```diff
@@ -0,0 +1,71 @@
...
+// backend database verbatim.
+type bypassPart struct {
+	chunk string
+}
+
+func (p *bypassPart) toString() string {
+	return "BypassPart[" + p.chunk + "]"
```
TheSignPainter98: Also here, `s/BypassPart/bypassPart/`

internal/expr/parts.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// InputPart represents a named parameter that will be sent to the database
+// while performing the query.
+type inputPart struct {
+	source fullName
+}
+
+func (p *inputPart) toString() string {
+	return fmt.Sprintf("InputPart[%+v]", p.source)
```
TheSignPainter98: Capitalisation---should this be `inputPart[...]`?

internal/expr/parts.go
```diff
@@ -0,0 +1,71 @@
...
+
+// OutputPart represents a named target output variable in the SQL expression,
+// as well as the source table and column where it will be read from.
+type outputPart struct {
+	source []fullName
+	target []fullName
+}
+
+func (p *outputPart) toString() string {
+	return fmt.Sprintf("OutputPart[%+v %+v]", p.source, p.target)
```
TheSignPainter98: Likewise, should this be `outputPart`?

Aflynn50: Changed.

internal/expr/parts.go
Aflynn50: Changed.

internal/expr/parts.go
Aflynn50: Changed.

internal/expr/parts.go
Aflynn50: Changed.

internal/expr/parser_test.go
```diff
@@ -1,9 +1,8 @@
-package parse_test
+package expr
```
niemeyer: PR looks good and seems ready to go in. As we discussed, this change is the only one that seems like a step backwards, as the tests were already using the public API.  If we revert this, we can merge the PR pretty much as-is and convert the other tests to also use the public API and poke holes through when actually required via the export_test.go pattern.

niemeyer (changes requested):

niemeyer (approved):
Thanks for the changes.

PR #30, opened 2023-01-10
Improve parse tests

Aflynn50:
This PR moves the large chunks of SQL and `ParsedExpr` onto single lines. It also adds some validity tests.

These are improvements that were added in the now closed #26 but would still be beneficial to have.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -44,234 +44,115 @@ var tests = []struct {
 }, {
 	"multiple-quoted bypass expression",
 	`SELECT '''' AS &Person.*`,
-	`[Bypass[SELECT ] Bypass[''''] Bypass[ AS &Person.*]]`,
+	`[Bypass[SELECT ] Bypass[''''] Bypass[ AS ] Output[[] [Person.*]]]`,
 }, {
 	"single quote in double quotes",
 	`SELECT "'" AS &Person.*`,
-	`[Bypass[SELECT ] Bypass["'"] Bypass[ AS &Person.*]]`,
-}, {
-	"dollar without preceding space",
-	"SELECT p.* AS&Person.*",
-	"[Bypass[SELECT p.* AS&Person.*]]",
+	`[Bypass[SELECT ] Bypass["'"] Bypass[ AS ] Output[[] [Person.*]]]`,
 }, {
 	"quoted output expression",
 	"SELECT p.* AS &Person.*, '&notAnOutputExpresion.*' AS literal FROM t",
-	"[Bypass[SELECT ] " +
-		"Output[[p.*] [Person.*]] " +
-		"Bypass[, ] " +
-		"Bypass['&notAnOutputExpresion.*'] " +
-		"Bypass[ AS literal FROM t]]",
+	"[Bypass[SELECT ] Output[[p.*] [Person.*]] Bypass[, ] Bypass['&notAnOutputExpresion.*'] Bypass[ AS literal FROM t]]",
 }, {
 	"star as output",
 	"SELECT * AS &Person.* FROM t",
-	"[Bypass[SELECT ] " +
-		"Output[[*] [Person.*]] " +
-		"Bypass[ FROM t]]",
+	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[ FROM t]]",
 }, {
 	"input v1",
 	"SELECT foo, bar FROM table WHERE foo = $Person.id",
-	"[Bypass[SELECT foo, bar FROM table WHERE foo = ] " +
-		"Input[Person.id]]",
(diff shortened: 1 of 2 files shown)


internal/parse/parser_test.go
```diff
@@ -41,155 +41,95 @@ var tests = []struct {
 }, {
 	"quoted output expression",
 	"SELECT p.* AS &Person.*, '&notAnOutputExpresion.*' AS literal FROM t",
-	"ParsedExpr[BypassPart[SELECT p.* AS &Person.*, ] " +
-		"BypassPart['&notAnOutputExpresion.*'] " +
-		"BypassPart[ AS literal FROM t]]",
+	"ParsedExpr[BypassPart[SELECT p.* AS &Person.*, ] BypassPart['&notAnOutputExpresion.*'] BypassPart[ AS literal FROM t]]",
```
TheSignPainter98: Should the first two `BypassPart`s be `OutputPart`s? There's a few other places where the same could apply, perhaps I am missing some context here.

internal/parse/parser_test.go
```diff
@@ -41,155 +41,95 @@ var tests = []struct {
...
-	"ParsedExpr[BypassPart[SELECT p.* AS &Person.*, " +
-		"(a.district, a.street) AS &Address.* " +
-		"FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = ] " +
-		"InputPart[Person.name] " +
-		"BypassPart[) UNION SELECT p.* AS &Person.*, (a.district, a.street) AS &Address.* " +
-		"FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = ] " +
-		"InputPart[Person.name] " +
-		"BypassPart[)]]",
+	"SELECT p.* AS &Person.*, (a.district, a.street) AS &Address.* FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = $Person.name) UNION SELECT p.* AS &Person.*, (a.district, a.street) AS &Address.* FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = $Person.name)",
+	"ParsedExpr[BypassPart[SELECT p.* AS &Person.*, (a.district, a.street) AS &Address.* FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = ] InputPart[Person.name] BypassPart[) UNION SELECT p.* AS &Person.*, (a.district, a.street) AS &Address.* FROM person WHERE p.name IN (SELECT name FROM table WHERE table.n = ] InputPart[Person.name] BypassPart[)]]",
```
TheSignPainter98: These are some very long lines...

internal/parse/parser_test.go
Aflynn50: The output parse isn't actually merged yet! These tests were originally designed for the whole parse which we wrote at once but have submitted as separate PRs (though its all changed a bit since then).

internal/parse/parser_test.go
Aflynn50: The aim is to be following this:
https://google.github.io/styleguide/go/decisions.html#comment-line-length
I would argue this falls in the category of lines that can be long because they're not read very often.

internal/parse/parser_test.go
TheSignPainter98: Ah, I understand. Thanks!

internal/parse/parser_test.go
TheSignPainter98: This seems like a good intuition, thanks for clarifying.

internal/expr/expr_test.go
```diff
@@ -234,6 +168,25 @@ func (s *ExprSuite) TestExpr(c *C) {
...
+	}{{
+		"SELECT street FROM t WHERE x = $Address.street",
+		"[Bypass[SELECT street FROM t WHERE x = ] Input[Address.street]]",
+	}, {
+		"SELECT p FROM t WHERE x = $Person.id",
+		"[Bypass[SELECT p FROM t WHERE x = ] Input[Person.id]]",
+	}}
+	for _, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.input)
```
solmyrcan, line 216: Just wondering if we could do something like this here, so that err gets logged only when it's not nil (or do we need to log it always no matter what its value is?), and apply the assertion only when err is nil. I might be missing a point though :)
```
if parsedExpr, err := parser.Parse(test.input); err != nil {
	c.Log(err)
} else {
	c.Assert(parsedExpr.String(), Equals, test.expectedParsed)
}
```

Aflynn50: Good point, I've changed it to:
```Go
if err != nil {
	c.Fatal(err)
}
c.Assert(parsedExpr.String(), Equals, test.expectedParsed)
```

niemeyer (approved):

niemeyer:
Thanks, Alastair.

PR #31, opened 2023-01-10
Full advance function

Aflynn50:
This PR adds a complete advance function.

The advance function passes over quoted strings and stops at interesting characters. When it finds them it passes control back to the main `Parse` loop.

This advance function changes the behaviour of the parser. Instead of quoted strings being handled in the parser loop they are handled here in the advance function (as a result of this the BypassParts in the tests had to be merged).

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -44,15 +44,41 @@ var tests = []struct {
 }, {
 	"multiple-quoted bypass expression",
 	`SELECT '''' AS &Person.*`,
-	`[Bypass[SELECT ] Bypass[''''] Bypass[ AS ] Output[[] [Person.*]]]`,
+	`[Bypass[SELECT '''' AS ] Output[[] [Person.*]]]`,
 }, {
 	"single quote in double quotes",
 	`SELECT "'" AS &Person.*`,
-	`[Bypass[SELECT ] Bypass["'"] Bypass[ AS ] Output[[] [Person.*]]]`,
+	`[Bypass[SELECT "'" AS ] Output[[] [Person.*]]]`,
+}, {
+	"spaces and tabs",
+	"SELECT p.* 	AS 		   &Person.*",
+	"[Bypass[SELECT ] Output[[p.*] [Person.*]]]",
+}, {
+	"new lines",
+	`SELECT
+		p.* AS &Person.*,
+		foo
+	 FROM t
+	 WHERE
+		foo = bar
+		and
+		x = y`,
+	`[Bypass[SELECT
+		] Output[[p.*] [Person.*]] Bypass[,
+		foo
+	 FROM t
+	 WHERE
+		foo = bar
+		and
+		x = y]]`,
 }, {
 	"quoted output expression",
 	"SELECT p.* AS &Person.*, '&notAnOutputExpresion.*' AS literal FROM t",
-	"[Bypass[SELECT ] Output[[p.*] [Person.*]] Bypass[, ] Bypass['&notAnOutputExpresion.*'] Bypass[ AS literal FROM t]]",
+	"[Bypass[SELECT ] Output[[p.*] [Person.*]] Bypass[, '&notAnOutputExpresion.*' AS literal FROM t]]",
+}, {
+	"quoted input expression",
+	"SELECT foo FROM t WHERE bar = '$NotAn.input'",
+	"[Bypass[SELECT foo FROM t WHERE bar = '$NotAn.input']]",
 }, {
 	"star as output",
 	"SELECT * AS &Person.* FROM t",
@@ -80,23 +106,23 @@ var tests = []struct {
 }, {
 	"output and quote",
 	"SELECT foo, bar, &Person.id FROM table WHERE foo = 'xx'",
-	"[Bypass[SELECT foo, bar, ] Output[[] [Person.id]] Bypass[ FROM table WHERE foo = ] Bypass['xx']]",
+	"[Bypass[SELECT foo, bar, ] Output[[] [Person.id]] Bypass[ FROM table WHERE foo = 'xx']]",
 }, {
 	"two outputs and quote",
 	"SELECT foo, &Person.id, bar, baz, &Manager.name FROM table WHERE foo = 'xx'",
-	"[Bypass[SELECT foo, ] Output[[] [Person.id]] Bypass[, bar, baz, ] Output[[] [Manager.name]] Bypass[ FROM table WHERE foo = ] Bypass['xx']]",
+	"[Bypass[SELECT foo, ] Output[[] [Person.id]] Bypass[, bar, baz, ] Output[[] [Manager.name]] Bypass[ FROM table WHERE foo = 'xx']]",
 }, {
 	"star as output and quote",
 	"SELECT * AS &Person.* FROM person WHERE name = 'Fred'",
-	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[ FROM person WHERE name = ] Bypass['Fred']]",
+	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[ FROM person WHERE name = 'Fred']]",
 }, {
 	"star output and quote",
 	"SELECT &Person.* FROM person WHERE name = 'Fred'",
-	"[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ FROM person WHERE name = ] Bypass['Fred']]",
+	"[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ FROM person WHERE name = 'Fred']]",
 }, {
 	"two star as outputs and quote",
 	"SELECT * AS &Person.*, a.* AS &Address.* FROM person, address a WHERE name = 'Fred'",
-	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[, ] Output[[a.*] [Address.*]] Bypass[ FROM person, address a WHERE name = ] Bypass['Fred']]",
+	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[, ] Output[[a.*] [Address.*]] Bypass[ FROM person, address a WHERE name = 'Fred']]",
 }, {
 	"multicolumn output",
 	"SELECT (a.district, a.street) AS (&Address.district, &Address.street) FROM address AS a",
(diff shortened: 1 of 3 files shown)


internal/parse/parser.go
```diff
@@ -129,25 +129,79 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
-		}
-
 		if p.pos == len(p.input) {
 			break
 		}
 
-		// If nothing above can be parsed we advance the parser.
-		p.pos++
+		// If nothing above can be parsed we advance the parser to the next
+		// charater of interest.
```
TheSignPainter98: `s/charater/character/` :)

internal/parse/parser.go
```diff
@@ -129,25 +129,79 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+			return nil, err
+		}
 	}
 	// Add any remaining unparsed string input to the parser.
 	p.add(nil)
 	return &ParsedExpr{p.parts}, nil
 }
 
+// advance increments p.pos until we reach content that may be the start of a
+// token we want to parse.
```
TheSignPainter98: I'm not sure about using 'we' here, it seems like `p.pos` which is reaching the start of an interesting token. I think 'it' makes more sense, unless this is a deliberate convention, as I see this appearing in a few places :P

internal/parse/parser.go
```diff
@@ -129,25 +129,79 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+		}
+	}
+
+	p.skipSpaces()
+
+	return nil
+
+}
+
+func (p *Parser) ignoreStringLiteral() (bool, error) {
```
TheSignPainter98: I think `skipStringLiteral` might be more consistent with `*Parser.skipSpaces`. Also does this function need a doc comment?

internal/parse/parser.go
```diff
@@ -129,25 +129,79 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+			// If this looks like a closing quote, check if it might be an
+			// escape for a following quote. If not, we're done.
+			if maybeCloser && !p.peekByte(c) {
+				return true, nil
+			}
+			maybeCloser = !maybeCloser
+		}
+
+		// Reached end of string and didn't find the closing quote
+		return false, fmt.Errorf("missing right quote for char %d in string literal", mark)
```
TheSignPainter98: I think calling it a 'closing' quote here would be more consistent with the comments. I don't recall seeing a 'left' quote being mentioned ;)

Also I think 'at char n' reads slightly more clearly than 'for char n'

internal/parse/parser_internal_test.go
```diff
@@ -64,3 +64,50 @@ func TestRunTable(t *testing.T) {
...
+		} else if s == unfinishedQuote {
+			if err == nil {
+				t.Errorf("test failed. the string %s was parsed but is not valid", q)
+			}
+		} else {
+			if ok {
+				t.Errorf("test failed. %s is not a valid quoted string but is recognised as one", q)
+			}
+		}
+	}
```
TheSignPainter98: If this is the only place `validQuote`, `unfinishedQuote` and `invalidQuote` will be used, it could be simpler to break `quoteState` and the for loop into different parts. This loop also doesn't always test `err` (although this may be intentional). Also I'm not sure the parser should be shared between the same tests?

Perhaps something like this?

```go
validQuotes := []string {
    `'stringy string'`,
    `'O''Flan'`,
    ...
}
for _, q := range validQuotes {
    p := NewParser()
    p.init(q)
    if ok, err := p.ignoreStringLiteral(); err != nil {
        ...
    } else if !ok {
        ...
    }
}
// Then similar code for slices perhaps named `unfinishedQuotes` and `invalidQuotes`

File: internal/parse/parser.go (hunk 1 above)

File: internal/parse/parser.go (hunk 2 above)

File: internal/parse/parser.go (hunk 3 above)

File: internal/parse/parser.go (hunk 4 above)

File: internal/parse/parser_internal_test.go (hunk 5 above)
Comment (Aflynn50) [reply internal/parse/parser_internal_test.go]: Changed. Now that I've read the google style guide I very much agree!

File: internal/expr/expr_test.go [hunk 6]
```diff
@@ -258,22 +286,14 @@ func (s *ExprSuite) TestParseEmptyStringLiteral(c *C) {
 	c.Assert(err, IsNil)
 }
 
-// Detect bad escaped string literal
-func (s *ExprSuite) TestParseBadEscaped(c *C) {
-	sql := "SELECT foo FROM t WHERE x = 'O'Donnell'"
-	parser := expr.NewParser()
-	_, err := parser.Parse(sql)
-	c.Assert(err, ErrorMatches, "cannot parse expression: missing right quote in string literal")
```
solmyrcan: "closing" rather than "right" here in the error string? :)

internal/expr/expr_test.go
```diff
@@ -258,22 +286,14 @@ func (s *ExprSuite) TestParseEmptyStringLiteral(c *C) {
...
 		"SELECT foo FROM t WHERE x = $Address.",
 		"SELECT foo FROM t WHERE x = $Address.&d",
 		"SELECT foo FROM t WHERE x = $Address.-",
 	}
 
-	for _, sql := range testListInvalidId {
+	for _, sql := range testList {
 		parser := expr.NewParser()
 		expr, err := parser.Parse(sql)
 		c.Assert(err, ErrorMatches, "cannot parse expression: invalid identifier near char 37")
```
solmyrcan: "near" or "at"? or is it possible for it not to be exactly at that char?

internal/expr/parser.go
```diff
@@ -129,25 +129,81 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+	}
+
+	for p.pos < len(p.input) {
+		if ok, err := p.skipStringLiteral(); err != nil {
+			return err
+		} else if ok {
+			continue
+		}
+		p.pos++
+		if p.pos < len(p.input) && delimiterBytes[p.input[p.pos]] {
```
solmyrcan: Maybe we don't need `p.pos < len(p.input)` here. Because when `p.pos == len(p.input)`, the loop will break immediately after anyway :)
This made me think: when `p.pos == len(p.input)`, are any of these delimiter characters supposed to be not there? In other words, is it an error to find any of these characters at the end of the input? (e.g. '=') I may be missing the context here of course :)

internal/expr/parser_helper_test.go
```diff
@@ -64,3 +69,64 @@ func (s *ExprInternalSuite) TestRunTable(c *C) {
...
+		}
+	}
+}
+
+func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
+	var p = NewParser()
+
+	unfinishedQuotes := []string{
+		`"`,
+		`' ''`,
```
solmyrcan, line 120: maybe add these as well:
```
`'`
`'string`
```

internal/expr/parser.go
```diff
@@ -129,25 +129,81 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+
+// skipStringLiteral jumps over single and double quoted sections of input.
+// Doubled up quotes are escaped.
+func (p *Parser) skipStringLiteral() (bool, error) {
+	mark := p.pos
+	if p.skipByte('"') || p.skipByte('\'') {
+		c := p.input[p.pos-1]
+
+		// We keep track of whether the next quote has been previously
+		// escaped. If not, it might be a closer.
```
solmyrcan: Might be easier to understand, if we change the comment to: 
"If not, it might a closing quote."

internal/expr/parser.go
```diff
@@ -129,25 +129,81 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+		maybeCloser := true
```
solmyrcan, line 191: I had trouble figuring out what the `for` loop was doing because I thought something was "getting closer" :P I was under the wrong impression with the variable name. Maybe if we could change that boolean's name, it might be easier to read the logic of the for loop that follows:

`isClosingQuote` rather than `maybeCloser`

internal/expr/expr_test.go
Aflynn50: Yes, though this is code thats getting removed rather than added, we are adding `closing`.

internal/expr/expr_test.go
Aflynn50: We can't be that exact because this error will be thrown when the parser gives up trying to parse the identifier. Also since its invalid, its hard to tell where it is intended to start and end, so its safest just to say near.

internal/expr/parser.go
Aflynn50: We have it here because of the `p.pos++` on the line before.

Generally we dont worry about any characters we don't recognise as our own. If the user has put an `=` at the end of the SQL then that's their problem.

internal/expr/parser_helper_test.go
Aflynn50: Done.

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: The reason it's "maybe" is becuase it may not be a closer. We might have a string `"string"` where the second quote looks like a closer but then we keep reading and its actually `"string""keeps going"` so quote we found at first was actually escaped (because in SQL you escape a quote by doubling it up).

internal/expr/expr_test.go
solmyrcan: ah that's right :P

internal/expr/parser.go
solmyrcan: If I'm getting it correctly, `skipStringLiteral()` returns `false, nil` when the byte at the current position is not a single or double quote. In such a case, it doesn't skip that byte. I believe, after the `if-else` block here, it's this same case where `ok == false` and `err==nil` from `p.skipStringLiteral()`. At that point, perhaps `p.pos++` shouldn't be here, because we then end up not checking the character at the current position (which wasn't skipped by `skipStringLiteral()` at this point. So it feels like perhaps `p.pos++` needs to be after we check the byte at the current position for the set of delimiters listed, i.e. after this `if-block` (and in that case, `p.pos < len(p.input)` in the `if-condition` seems to be redundant, because the `for` loop condition has already checked it.

internal/expr/parser.go
solmyrcan: Ah I didn't know that - thanks :) I guess what confused me was the word "closer", which made me think we were getting closer to something as we progressed. How about `maybeClosingQuote` :P - all fine though either way.

internal/expr/parser.go
Aflynn50: Fixed with a small shuffle around.

internal/expr/parser.go
```diff
@@ -135,24 +144,73 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 			p.add(in)
 			continue
 		}
+	}
 
-		if bypass, ok, err := p.parseStringLiteral(); err != nil {
-			return nil, err
+	// Add any remaining unparsed string input to the parser.
+	p.partStart = p.pos
```
niemeyer: If line 132 above is moved up before the break, it looks like this can go away.

internal/expr/parser.go
```diff
@@ -135,24 +144,73 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+	p.add(nil)
+	return &ParsedExpr{p.parts}, nil
+}
+
+// advance increments p.pos until it reaches content that might preceed a token
+// we want to parse.
+func (p *Parser) advance() error {
+
+	// The byte following these bytes might occur before the start of an expression.
+	delimiterBytes := map[byte]bool{
```
niemeyer: This should really be an inline switch below.
```
switch ... {
case '\n', ';', ...:
    ...
}
```
It will be orders of magnitude faster.

internal/expr/parser.go
```diff
@@ -135,24 +144,73 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+		' ':  true,
+		'\t': true,
+		'\n': true,
+		'\r': true,
+		')':  true,
```
niemeyer: This is a curious one. Do we accept the variables directly after the parenthesis, or is there some other case I'm missing?

I believe this list is also missing quite a bit of content (math ops, for instance). It's okay to move on like this and grow the map in future PRs with proper testing, though.

internal/expr/parser.go
```diff
@@ -135,24 +144,73 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+
+	return nil
+
+}
+
+// skipStringLiteral jumps over single and double quoted sections of input.
+// Doubled up quotes are escaped.
+func (p *Parser) skipStringLiteral() (bool, error) {
+	mark := p.pos
+	if p.skipByte('"') || p.skipByte('\'') {
```
niemeyer, line 186: This is missing save/restore, which means on false this function will corrupt the state.

internal/expr/parser.go
```diff
@@ -135,24 +144,73 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
...
+func (p *Parser) advance() error {
+
+	// The byte following these bytes might occur before the start of an expression.
+	delimiterBytes := map[byte]bool{
+		' ':  true,
+		'\t': true,
+		'\n': true,
+		'\r': true,
+		')':  true,
+		';':  true,
```
niemeyer: Also a curious one. Where do we see this?  If it's splitting statements, we don't expect our variables to start a statement.

niemeyer (changes requested):
Looks nice, Alastair, thanks. A few comments below, and probably the last pass:

internal/expr/parser.go
Aflynn50: True: changed.

internal/expr/parser.go
Aflynn50: I think you're right that the closing bracket isnt needed here.

We certainly can add more in the future, it'll be simple enough to do.

internal/expr/parser.go
Aflynn50: I can't see how it would corrupt the state unless I'm missing something. On false both of the `p.skipByte` have returned false and thus not changed the state, we then return immediately.

We do return false and an error on line 201 which would result in a corrupted state but since it is an error state I assume this is ok. There are other places in the parser (`parseFullName` or `parseList`) where we return with an error and a corrupted state.

internal/expr/parser.go
Aflynn50: True, that's not needed.

internal/expr/parser.go
niemeyer: It's not okay, even if there is an error result, because it's up to the call site to decide if that's a fatal error or not. If any of these functions don't save/restore, though, we take away the ability of the call site to decide for itself if errors are final or if there's a an alternative to be attempted for the given scenario.

It's also much easier to keep a strong invariant in mind than doing pinpointed decisions everywhere. The invariant here is: if it fails the state is untouched.

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: Ok, I've changed it. I'll create a new PR to preserve it in other places to.

niemeyer (approved):
Thanks for the tuning.

PR #32, opened 2023-01-12
Add export_test.go

Aflynn50:
This PR moves all black box tests to expr_test and adds export_test.go to allow for some testing of internal parts.

Certain tests are also updated to take advantage of the `go check` library.


niemeyer (approved):
We've discussed this one over a call. Thanks!

PR #33, opened 2023-01-17
Add Prepare stage for outputs

Aflynn50:
This PR adds the prepare stage for outputs. The prepare function takes instansiations of all the types mentioned in the output expressions of the query. The output expressions are checked against these types. If an output expression contains a * this is substituted for a list of all the tags in the output expression struct.

All output expression columns are given an alias with prefix `_sqlair` and suffix `_n` where `n` is the position of the output in the query. This will be used in the scan/decode stage to retrieve the relevant columns.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -538,7 +538,7 @@ func (s *ExprSuite) TestPrepareMixedTypes(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: multiple types in single output expression",
+		c.Assert(err, ErrorMatches, "cannot prepare expression: multiple target types in: .*",
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
@@ -565,7 +565,7 @@ func (s *ExprSuite) TestPrepareAsteriskMix(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid mix of asterisk and none asterisk columns in output expression",
+		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid asterisk columns in: .*",
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
@@ -589,7 +589,7 @@ func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: mismatched number of cols and targets in output expression",
+		c.Assert(err, ErrorMatches, "cannot prepare expression: mismatched number of cols and targets in: .*",
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -415,17 +415,21 @@ func FuzzParser(f *testing.F) {
 
 func (s *ExprSuite) TestPrepareMismatchedStructName(c *C) {
 	testList := []struct {
-		sql     string
-		structs []any
+		sql      string
+		structs  []any
+		errorstr string
 	}{{
-		sql:     "SELECT street FROM t WHERE x = $Address.street",
-		structs: []any{Person{ID: 1}},
+		sql:      "SELECT street FROM t WHERE x = $Address.street",
+		structs:  []any{Person{ID: 1}},
+		errorstr: "cannot prepare expression: type Address unknown, have: Person",
 	}, {
-		sql:     "SELECT street AS &Address.street FROM t",
-		structs: []any{},
+		sql:      "SELECT street AS &Address.street FROM t",
+		structs:  []any{},
+		errorstr: "cannot prepare expression: type Address unknown, have: ",
 	}, {
-		sql:     "SELECT street AS &Address.id FROM t",
-		structs: []any{Person{ID: 1}},
+		sql:      "SELECT street AS &Address.id FROM t",
+		structs:  []any{Person{ID: 1}},
+		errorstr: "cannot prepare expression: type Address unknown, have: Person",
 	}}
 
 	for i, test := range testList {
@@ -435,7 +439,7 @@ func (s *ExprSuite) TestPrepareMismatchedStructName(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, `cannot prepare expression: type [a-zA-Z_\.0-9]+ unknown, have: .*`,
+		c.Assert(err, ErrorMatches, test.errorstr,
 			Commentf("test %d failed:\nsql: '%s'\nstructs: '%+v'", i, test.sql, test.structs))
 	}
 
@@ -443,17 +447,21 @@ func (s *ExprSuite) TestPrepareMismatchedStructName(c *C) {
 
 func (s *ExprSuite) TestPrepareMissingTag(c *C) {
 	testList := []struct {
-		sql     string
-		structs []any
+		sql      string
+		structs  []any
+		errorstr string
 	}{{
-		sql:     "SELECT street FROM t WHERE x = $Address.number",
-		structs: []any{Address{ID: 1}},
+		sql:      "SELECT street FROM t WHERE x = $Address.number",
+		structs:  []any{Address{ID: 1}},
+		errorstr: `cannot prepare expression: type Address has no "number" db tag`,
 	}, {
-		sql:     "SELECT (street, road) AS &Address.* FROM t",
-		structs: []any{Address{ID: 1}},
+		sql:      "SELECT (street, road) AS &Address.* FROM t",
+		structs:  []any{Address{ID: 1}},
+		errorstr: `cannot prepare expression: type Address has no "road" db tag`,
 	}, {
-		sql:     "SELECT &Address.road FROM t",
-		structs: []any{Address{ID: 1}},
+		sql:      "SELECT &Address.road FROM t",
+		structs:  []any{Address{ID: 1}},
+		errorstr: `cannot prepare expression: type Address has no "road" db tag`,
 	}}
 
 	for i, test := range testList {
@@ -463,7 +471,7 @@ func (s *ExprSuite) TestPrepareMissingTag(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, `cannot prepare expression: type [a-zA-Z_\.0-9]+ has no "[a-zA-Z_\.0-9]+" db tag`,
+		c.Assert(err, ErrorMatches, test.errorstr,
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 
 	}
@@ -471,17 +479,21 @@ func (s *ExprSuite) TestPrepareMissingTag(c *C) {
 
 func (s *ExprSuite) TestPrepareInvalidAsteriskPlacement(c *C) {
 	testList := []struct {
-		sql     string
-		structs []any
+		sql      string
+		structs  []any
+		errorstr string
 	}{{
-		sql:     "SELECT (&Person.*, &Person.*) FROM t",
-		structs: []any{Address{}, Person{}},
+		sql:      "SELECT (&Person.*, &Person.*) FROM t",
+		structs:  []any{Address{}, Person{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[] [Person.* Person.*]]",
 	}, {
-		sql:     "SELECT (p.*, t.*) AS &Address.* FROM t",
-		structs: []any{Address{}},
+		sql:      "SELECT (p.*, t.*) AS &Address.* FROM t",
+		structs:  []any{Address{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[p.* t.*] [Address.*]]",
 	}, {
-		sql:     "SELECT p.* AS &Address.street FROM t",
-		structs: []any{Address{}},
+		sql:      "SELECT p.* AS &Address.street FROM t",
+		structs:  []any{Address{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[p.*] [Address.street]]",
 	}}
 
 	for i, test := range testList {
@@ -491,24 +503,28 @@ func (s *ExprSuite) TestPrepareInvalidAsteriskPlacement(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid asterisk in output expression: .*",
+		c.Assert(err.Error(), Equals, test.errorstr,
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
 
 func (s *ExprSuite) TestPrepareAsteriskMix(c *C) {
 	testList := []struct {
-		sql     string
-		structs []any
+		sql      string
+		structs  []any
+		errorstr string
 	}{{
-		sql:     "SELECT (&Address.*, &Address.id) FROM t",
-		structs: []any{Address{}, Person{}},
+		sql:      "SELECT (&Address.*, &Address.id) FROM t",
+		structs:  []any{Address{}, Person{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[] [Address.* Address.id]]",
 	}, {
-		sql:     "SELECT (p.*, t.name) AS &Address.* FROM t",
-		structs: []any{Address{}},
+		sql:      "SELECT (p.*, t.name) AS &Address.* FROM t",
+		structs:  []any{Address{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[p.* t.name] [Address.*]]",
 	}, {
-		sql:     "SELECT (name, p.*) AS (&Person.id, &Person.*) FROM t",
-		structs: []any{Address{}, Person{}},
+		sql:      "SELECT (name, p.*) AS (&Person.id, &Person.*) FROM t",
+		structs:  []any{Address{}, Person{}},
+		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[name p.*] [Person.id Person.*]]",
 	}}
 
 	for i, test := range testList {
@@ -518,21 +534,24 @@ func (s *ExprSuite) TestPrepareAsteriskMix(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid asterisk in output expression: .*",
+		c.Assert(err.Error(), Equals, test.errorstr,
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
 
 func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
 	testList := []struct {
-		sql     string
-		structs []any
+		sql      string
+		structs  []any
+		errorstr string
 	}{{
-		sql:     "SELECT (p.name, t.id) AS &Address.id FROM t",
-		structs: []any{Address{}},
+		sql:      "SELECT (p.name, t.id) AS &Address.id FROM t",
+		structs:  []any{Address{}},
+		errorstr: "cannot prepare expression: mismatched number of cols and targets in output expression: Output[[p.name t.id] [Address.id]]",
 	}, {
-		sql:     "SELECT p.name AS (&Address.district, &Address.street) FROM t",
-		structs: []any{Address{}},
+		sql:      "SELECT p.name AS (&Address.district, &Address.street) FROM t",
+		structs:  []any{Address{}},
+		errorstr: "cannot prepare expression: mismatched number of cols and targets in output expression: Output[[p.name] [Address.district Address.street]]",
 	}}
 
 	for i, test := range testList {
@@ -542,7 +561,7 @@ func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
 			c.Fatal(err)
 		}
 		_, err = parsedExpr.Prepare(test.structs...)
-		c.Assert(err, ErrorMatches, "cannot prepare expression: mismatched number of cols and targets in output expression: .*",
+		c.Assert(err.Error(), Equals, test.errorstr,
 			Commentf("test %d failed:\nsql: '%s'\nstructs:'%+v'", i, test.sql, test.structs))
 	}
 }
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -46,13 +46,13 @@ var tests = []struct {
 	"SELECT p.* AS &Person.*",
 	"[Bypass[SELECT ] Output[[p.*] [Person.*]]]",
 	[]any{Person{}},
-	"SELECT p.address_id AS [a-zA-Z_0-9]+, p.id AS [a-zA-Z_0-9]+, p.name AS [a-zA-Z_0-9]+",
+	"SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
 	"spaces and tabs",
 	"SELECT p.* 	AS 		   &Person.*",
 	"[Bypass[SELECT ] Output[[p.*] [Person.*]]]",
 	[]any{Person{}},
-	"SELECT p.address_id AS [a-zA-Z_0-9]+, p.id AS [a-zA-Z_0-9]+, p.name AS [a-zA-Z_0-9]+",
+	"SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
 	"new lines",
 	`SELECT
@@ -73,7 +73,7 @@ var tests = []struct {
 		x = y]]`,
 	[]any{Person{}},
 	`SELECT
-		p.address_id AS [a-zA-Z_0-9]+, p.id AS [a-zA-Z_0-9]+, p.name AS [a-zA-Z_0-9]+,
+		p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2,
 		foo
 	 FROM t
 	 WHERE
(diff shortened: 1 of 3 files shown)


internal/expr/expr_test.go
```diff
@@ -297,3 +314,144 @@ func FuzzParser(f *testing.F) {
...
+		sql:     "SELECT street AS &Address.street FROM t",
+		structs: []any{},
+	}, {
+		sql:     "SELECT street AS &Address.id FROM t",
+		structs: []any{Person{ID: 1}},
+	}}
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
```
solmyrcan, line 437: Perhaps change to `parsedExpr, _ :=` , since `err` gets over-written in the next line.

internal/expr/expr_test.go
```diff
@@ -297,3 +314,144 @@ func FuzzParser(f *testing.F) {
...
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, `cannot prepare expression: unknown type: .*`,
+			Commentf("test %d failed:\nsql: '%s'\nstructs: '%+v'\np:'%s'", i, test.sql, test.structs, parsedExpr.String()))
```
solmyrcan: Perhaps change `p:'%s'` to something bit more readable like `parsed to: '%s'`

internal/expr/expr_test.go
```diff
@@ -297,3 +314,144 @@ func FuzzParser(f *testing.F) {
...
+		sql:     "SELECT (street, road) AS &Address.* FROM t",
+		structs: []any{Address{ID: 1}},
+	}, {
+		sql:     "SELECT &Address.road FROM t",
+		structs: []any{Address{ID: 1}},
+	}}
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
```
solmyrcan, line 469: Same here - perhaps change to `parsedExpr, _ :=` , since `err` gets over-written in the next line. If it makes sense, then it's worth checking the rest of the file for similar lines :)

internal/expr/expr_test.go
```diff
@@ -297,3 +314,144 @@ func FuzzParser(f *testing.F) {
...
+		c.Assert(preparedExpr.SQL, Equals, test.expectedPrepared)
+	}
+}
+
+func (s *ExprSuite) TestPrepareMismatchedStructName(c *C) {
+	testList := []struct {
+		sql     string
+		structs []any
+	}{{
+		sql:     "SELECT street FROM t WHERE x = $Address.street",
```
solmyrcan: Would it make sense to add a one-liner brief comment for each of these test cases to say why they must fail?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
+// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// its table name. parseColumn returns an error so that it can be used with
```
solmyrcan: The code doesn't seem to return an error :) Perhaps we need to replace `nil` with an `error` in error cases.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
 }
 
-// parseInputExpression parses an input expression of the form $Type.name.
+// parseList takes a parsing function that returns a fullName and parses a
+// bracketed, comma seperated, list.
+func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+	cp := p.save()
+
+	if !p.skipByte('(') {
+		return nil, false, nil
```
solmyrcan: Should we return an error here as done elsewhere, replacing the last `nil` with an `fmt.Errorf()` with an relevant error message?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	var objs []fullName
+	for i := 0; nextItem; i++ {
+		if obj, ok, err := parseFn(p); ok {
+			objs = append(objs, obj)
+		} else if err != nil {
+			return nil, false, err
+		} else if i == 0 {
+			// If the first item is not what we are looking for, we exit.
+			cp.restore()
+			return nil, false, nil
```
solmyrcan: Should we return an error here as done elsewhere, replacing the last `nil` with an `fmt.Errorf()` with an relevant error message?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		return []fullName{col}, true
+	}
+
+	// Case 2: Multiple columns e.g. "(p.name, p.id)".
+	if cols, ok, _ := p.parseList((*Parser).parseColumn); ok {
+		return cols, true
+	}
+
+	cp.restore()
+	return nil, false
```
solmyrcan: Would this method need to return an error when both `parseColumn()` and `parseList()` do not return a happy result? i.e. when both Case 1 and Case 2 are not applicable. Or, similar to what's done at `parseTargets()` below, would it need to return within Case 1 and Case 2 when an error occurs?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+}
+
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more references Go objects.
```
solmyrcan: Change to "Go object references" maybe? :)

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+// If the ampersand is not preceded by a space and succeeded by a name or
```
solmyrcan: Should this be "preceded by a space or an opening bracket and succeeded by a name" instead?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+// opening bracket the targets are not parsed.
+func (p *Parser) parseTargets() ([]fullName, bool, error) {
+	cp := p.save()
+
+	// Case 1: A single target e.g. "&Person.name".
+	if target, ok, err := p.parseTarget(); err != nil {
+		return nil, false, err
```
solmyrcan: Would we need to call `cp.restore()` before returning? Same thing in Case 2.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
+	cp := p.save()
+
+	// Case 1: There are no columns e.g. "&Person.*".
+	if targets, ok, err := p.parseTargets(); ok {
+		return &outputPart{[]fullName{}, targets}, true, nil
```
solmyrcan: Do we need to call `cp.restore()` before each return statement in this method?

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+	ParsedExpr *ParsedExpr
+	SQL        string
+}
+
+type typeNameToInfo map[string]*info
+
+// prepareInput checks that the input expression corresponds to a Go struct
+// passed to Prepare.
+func prepareInput(ti typeNameToInfo, p *inputPart) error {
+	if inf, ok := ti[p.source.prefix]; ok {
```
solmyrcan: Could this method be modified to follow the style of first checking the errors and then leaving the success case as the last one :)

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+		_, ok = inf.tagToField[t.name]
+		if !ok && t.name != "*" {
+			return nil, fmt.Errorf(`there is no tag with name "%s" in "%s"`, t.name, inf.structType.Name())
+		}
+	}
+
+	// Check asterisk are in correct places.
+
+	sct := starCount(p.target)
+	scc := starCount(p.source)
```
solmyrcan: `scs` (`s` for source) instead? :P

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if sct > 1 || scc > 1 || (scc == 1 && sct == 0) {
+		return nil, fmt.Errorf("invalid asterisk in output expression")
+	}
+
+	starTarget := sct == 1
+	starColumn := scc == 1
```
solmyrcan: Would `starSource` be a better fit as a name here?

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+	if starTarget {
+
+		inf, _ := ti[p.target[0].prefix]
+
+		// Case 1.1: Single star e.g. "t.* AS &P.*" or "&P.*"
+		if starColumn || lenS == 0 {
+			pref := ""
+
+			// Prepend table name. E.g. "t" in "t.* AS &P.*".
+			if lenS > 0 && p.source[0].prefix != "" {
```
solmyrcan: Do we need to return any error if `lenS >0` but there's no prefix? Or has it been checked at this point?

internal/expr/expr_test.go
TheSignPainter98: Alternatively, I think asserting that `err` is nil could also be good here :)

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
+// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
```
TheSignPainter98: I think 'column made up of name bytes' could be shortened to 'a column name'

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
+func (p *Parser) parseTarget() (fullName, bool, error) {
+	cp := p.save()
+
+	if p.skipByte('&') {
```
TheSignPainter98: I might be missing context here, but I recall that target names should be proceded by ` &`, does this need to handle the space? Or should it be handled elsewhere?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		return p.parseGoFullName()
+	}
+
+	cp.restore()
+	return fullName{}, false, nil
+}
+
 // parseGoFullName parses a Go type name qualified by a tag name (or asterisk)
-// of the form "TypeName.col_name". On success it returns the parsed fullName,
+// of the form "&TypeName.col_name". On success it returns the parsed FullName,
```
TheSignPainter98: If my comment above about ` &` is true, does this need to to mention the preceding space?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
 	}
+
 	cp.restore()
 	return fullName{}, false, nil
 }
 
-// parseInputExpression parses an input expression of the form $Type.name.
+// parseList takes a parsing function that returns a fullName and parses a
+// bracketed, comma seperated, list.
+func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
```
TheSignPainter98: This might be over-engineering somewhat, but I wonder whether the need to parse a comma-separated (or indeed _something_-separated) list may arise elsewhere. Perhaps this function could be generalised?
```go
func (p *Parser) parseList(parseFn func(p *Parser) (any, bool, error)) ([]any, bool, error)
```
Generics could be nice, but Go doesn't seem to support generic methods (yet?)

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+			return nil, false, fmt.Errorf("column %d: invalid expression", p.pos)
+		}
+
+		p.skipSpaces()
+		if p.skipByte(')') {
+			return objs, true, nil
+		}
+
+		nextItem = p.skipByte(',')
+		p.skipSpaces()
```
TheSignPainter98: Just to save one line of code, could this `p.SkipSpaces()` be moved to the top of the loop, then the `p.SkipSpaces()` line be removed?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	} else if ok {
+		return targets, true, nil
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
+// preceded by a space and followed by a name byte.
```
TheSignPainter98: I think 'byte' points to an implementation detail, maybe just 'followed by a name' would be clearer?

Also, this function doesn't list its failure conditions, like others :P

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+	}
+
+	cp.restore()
+	return nil, false
+}
+
+// parseTargets parses the part of the output expression following the
+// ampersand. This can be one or more references Go objects.
+// If the ampersand is not preceded by a space and succeeded by a name or
+// opening bracket the targets are not parsed.
```
TheSignPainter98: I'm not sure this note about the spaces and ampersand belongs here? This function doesn't parse these spaces itself, so perhaps it belongs elsewhere?

Also, I don't think this forces that a space should come before the first target? Is `SELECT&Person.name` rejected elsewhere?

internal/expr/expr_test.go
Aflynn50: Actually it'll probably be better if I put this in
```golang
if err != nil {
	c.Fatal(err)
}
```
Since we get a null pointer expression for some reason that did happen to be an error.

internal/expr/expr_test.go
Aflynn50: Ah, that was only there for my debugging! I'll remove it.

internal/expr/expr_test.go
Aflynn50: Same comment as above.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		return nil, false, err
+	} else if ok {
+		return targets, true, nil
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseOutputExpression parses all output expressions. The ampersand must be
```
TheSignPainter98: I think 'all' here is somewhat misleading, to me, it implies that this function could also parse expressions later on, even when they are out of place. Perhaps something like 
```go
// parseOutputExpression parses an output expression which may consist of multiple parts. ...
```

Also, I think "The ampersand _before identifiers_ must be preceded..." could be clearer perhaps?

internal/expr/expr_test.go
Aflynn50: Hopefully the error message should have enough context for these. I've just split this one into two separate tests so that the test titles give a bit more context too.

internal/expr/parser.go
Aflynn50: Ah this is saying it returns an error type, i.e. the parseList function takes this function as an argument so the type that parseList wants must match the type of this function. Even though we never actually need to return an error from this function.

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+// preceded by a space and followed by a name byte.
+func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
+	cp := p.save()
+
+	// Case 1: There are no columns e.g. "&Person.*".
+	if targets, ok, err := p.parseTargets(); ok {
+		return &outputPart{[]fullName{}, targets}, true, nil
+	} else if err != nil {
+		return nil, false, err
+	}
```
TheSignPainter98: Just a minor re-ordering note, should `err` be checked first and `ok` second?

internal/expr/parser.go
```diff
@@ -265,28 +272,174 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+
+	// Case 2: There are columns e.g. "p.col1 AS &Person.*".
+	if cols, ok := p.parseColumns(); ok {
+		p.skipSpaces()
+		if p.skipString("AS") {
+			p.skipSpaces()
```
TheSignPainter98: Does the return value of the second `skipSpaces` call need to be read? Otherwise, I don't think the space-before-ampersands condition is checked

internal/expr/parser.go
Aflynn50: When we return false but not an error in the parser its because we've not found the thing we are looking for. If we return an error then we have found it but we thing it is incorrectly formatted.
Here we return `nil, false` if we cant find an opening bracket that signifies the start of the list, i.e. we cant find a list to parse.

internal/expr/parser.go
Aflynn50: See comment above.

internal/expr/parser.go
Aflynn50: See comment above.

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: True, the syntax has changed since this was written.

internal/expr/parser.go
Aflynn50: No, because the state hasnt actually changed if we fail. Also its an error case here so the state is discarded anyway.

internal/expr/parser.go
Aflynn50: Its not actually needed here, since the functions dont change the state.

internal/expr/prepare.go
Aflynn50: Indeed it can, and it has! I just forgot to merge the branch it was done in.

internal/expr/prepare.go
Aflynn50: That does make sense. I was debating between the two and chose c for columns but I think s might be better.

internal/expr/prepare.go
Aflynn50: Yea probably. I've changed it.

internal/expr/prepare.go
TheSignPainter98: I'm having a hard time reading the code below, perhaps these could be renamed `targetStars` and `sourceStars`?

internal/expr/prepare.go
TheSignPainter98: I think there's a small mismatch with naming with `starColumn`, if `scc` is inlined, the variable and field name differ:
```go 
starColumn := starCount(p.source)
```

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if sct > 1 || scc > 1 || (scc == 1 && sct == 0) {
+		return nil, fmt.Errorf("invalid asterisk in output expression")
+	}
+
+	starTarget := sct == 1
+	starColumn := scc == 1
+
+	lenS := len(p.source)
+	lenT := len(p.target)
```
TheSignPainter98: I think `numSources` and `numTargets` might be clearer?

internal/expr/prepare.go
Aflynn50: Ah yes, I think this should actually be `if starSource`

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+		// Case 1.1: Single star e.g. "t.* AS &P.*" or "&P.*"
+		if starColumn || lenS == 0 {
+			pref := ""
+
+			// Prepend table name. E.g. "t" in "t.* AS &P.*".
+			if lenS > 0 && p.source[0].prefix != "" {
+				pref = p.source[0].prefix + "."
+			}
+
+			for tag := range inf.tagToField {
```
TheSignPainter98: Perhaps it's because I haven't seen this form so much, but instead of implicitly ignoring the value, is it clearer to write `for tag, _ := range ...`? I'm honestly not sure on this one, short is nice too

internal/expr/expr_test.go
solmyrcan: sounds good!

internal/expr/prepare.go
```diff
@@ -0,0 +1,183 @@
...
+	return outCols, nil
+}
+
+// Prepare takes a parsed expression and all the Go objects mentioned in it.
+// The IO parts of the statement are checked for validity against the Go objects
+// and expanded if necessary.
+func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot prepare expression: %s", err)
```
TheSignPainter98: I think I recall seeing this in an older PR, it still seems a little magical to me (and hence un-Go-like?)

Would it be cleaner to have a function to do this (name tbd) and to call it explicitly each time?
```go
func (pe *ParsedExpr) Prepare(...) (...) {
    errFmt := func(e error) error {
        return fmt.Errorf("cannot prepare expression: %s", e)
    }
    ...
        return nil, errFmt(err)
}
```

internal/expr/typeinfo_test.go
```diff
@@ -3,14 +3,10 @@ package expr
 import (
 	"reflect"
 	"sync"
-	"testing"
 
 	. "gopkg.in/check.v1"
 )
 
-// Hook up gocheck into the "go test" runner.
-func TestInternal(t *testing.T) { TestingT(t) }
```
TheSignPainter98: I assume this has been replaced elsewhere ;)

internal/expr/parser.go
Aflynn50: It's being handled in the advance function, which has yet to be approved/reviewed! I'll change this comment in that PR.

internal/expr/parser.go
Aflynn50: See above.

internal/expr/parser.go
Aflynn50: It probably would, though I doubt this library will have any further need of it. I would also hope that someone someone has already done this better haha.

internal/expr/parser.go
Aflynn50: True, it does change the behaviour slightly but since it enters an error state when it exits the loops I think its ok.

internal/expr/parser.go
Aflynn50: Name bytes here refers to the specific list of bytes that can make up names (which right now is particular to this library)

internal/expr/parser.go
Aflynn50: See above

internal/expr/parser.go

internal/expr/parser.go

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: True, this will be tightened up once the advance is in.

internal/expr/prepare.go
Aflynn50: That might get confusing with `starTarget` and `starSource`. Since their scope is only a few lines I think its ok as it is.

internal/expr/prepare.go

internal/expr/prepare.go

internal/expr/prepare.go
Aflynn50: I think the var names make it clear enough whats going on here. The unclear thing here is Go's for loop syntax tbh.

internal/expr/prepare.go
Aflynn50: It will, since both parse and prepare are actually in the internal subdirectory we will have a nice user friendly wrapper round the outside that brushes over confusing names like parse prepare and complete. But thats all for the future.

internal/expr/typeinfo_test.go
Aflynn50: Yes, it turns out you only need one for both the internal and external testing, even though they're in different packages (`expr` vs `expr_test`). All the test results were appearing twice.

internal/expr/parts.go
```diff
@@ -10,7 +14,7 @@ type queryPart interface {
 	String() string
 
 	// ToSQL returns the SQL representation of the part.
-	toSQL() string
+	toSQL([]string, int) string
```
niemeyer: As we discussed during the call, you're right that this is a bad smell. Let's just kill this method and move that trivial logic to the call site, and implement a private `part()` marker method that does the validation work.

Aflynn50: I changed this a few days ago, forgot to reply here!

internal/expr/prepare.go
```diff
@@ -41,7 +82,91 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
...
+	var ok bool
+
+	for i, t := range p.target {
+		if i == 0 {
+			info, ok = ti[t.prefix]
+			if !ok {
+				return nil, fmt.Errorf(`type %s unknown, have: %s`, t.prefix, strings.Join(getKeys(ti), ", "))
+			}
+		} else if t.prefix != info.structType.Name() {
+			return nil, fmt.Errorf("multiple target types in: %s", p.String())
```
niemeyer: As we discussed, looks like we could handle this well as it's just a different spelling of the individual parts (`one AS &Foo.one, two AS &Bar.two` vs. `(one, two) AS (&Foo.one, &Bar.two)`).

internal/expr/prepare.go
```diff
@@ -41,7 +82,91 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
...
+				if _, ok := info.tagToField[c.name]; !ok {
+					return nil, fmt.Errorf(`type %s has no %q db tag`, info.structType.Name(), c.name)
+				}
+				outCols = append(outCols, c)
+			}
+			return outCols, nil
+		}
+	}
+
+	// Case 2: None star target cases e.g. "...&(P.name, P.id)".
```
niemeyer: Typo in the comment I think (`(&P...` vs `&(P...`). Same below.

internal/expr/prepare.go
```diff
@@ -77,12 +204,30 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
 		case *outputPart:
+			outCols, err := prepareOutput(ti, p)
+			if err != nil {
+				return nil, err
+			}
+			startCol := n
+			for i, c := range outCols {
+				sql.WriteString(c.String())
+				sql.WriteString(" AS _sqlair_")
+				sql.WriteString(c.name)
```
niemeyer: Per conversation, I think we can use just `_sqlair_1`, without the name, and then have a private slice that maps from those indexes into the fields.

internal/expr/expr_test.go
```diff
@@ -36,25 +36,23 @@ type District struct {
...
-}, {
-	"multiple-quoted bypass expression",
-	`SELECT '''' AS &Person.*`,
-	`[Bypass[SELECT '''' AS ] Output[[] [Person.*]]]`,
-}, {
-	"single quote in double quotes",
-	`SELECT "'" AS &Person.*`,
-	`[Bypass[SELECT "'" AS ] Output[[] [Person.*]]]`,
+	[]any{Person{}},
+	"SELECT p.address_id AS [a-zA-Z_0-9]+, p.id AS [a-zA-Z_0-9]+, p.name AS [a-zA-Z_0-9]+",
```
niemeyer: Given the other comment about simplifying the field name to `_sqlair_1` and so on, it would be nice to see those in the tests so we can both understand how it works here and also have logic that ensures our logic is right.

niemeyer (changes requested):
Looking good! Just a few details:

internal/expr/expr_test.go
Aflynn50: Changed. They now don't use regex.

internal/expr/prepare.go

internal/expr/prepare.go
Aflynn50: Fixed.

internal/expr/prepare.go
Aflynn50: Changed, this is supported now.

internal/expr/expr_test.go
```diff
@@ -36,25 +36,23 @@ type District struct {
...
-}, {
-	"multiple-quoted bypass expression",
-	`SELECT '''' AS &Person.*`,
-	`[Bypass[SELECT '''' AS ] Output[[] [Person.*]]]`,
-}, {
-	"single quote in double quotes",
-	`SELECT "'" AS &Person.*`,
-	`[Bypass[SELECT "'" AS ] Output[[] [Person.*]]]`,
+	[]any{Person{}},
+	"SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
```
niemeyer, line 49: Suddenly it starts to feel real. :)

internal/expr/expr_test.go
```diff
@@ -326,44 +413,136 @@ func FuzzParser(f *testing.F) {
...
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
+		if err != nil {
+			c.Fatal(err)
+		}
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, `cannot prepare expression: type [a-zA-Z_\.0-9]+ unknown, have: .*`,
+			Commentf("test %d failed:\nsql: '%s'\nstructs: '%+v'", i, test.sql, test.structs))
```
niemeyer, line 443: It would be nice to have the real error here, as that's the only way we'll know the error message is indeed correct. As it is, even if it's right today, we can break it tomorrow and not notice.

For that, have you considered adding an `error string` field on the big testing list above, and if it's set, we fail unless an error is found, and do `ErrroMatches` with the set string?

internal/expr/expr_test.go
```diff
@@ -326,44 +413,136 @@ func FuzzParser(f *testing.F) {
...
+	for i, test := range testList {
 		parser := expr.NewParser()
-		parsedExpr, _ := parser.Parse(test.input)
-		preparedExpr, err := parsedExpr.Prepare(test.prepareArgs...)
+		parsedExpr, err := parser.Parse(test.sql)
+		if err != nil {
+			c.Fatal(err)
+		}
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, `cannot prepare expression: type [a-zA-Z_\.0-9]+ has no "[a-zA-Z_\.0-9]+" db tag`,
```
niemeyer: Same here.

internal/expr/expr_test.go
```diff
@@ -326,44 +413,136 @@ func FuzzParser(f *testing.F) {
...
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
 		if err != nil {
 			c.Fatal(err)
 		}
-		c.Assert(preparedExpr.SQL, Equals, test.expectedPrepared)
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid asterisk in output expression: .*",
```
niemeyer: Same.

internal/expr/expr_test.go
```diff
@@ -326,44 +413,136 @@ func FuzzParser(f *testing.F) {
...
+	}}
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
+		if err != nil {
+			c.Fatal(err)
+		}
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, "cannot prepare expression: invalid asterisk in output expression: .*",
```
niemeyer: Same.

internal/expr/expr_test.go
```diff
@@ -326,44 +413,136 @@ func FuzzParser(f *testing.F) {
...
+	}}
+
+	for i, test := range testList {
+		parser := expr.NewParser()
+		parsedExpr, err := parser.Parse(test.sql)
+		if err != nil {
+			c.Fatal(err)
+		}
+		_, err = parsedExpr.Prepare(test.structs...)
+		c.Assert(err, ErrorMatches, "cannot prepare expression: mismatched number of cols and targets in output expression: .*",
```
niemeyer: Same.

internal/expr/prepare.go
```diff
@@ -41,7 +79,99 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
 	return nil
 }
 
-// Prepare takes a parsed expression and all the Go objects mentioned in it.
+// prepareOutput checks that the output expressions are correspond to a known types.
```
niemeyer: Some typos in this sentence.

internal/expr/prepare.go
```diff
@@ -41,7 +79,99 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
...
+// It then checks they are formatted correctly and finally generates the columns for the query.
+func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []outputDest, error) {
+
+	var outCols = make([]fullName, 0)
+	var outDests = make([]outputDest, 0)
+
+	// Check the asterisk are well formed (if present).
```
niemeyer: s/the asterisk/that asterisks/

internal/expr/prepare.go
```diff
@@ -41,7 +79,99 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
...
+		if t.name != "*" {
+			f, ok := info.tagToField[t.name]
+			if !ok {
+				return nil, nil, fmt.Errorf(`type %s has no %q db tag`, info.structType.Name(), t.name)
+			}
+			// For a none star expression we record output destinations here.
+			// For a star expression we fill out the destinations as we generate the columns.
+			outDests = append(outDests, outputDest{info.structType, f})
+
+		}
```
niemeyer, line 112: Nitpick: extra spaces before closings are generally not visually helpful.

internal/expr/prepare.go
```diff
@@ -41,7 +79,99 @@ func prepareInput(ti typeNameToInfo, p *inputPart) error {
...
+	}
+
+	// Generate columns to inject into SQL query.
+
+	// Case 1: Star target cases e.g. "...&P.*".
+	if p.target[0].name == "*" {
+		info, _ := ti[p.target[0].prefix]
+
+		// Case 1.1: Single star i.e. "t.* AS &P.*" or "&P.*"
```
niemeyer, line 121: Nice comments all around here, thanks.

niemeyer (approved):
Thanks, Alastair. The comments here are pretty much all cosmetic now, so I'm going ahead and merging this PR as it's blocking the follow ups from being clean. Can you please apply these trivials on an independent PR later?  Thanks again.

internal/expr/expr_test.go
Aflynn50: Changed.

internal/expr/expr_test.go

internal/expr/expr_test.go

internal/expr/expr_test.go

internal/expr/expr_test.go

internal/expr/prepare.go

internal/expr/prepare.go

internal/expr/prepare.go
Aflynn50: Changed.

PR #34, opened 2023-01-24
Add complete

Aflynn50:
This PR adds the complete stage. In this stage, we process any query arguments that are needed for execution. All structs mentioned in input expressions are passed to complete and the values that we need are extracted from them.

### DIFF
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -31,7 +31,7 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 		}
 		v := reflect.ValueOf(arg)
 		tv[v.Type()] = v
-		typeNames = append(typeNames, v.Type().String())
+		typeNames = append(typeNames, v.Type().Name())
 	}
 
 	// Query parameteres.
@@ -40,7 +40,7 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 	for i, in := range pe.inputs {
 		v, ok := tv[in.typ]
 		if !ok {
-			return nil, fmt.Errorf(`type %s not found, have: %s`, in.typ.String(), strings.Join(typeNames, ", "))
+			return nil, fmt.Errorf(`type %s not found, have: %s`, in.typ.Name(), strings.Join(typeNames, ", "))
 		}
 		named := sql.Named("sqlair_"+strconv.Itoa(i), v.FieldByIndex(in.field.index).Interface())
 		qargs = append(qargs, named)
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -1,0 +1,37 @@
+package expr
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type typeNameToValue = map[string]any
+
+// Complete gathers the query arguments that are specified in inputParts from
+// structs passed as parameters.
+func (pe *PreparedExpr) Complete(args ...any) ([]any, error) {
+	var tv = make(typeNameToValue)
+	for _, arg := range args {
+		if arg == (any)(nil) {
+			return nil, fmt.Errorf("nil parameter")
+		}
+		tv[reflect.TypeOf(arg).Name()] = arg
+	}
+
+	// Query parameteres.
+	qargs := []any{}
+
+	for _, p := range pe.inputs {
+		v, ok := tv[p.source.prefix]
+		if !ok {
+			return nil, fmt.Errorf(`type %s not passed as a parameter`, p.source.prefix)
+		}
+		qp, err := fieldValue(v, p.source)
+		if err != nil {
+			return nil, err
+		}
+		qargs = append(qargs, qp)
+	}
+
+	return qargs, nil
+}
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 )
 
 type CompletedExpr struct {
@@ -23,12 +24,14 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 	}()
 
 	var tv = make(map[reflect.Type]reflect.Value)
+	var typeNames []string
 	for _, arg := range args {
 		if arg == nil {
 			return nil, fmt.Errorf("nil parameter")
 		}
 		v := reflect.ValueOf(arg)
 		tv[v.Type()] = v
+		typeNames = append(typeNames, v.Type().String())
 	}
 
 	// Query parameteres.
@@ -37,7 +40,7 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 	for i, in := range pe.inputs {
 		v, ok := tv[in.typ]
 		if !ok {
-			return nil, fmt.Errorf(`type %s not passed as a parameter`, in.typ.Name())
+			return nil, fmt.Errorf(`type %s not found, have: %s`, in.typ.String(), strings.Join(typeNames, ", "))
 		}
 		named := sql.Named("sqlair_"+strconv.Itoa(i), v.FieldByIndex(in.field.index).Interface())
 		qargs = append(qargs, named)
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -69,5 +69,5 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 		qargs = append(qargs, named)
 	}
 
-	return &CompletedExpr{outputs: pe.outputs, sql: pe.SQL, args: qargs}, nil
+	return &CompletedExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -9,7 +9,7 @@ import (
 )
 
 type CompletedExpr struct {
-	outputs []loc
+	outputs []field
 	sql     string
 	args    []any
 }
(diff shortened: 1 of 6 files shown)


internal/expr/complete.go
```diff
@@ -0,0 +1,49 @@
...
+func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot complete expression: %s", err)
+		}
+	}()
+
+	var tv = make(typeNameToValue)
+	for _, arg := range args {
+		if arg == (any)(nil) {
```
TheSignPainter98: Is the `(any)` cast needed here?

internal/expr/complete.go
```diff
@@ -0,0 +1,49 @@
...
+			return nil, fmt.Errorf("cannot reflect nil value")
+		}
+		tv[reflect.TypeOf(arg).Name()] = arg
+	}
+
+	// Query parameteres.
```
TheSignPainter98, line 60: Minor typo `s/parameteres/parameters/`

internal/expr/complete.go
```diff
@@ -0,0 +1,49 @@
...
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot complete expression: %s", err)
+		}
+	}()
+
+	var tv = make(typeNameToValue)
+	for _, arg := range args {
+		if arg == (any)(nil) {
+			return nil, fmt.Errorf("cannot reflect nil value")
```
TheSignPainter98: Somehow 'reflect' seems too implementation-specific, perhaps `"got a nil argument"`?

internal/expr/prepare.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, t := range p.target {
+		outCols = append(outCols, t.name)
+	}
+	return outCols, nil
+}
+
+// Prepare takes a parsed expression and struct instantiations of all the type
+// mentioned in it.
+// The IO parts of the statement are checked for validity against the types
+// and expanded if necessary.
```
TheSignPainter98: Super two minor things---the last word on the first line should be 'types' and I think the line-break between the two sentences could be removed.

I'm not sure whether this is the PR to fix these, though,

internal/expr/complete.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		tv[reflect.TypeOf(arg).Name()] = arg
+	}
+
+	// Query parameteres.
+	qargs := []any{}
+
+	for _, p := range pe.ParsedExpr.queryParts {
+		if p, ok := p.(*inputPart); ok {
+			v, ok := tv[p.source.prefix]
```
TheSignPainter98: I realise this probably isn't something to solve here, but somehow the `prefix` field storing the name of a type feels odd. Perhaps calling this `typeName` would be more natural?

internal/expr/complete.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type CompletedExpr struct {
+	ParsedExpr  *ParsedExpr
+	InputValues []any
+}
+
+type typeNameToValue = map[string]any
+
+// Complete extracts query arguments specified in inputParts from structs.
```
TheSignPainter98: I misunderstood this as `inputParts` being extracted from structs, I think the terms need a little re-ordering, perhaps something like
```go
// Complete populates query arguments
```

Also the name 'Complete' seems a little vague to me? Perhaps it could be something like 'Populate' or 'Args'?

internal/expr/complete.go
Aflynn50: It doesn't seem like it, I copied this snippet from the reflection bit, not sure if its needed there either actually.

internal/expr/complete.go
Aflynn50: Changed

internal/expr/complete.go
Aflynn50: Changed

internal/expr/prepare.go
Aflynn50: True, they could look a lot better, I could fix it in the other PR.

internal/expr/complete.go
Aflynn50: Yea, I'm not happy with the names of these either. They were decided on at the start after a lot of back and forth though, so I think we have to wait until that's a distant memory before changing them again.

internal/expr/complete.go
Aflynn50: Changed it make it more clear

internal/expr/prepare.go
Aflynn50: Ah, just checked and it has already been fixed in the other PR. I guess the merge got confused or something.

internal/expr/complete.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+type CompletedExpr struct {
+	outputs []loc
+	SQL     string
+	Args    []any
```
niemeyer: Per conversation a couple of weeks ago, let's please keep the details of the expression internal for now. For testing we can have a method or a function in export_test.go.

internal/expr/complete.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Complete gathers the arguments that are specified in the query input parts from
+// structs passed as parameters.
+func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
```
niemeyer, line 24: The doc seems oriented a bit towards the internal implementation instead of the user of the function. Perhaps:
```
// Complete returns a completed expression ready for execution, using the provided values to
// substitute the input placeholders in the prepared expression. These placeholders use
// the syntax "$Person.fullname", where Person would be a type such as:
//
//        type Person struct {
//                Name string `db:"fullname"`
//        }
```

internal/expr/complete.go
```diff
@@ -0,0 +1,52 @@
...
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("parameter issue: %s", err)
```
niemeyer: Maybe: "invalid input parameters: %s"

internal/expr/complete.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}()
+
+	var tv = make(map[reflect.Type]reflect.Value)
+	var typeNames []string
```
niemeyer, line 37: Eitehr tv and tn, or typeValue and typeNames?   Latter seems better. :)

internal/expr/complete.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, arg := range args {
+		if arg == nil {
+			return nil, fmt.Errorf("nil parameter")
```
niemeyer: Maybe: "need valid struct values, got nil"

internal/expr/expr_test.go
```diff
@@ -565,3 +566,95 @@ func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
...
+	parsedExpr, err := parser.Parse(sql)
+	if err != nil {
+		c.Fatal(err)
+	}
+	preparedExpr, err := parsedExpr.Prepare(outerP)
+	if err != nil {
+		c.Fatal(err)
+	}
+	_, err = preparedExpr.Complete(shadowedP)
+	c.Assert(err, ErrorMatches, `parameter issue: type Person not found, have: Person`)
```
niemeyer: Instead of the confusing error message, we might special case this situation and if the name is already missing (in other words, we're definitely in an error case), we might iterate over the set of types to see if there is another type with the same name. If so, we might special case the error message along the lines of:
```
"invalid input parameters: expected type foobar.Type but got fooser.Type"
```

internal/expr/prepare.go
```diff
@@ -11,14 +11,14 @@ import (
 
 // PreparedExpr contains an SQL expression that is ready for execution.
 type PreparedExpr struct {
-	outputs []outputDest
-	inputs  []*inputPart
+	outputs []loc
+	inputs  []loc
 	SQL     string
```
niemeyer: I've overlooked this on previous PRs. I believe we also want this as an opaque type.

internal/expr/prepare.go
```diff
@@ -11,14 +11,14 @@ import (
...
 }
 
-type outputDest struct {
-	structType reflect.Type
-	field      field
+type loc struct {
+	typ   reflect.Type
```
niemeyer: The previous names seemed more clear and descriptive, even more for a type that is global to the package. Maybe the original meaning must change because you want it more generic, but would be nice to have a name that defines a bit better when we shoud expect to see this and for what purpose.

As we discussed, I think we can use just the `field` value itself here, as long as we have a backreference to its type.

niemeyer (changes requested):
Last pass, per our conversation. Looking good!

internal/expr/complete.go
Aflynn50: Changed.

internal/expr/complete.go
Aflynn50: Changed.

internal/expr/complete.go
Aflynn50: Changed.

internal/expr/complete.go
Aflynn50: Changed to the latter.

internal/expr/complete.go
Aflynn50: Changed.

internal/expr/expr_test.go
Aflynn50: Added special case.

internal/expr/prepare.go
Aflynn50: Changed and function in `export_test.go` added.

internal/expr/prepare.go
Aflynn50: Changed to use just `field`s and to store the struct type in the `field` struct defined in `info.go`

niemeyer (approved):
Thank you!

PR #36, opened 2023-02-07
Add support for `INSERT ... VALUES ...` input statements

Aflynn50:
This PR extends the syntax of the sqlair DSL to support bulk `INSERT` SQL DML statements. This will allow the user to write statments such as `INSERT INTO person (name, id) VALUES ($Person.*)`, or `INSERT INTO person (*) VALUES ($Person.*)`. This will allow for a more concise syntax when inserting multiple fields of structs.

We take advantage of the go/sql libraries support of named parameters to ensure the order of the inputs can be kept consistent in the complete stage. We name the parameters with their type and tag name. When we extract the parameter values from the structs at the complete stage we can use this information to match each value to to the corresponding input position.

We only parse the column and value information around the `VALUES` keyword, the `INSERT INTO` can be ignored. Specifically, this transforms the query as follows:
```
(*) VALUES ($P.*) 				             => (col1, col2, col3) VALUES (@Pcol1, @Pcol2, @Pcol3)
(col1, col2) VALUES ($P.*) 		    	             => (col1, col2) VALUES (@Pcol1, @Pcol2)
(col1) VALUES ($P.col1)                                      => (col1) VALUES (@Pcol1)
(mycol1, mycol2, mycol3) VALUES	 ($P.col1, $A.col1, $M.col1) => (mycol1, mycol2, mycol3) VALUES (@Pcol1, @Acol1, @Mcol1)
```
Note that the final two are of the same form but the last one uses multiple types and renames the columns.

The parsing resembles that of output expressions. Where they use the keyword `AS` here we use the keyword `VALUES`.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -342,14 +342,6 @@ func (p *Parser) parseColumn() (fullName, bool, error) {
 	return fullName{}, false, nil
 }
 
-func (p *Parser) parseTarget() (fullName, bool, error) {
-	if p.skipByte('&') {
-		return p.parseGoFullName()
-	}
-
-	return fullName{}, false, nil
-}
-
 // parseGoFullName parses a Go type name qualified by a tag name (or asterisk)
 // of the form "&TypeName.col_name".
 func (p *Parser) parseGoFullName() (fullName, bool, error) {
@@ -373,7 +365,7 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 
 // parseList takes a parsing function that returns a fullName and parses a
 // bracketed, comma seperated, list.
-func parseList[T any](p *Parser, parseFn func(p *Parser) (T, bool, error)) ([]T, bool, error) {
+func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
 	cp := p.save()
 	if !p.skipByte('(') {
 		return nil, false, nil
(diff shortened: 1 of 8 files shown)


internal/expr/parser.go
```diff
@@ -302,14 +302,14 @@ func (p *Parser) skipName() bool {
 //  - bool == false, err == nil
 //		The construct was not the one we are looking for
 
-// parseIdentifierAsterisk parses a name made up of only nameBytes or of a
-// single asterisk. On success it returns the parsed string and true. Otherwise,
-// it returns the empty string and false.
-func (p *Parser) parseIdentifierAsterisk() (string, bool) {
+// parseIdentifierAsterisk parses a name made up of only nameBytes or of a single asterisk.
+// It returns an error (always nil) so it may be used with parseList.
```
TheSignPainter98: I'm not sure the second line is necessary? It relates to the implementation of a different function, so it might be okay to omit it. Also 'always nil' feels like an implementation detail of this one

internal/expr/prepare.go
```diff
@@ -26,21 +36,145 @@ func getKeys(m map[string]*info) []string {
...
+	return s.String()
+}
+
+// prepareInput first checks the types mentioned in the expression are known, it
+// then checks the expression is valid and generates the SQL to print in place
+// of it.
+// As well as the SQL string it also returns a list of fullNames containing the
+// type and tag of each input parameter. These are used in the complete stage to
+// extract the arguments from the relevent structs.
+func prepareInput(ti typeNameToInfo, p *inputPart, n int) (string, []inputLocation, error) {
```
TheSignPainter98: I think this doc exposes a few implementation details, perhaps it could be simplified to something like this?
```go
// prepareInput generates SQL identifiers and associated input locations for the given input part.
```

internal/expr/parser.go
Aflynn50: It is an implementation detail, but I think its good to make it clear that the returned error here can be safely ignored, since thats very much unexpected behaviour.

internal/expr/prepare.go
Aflynn50: True, the comment is also incorrect haha so worth changing anyway

PR #37, opened 2023-02-09
Skip SQL comments

Aflynn50:
This PR augments the parser to skip SQL comments as specified here https://www.sqlite.org/lang_comment.html.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -79,6 +79,32 @@ var tests = []struct {
 		and
 		x = y`,
 }, {
+	"comments",
+	`SELECT &Person.* -- The line with &Person.* on it
+FROM person /* The start of a multi line comment
+It keeps going here with some weird chars /-*"/
+And now it stops */ WHERE "x" = /-*'' -- The "WHERE" line
+AND y =/* And now we have " */ "-- /* */" /* " some comments strings */
+AND z = $Person.id -- The line with $Person.id on it
+`,
+	`[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ 
+FROM person  WHERE "x" = /-*'' 
+AND y = "-- /* */" 
+AND z = ] Input[Person.id] Bypass[ 
+]]`,
+	[]any{Person{}},
+	`SELECT address_id AS _sqlair_0, id AS _sqlair_1, name AS _sqlair_2 
+FROM person  WHERE "x" = /-*'' 
+AND y = "-- /* */" 
+AND z = @sqlair_0 
+`,
+}, {
+	"comments v2",
+	`SELECT (&Person.name, /* ... */ &Person.id), (&Person.id /* ... */, &Person.name) FROM p -- End of the line`,
+	`[Bypass[SELECT ] Output[[] [Person.name Person.id]] Bypass[, ] Output[[] [Person.id Person.name]] Bypass[ FROM p ]]`,
+	[]any{Person{}},
+	`SELECT name AS _sqlair_0, id AS _sqlair_1, id AS _sqlair_2, name AS _sqlair_3 FROM p `,
+}, {
 	"quoted io expressions",
 	`SELECT "&notAnOutput.Expression" '&notAnotherOutputExpresion.*' AS literal FROM t WHERE bar = '$NotAn.Input' AND baz = "$NotAnother.Input"`,
 	`[Bypass[SELECT "&notAnOutput.Expression" '&notAnotherOutputExpresion.*' AS literal FROM t WHERE bar = '$NotAn.Input' AND baz = "$NotAnother.Input"]]`,
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -87,23 +87,27 @@ And now it stops */ WHERE "x" = /-*'' -- The "WHERE" line
 AND y =/* And now we have " */ "-- /* */" /* " some comments strings */
 AND z = $Person.id -- The line with $Person.id on it
 `,
-	`[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ 
-FROM person  WHERE "x" = /-*'' 
-AND y = "-- /* */" 
-AND z = ] Input[Person.id] Bypass[ 
+	`[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ -- The line with &Person.* on it
+FROM person /* The start of a multi line comment
+It keeps going here with some weird chars /-*"/
+And now it stops */ WHERE "x" = /-*'' -- The "WHERE" line
+AND y =/* And now we have " */ "-- /* */" /* " some comments strings */
+AND z = ] Input[Person.id] Bypass[ -- The line with $Person.id on it
 ]]`,
 	[]any{Person{}},
-	`SELECT address_id AS _sqlair_0, id AS _sqlair_1, name AS _sqlair_2 
-FROM person  WHERE "x" = /-*'' 
-AND y = "-- /* */" 
-AND z = @sqlair_0 
+	`SELECT address_id AS _sqlair_0, id AS _sqlair_1, name AS _sqlair_2 -- The line with &Person.* on it
+FROM person /* The start of a multi line comment
+It keeps going here with some weird chars /-*"/
+And now it stops */ WHERE "x" = /-*'' -- The "WHERE" line
+AND y =/* And now we have " */ "-- /* */" /* " some comments strings */
+AND z = @sqlair_0 -- The line with $Person.id on it
 `,
 }, {
 	"comments v2",
 	`SELECT (&Person.name, /* ... */ &Person.id), (&Person.id /* ... */, &Person.name) FROM p -- End of the line`,
-	`[Bypass[SELECT ] Output[[] [Person.name Person.id]] Bypass[, ] Output[[] [Person.id Person.name]] Bypass[ FROM p ]]`,
+	`[Bypass[SELECT ] Output[[] [Person.name Person.id]] Bypass[, ] Output[[] [Person.id Person.name]] Bypass[ FROM p -- End of the line]]`,
 	[]any{Person{}},
-	`SELECT name AS _sqlair_0, id AS _sqlair_1, id AS _sqlair_2, name AS _sqlair_3 FROM p `,
+	`SELECT name AS _sqlair_0, id AS _sqlair_1, id AS _sqlair_2, name AS _sqlair_3 FROM p -- End of the line`,
 }, {
 	"quoted io expressions",
 	`SELECT "&notAnOutput.Expression" '&notAnotherOutputExpresion.*' AS literal FROM t WHERE bar = '$NotAn.Input' AND baz = "$NotAnother.Input"`,
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -109,6 +109,61 @@ func (p *Parser) add(part queryPart) {
 	p.partStart = p.pos
 }
 
+// removeComments passes over the input removing comments.
+// It does not allow for nested comments.
+func (p *Parser) removeComments() error {
+	var res bytes.Buffer
+	var textStart = p.pos
+	var textEnd int
+
+	for p.pos < len(p.input) {
+		textEnd = p.pos
+		// Ignore comment chars in string literals.
+		if ok, err := p.skipStringLiteral(); err != nil {
+			return err
+		} else if ok {
+			continue
+		}
+
+		if p.skipByte('-') || p.skipByte('/') {
+			c := p.input[p.pos-1]
+			if (c == '-' && p.skipByte('-')) || (c == '/' && p.skipByte('*')) {
+				var end byte
+				if c == '-' {
+					end = '\n'
+				} else {
+					end = '*'
+				}
+				for p.pos < len(p.input) {
+					if p.input[p.pos] == end {
+						// if end == '\n' (i.e. its a -- comment) dont p.pos++ to keep the newline.
+						if end == '*' {
+							p.pos++
+							if !p.skipByte('/') {
+								continue
+							}
+						}
+						break
+					}
+					p.pos++
+				}
+
+				res.WriteString(p.input[textStart:textEnd])
+				textStart = p.pos
+				continue
+			}
+		} else {
+			// If we have skiped '-' or '/' but not found a comment don't do p.pos++.
+			p.pos++
+		}
+	}
+	// Add any remaining input
+	res.WriteString(p.input[textStart:])
+
+	p.init(res.String())
+	return nil
+}
+
 // Parse takes an input string and parses the input and output parts. It returns
 // a pointer to a ParsedExpr.
 func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
@@ -119,6 +174,11 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 	}()
 
 	p.init(input)
+
+	if err := p.removeComments(); err != nil {
+		return nil, err
+	}
+
 	for {
 		// Advance the parser to the start of the next expression.
 		if err := p.advance(); err != nil {
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -109,59 +109,39 @@ func (p *Parser) add(part queryPart) {
 	p.partStart = p.pos
 }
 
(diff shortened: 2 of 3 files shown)


internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
 		}
 	}
 }
+
+func (s *ExprInternalSuite) TestValidComments(c *C) {
+	var p = NewParser()
+
+	validComments := []string{
+		`-- Simple comment`,
```
solmyrcan: Maybe call it "Single-line"comment, which may be more descriptive :)

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+		`-- Simple with line break
+		`,
+		`/* muti
+		 line */`,
+		`/* unfinished mutiline`,
```
solmyrcan: mutiline -> multiline

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+		`/* -- */`,
+		`-- */`,
+	}
+
+	for _, q := range validComments {
+		p.init(q)
+		ok := p.skipComment()
+		if !ok {
+			c.Errorf("test failed. %s is a valid comment", q)
```
solmyrcan: maybe this could be a bit clearer message, what do you think?:
"test failed. Test outcome for valid comment %q is found to be invalid", q

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+	}
+}
+
+func (s *ExprInternalSuite) TestInvalidComments(c *C) {
+	var p = NewParser()
+
+	invalidComments := []string{
+		`- not comment`,
+		`- - not comment`,
+		`/ * not comment */`,
```
solmyrcan, line 151: Maybe add also: 
\``/ not comment */`\`
\``/* not comment /`\`

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+		`*/ not comment`,
+		`/- not comment`,
+		`-* not comment`,
+	}
+
+	for _, q := range invalidComments {
+		p.init(q)
+		ok := p.skipComment()
+		if ok {
+			c.Errorf("test failed. %s is an invalid comment", q)
```
solmyrcan: maybe this could be a bit clearer message, what do you think?:
"test failed. Test outcome for invalid comment %q is found to be valid", q

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,47 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+
+	validComments := []string{
+		`-- Simple comment`,
+		`-- Simple with line break
+		`,
+		`/* muti
+		 line */`,
+		`/* unfinished mutiline`,
+		`/* -- */`,
+		`-- */`,
```
TheSignPainter98, line 145: Perhaps it would be good to also test `"--"` and `"/*"`

internal/expr/parser_helper_test.go
TheSignPainter98: I think the second one is actually considered a valid comment, first makes sense though

internal/expr/expr_test.go
```diff
@@ -73,6 +73,20 @@ var tests = []struct {
...
+	 It keeps going here with some weird chars /-*"*/
+	 And now it stops */ WHERE x = /-*'' -- The WHERE line
+	 AND y = $Person.id -- The line with $Person.id on it
+	 `,
+	`[Bypass[SELECT ] Output[[] [Person.*]] Bypass[ -- The line with &Person.* on it
+	 FROM person /* The start of a multi line comment
+	 It keeps going here with some weird chars /-*"*/
+	 And now it stops */ WHERE x = /-*'' -- The WHERE line
+	 AND y = ] Input[Person.id] Bypass[ -- The line with $Person.id on it
+	 ]]`,
```
TheSignPainter98: I'm not sure whether the addition to `advance` handles this but either way, is it worth also testing `SELECT (&Person.foo, /* ... */ &Person.bar) ...` and `SELECT (&Person.foo /* ... */, &Person.bar) ...` and the same for single-line comments?

Aflynn50: These actually would break the parser. To fix this it would probably require some pretty big changes. The easiest way to do that would be to do an initial pass of the input string to remove comments. I'll have a chat with Gustavo about the approach to take here since if we kept this code as it is it would probably be good enough in 99% of use cases.

internal/expr/parser_helper_test.go
Aflynn50: Added the test cases

internal/expr/parser_helper_test.go
Aflynn50: Changed.

internal/expr/parser_helper_test.go
Aflynn50: Added the first one.

internal/expr/parser_helper_test.go
Aflynn50: Made a bit clearer.

internal/expr/parser_helper_test.go
Aflynn50: Changed.

internal/expr/parser_helper_test.go
Aflynn50: Changed.

internal/expr/expr_test.go
Aflynn50: Actually I take that back, the examples above would work, but this wouldn't `* AS (&Person.foo, /* ...*/, &Person.bar)`. This is because we would parse those as separate expressions but not this one.

internal/expr/expr_test.go
```diff
@@ -73,6 +73,22 @@ var tests = []struct {
 		foo = bar
 		and
 		x = y]]`,
+}, {
+	"comments",
+	`SELECT &Person.* -- The line with &Person.* on it
+FROM person /* The start of a multi line comment
+It keeps going here with some weird chars /-*"/
+And now it stops */ WHERE x = /-*'' -- The WHERE line
```
solmyrcan: Could be an interesting test to see what happens if two comments both include matching quotes, making everything in between the two quotes a string literal. 
For example, if we add a quote in the single line comment as well, as in here: `-- The WHERE line /-*"/`
Because `skipStringLiteral()` is the first in the loop of `removeComments()`, would this skip that entire enclosed part, and cause ignoring the comment(s)? Would this leave off half of the comments unfinished, which may then mix up comment content with actual statements ?

internal/expr/parser.go
```diff
@@ -109,6 +109,62 @@ func (p *Parser) add(part queryPart) {
...
+			continue
+		}
+
+		if p.skipByte('-') || p.skipByte('/') {
+			startPos := p.pos - 1
+			c := p.input[startPos]
+			var end byte
+			if (c == '-' && p.skipByte('-')) || (c == '/' && p.skipByte('*')) {
+				if c == '-' {
+					end = '\n'
```
solmyrcan: If the line endings are marked by `\r\n`, we'll need to keep and not skip both characters, and write them in the `res` buffer. Of course this is needed if such systems are to be supported :) 
Would we need to support those with `\r` only as well?

internal/expr/parser.go
```diff
@@ -109,6 +109,62 @@ func (p *Parser) add(part queryPart) {
...
+		textEnd = p.pos
+		// Ignore comment chars in string literals.
+		if ok, err := p.skipStringLiteral(); err != nil {
+			return err
+		} else if ok {
+			continue
+		}
+
+		if p.skipByte('-') || p.skipByte('/') {
+			startPos := p.pos - 1
```
solmyrcan: `startPos` doesn't seem to be used anywhere else besides the line below. Might instead do `c := p.input[p.pos-1]`

internal/expr/parser.go
```diff
@@ -109,6 +109,62 @@ func (p *Parser) add(part queryPart) {
...
+							p.pos++
+							if !p.skipByte('/') {
+								continue
+							}
+						}
+						break
+					}
+					p.pos++
+				}
+
```
solmyrcan: If the end character for the comment part is never found, this may be an issue for parser later on, as the whole content starting from the start of the comment which never ends until the end of input is skipped, perhaps removing parts of the statement along with it. Maybe this should be treated as an error, and we may need to only do `p.init()` below after clean parsing of the comments found? I think this issue applies to multi-line comments. Once this is handled, a test case could be added to check that such user error doesn't pass through undetected.

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,65 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+	}{{
+		input:    `-- Single line comment`,
+		expected: ``,
+	}, {
+		input: `-- Simple with line break
+		`,
+		expected: `
+		`,
+	}, {
+		input: `/* muti
```
solmyrcan: multi

internal/expr/parser_helper_test.go
```diff
@@ -132,3 +132,65 @@ func (s *ExprInternalSuite) TestUnfinishedQuote(c *C) {
...
+
+func (s *ExprInternalSuite) TestRemoveComments(c *C) {
+	testList := []struct {
+		input    string
+		expected string
+	}{{
+		input:    `-- Single line comment`,
+		expected: ``,
+	}, {
+		input: `-- Simple with line break
```
solmyrcan: Single line comment with line break

internal/expr/expr_test.go
Aflynn50: If there are quotes inside the comments they should be ignored, as per the spec. I've changed this test to include quotes.

internal/expr/parser.go
Aflynn50: In the sqlite spec it specifially says ascii character 0x01a so it should be ok with just `\n`

internal/expr/parser.go
Aflynn50: True, changed.

internal/expr/parser.go
Aflynn50: In the sqlite spec it defines the comment as ending at the end char or the end of input. So if there is no end character then the comment spans to the end of input.

internal/expr/parser_helper_test.go
Aflynn50: Changed.

internal/expr/parser_helper_test.go
Aflynn50: Changed.

internal/expr/parser.go
```diff
@@ -109,6 +109,61 @@ func (p *Parser) add(part queryPart) {
...
+	for p.pos < len(p.input) {
+		textEnd = p.pos
+		// Ignore comment chars in string literals.
+		if ok, err := p.skipStringLiteral(); err != nil {
+			return err
+		} else if ok {
+			continue
+		}
+
+		if p.skipByte('-') || p.skipByte('/') {
```
TheSignPainter98: Would it be cleaner to separate out the `--` and `/*` cases?

internal/expr/parser.go
```diff
@@ -109,6 +109,61 @@ func (p *Parser) add(part queryPart) {
...
+							if !p.skipByte('/') {
+								continue
+							}
+						}
+						break
+					}
+					p.pos++
+				}
+
+				res.WriteString(p.input[textStart:textEnd])
```
TheSignPainter98: There might be a bug around here, `textEnd` isn't updated after the start of the loop, so I think this line only ever writes `p.input[p.pos:p.pos]` to the buffer?

Perhaps the `textEnd` variable should be replaced with just `p.pos` here?

internal/expr/parser.go
```diff
@@ -109,6 +109,61 @@ func (p *Parser) add(part queryPart) {
 	p.partStart = p.pos
 }
 
+// removeComments passes over the input removing comments.
+// It does not allow for nested comments.
+func (p *Parser) removeComments() error {
+	var res bytes.Buffer
+	var textStart = p.pos
+	var textEnd = p.pos
```
TheSignPainter98: As `textEnd` is either immediately assigned or never used, it could be cleaner to declare it here as
```go
var textEnd int
```

internal/expr/parser.go
```diff
@@ -119,6 +174,11 @@ func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
 	}()
 
 	p.init(input)
+
+	if err := p.removeComments(); err != nil {
```
TheSignPainter98: should this call be placed inside `p.init`? This would avoid the second call to `p.init` made inside `removeCommants` (so line 163 could be replaced with just `p.input = res.String()`)

internal/expr/parser.go
Aflynn50: Text here refers to the non-commented text. So textEnd is where the comment starts and text start is the where it starts before the previous comment. i.e.: for the text `"noncommented text -- comment\n"` the position of the vars when we do WriteString would be: `"`textStart`noncommented text` textEnd `-- comment\n`p.pos`"`

Aflynn50: If you can think of any better names that would be great though haha

internal/expr/parser.go
Aflynn50: I just tried it, imo it looks better as one big block, theres a lot of similar code in the two blocks if they're separated.

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: The problem with that is more the error handling, I think its good that p.init doesn't return an error, it doens't feel like it should, and we should probably be aware we are removing the comments.

niemeyer (changes requested):
This is unusual for the project, thus far, in the sense that it's changing the input buffer rather than parsing it through and skipping. It's not clear why it was done this way, since the logic required to know what to remove correctly is pretty much the same logic required to know what to skip correctly, except the former is more work as it's manipulating buffers instead of just walking through.  Unless there's already some good rationale for doing it this way, I'd prefer to attempt an implementatino that does the same as everything else. A good analogy here is quoted strings: same idea, a start marker, and end marker, and stuff in the middle that must be skipped correctly.

Aflynn50:
> This is unusual for the project, thus far, in the sense that it's changing the input buffer rather than parsing it through and skipping. It's not clear why it was done this way, since the logic required to know what to remove correctly is pretty much the same logic required to know what to skip correctly, except the former is more work as it's manipulating buffers instead of just walking through. Unless there's already some good rationale for doing it this way, I'd prefer to attempt an implementatino that does the same as everything else. A good analogy here is quoted strings: same idea, a start marker, and end marker, and stuff in the middle that must be skipped correctly.

I've found the reason I for this that I couldn't remember in our meeting!
The issue is when a comment appears in the middle of something that we would parse as an output expression, e.g.:
```
SELECT
 col1, -- The first column
 col2, -- The second column
 col3, /* 
 possiblecol4 */
AS
 &Person.*
FROM
 table
```
In this case we have to worry about comments every time we have any gap between text. One alternative to this approach could be to worry about this in `skipBlanks`, and also parse them in a similar manner to string literals outside of the expressions. Would you prefer I try that approach?

niemeyer:
As mentioned above, how would that be different from parsing a quoted string?

Aflynn50:
> As mentioned above, how would that be different from parsing a quoted string?

I've changed it to skip the comments in place rather than a separate initial pass.

The main difference is that are not looking for quoted strings within input or output expressions. We now have two different entry points to `removeComment`, `advance` and `skipBlanks`.

Let me know what you think of this new way of doing it.

niemeyer (approved):
Thanks, good work there. As we discussed, I wonder if doing the comment skipping looking for the composed strings ("/*", etc) wouldn't result in a simpler function, but this looks correct and it's great to have it.

PR #38, opened 2023-02-09
Add more advance chars

Aflynn50:
The advance function in the parser looks for places in the SQL query that might be the start of an input or output expression. This PR adds more chars that will be considered as a possible start for an IO expression.


niemeyer (approved):

niemeyer (approved):
As we discussed back then, these are the nice problems to have. :)

PR #39, opened 2023-02-14
Adjust saved reflection information 

Aflynn50:
This PR removes the usued `fieldToTag` map from the `info` struct and adds in an ordered list of tags: `tags`. This change is made to reflect how the `info` struct was used in the `Prepare` stage and makes the code slightly more efficient.

It fixed the bug where a panic would occur if an unexported field has a db tag. It now throws an error.

It stops`typeInfo` from unnecessarley reflecting the value when only the type is used.

It changes the names `structType` and `fieldType` to `typ` as this is the more idiomatic way of doing things in Go (the struct they are fields of tells us what they are the type of).

### DIFF
--- a/internal/expr/info.go
+++ b/internal/expr/info.go
@@ -6,7 +6,7 @@ import (
 
 // Field represents a single field from a struct type.
 type field struct {
-	fieldType reflect.Type
+	typ reflect.Type
 
 	// Name is the name of the struct field.
 	name string
@@ -21,11 +21,11 @@ type field struct {
 
 // Info represents reflected information about a struct type.
 type info struct {
-	structType reflect.Type
+	typ reflect.Type
+
+	// Ordered list of tags
+	tags []string
 
 	// Relate tag names to fields.
 	tagToField map[string]field
-
-	// Relate field names to tags.
-	fieldToTag map[string]string
 }
--- a/internal/expr/info.go
+++ b/internal/expr/info.go
@@ -4,11 +4,10 @@ import (
 	"reflect"
 )
 
-// Field represents a single field from a struct type.
+// field represents reflection information about a field from some struct type.
 type field struct {
 	typ reflect.Type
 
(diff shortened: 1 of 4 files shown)


internal/expr/prepare.go
```diff
@@ -104,11 +104,11 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []outputDest,
 		if t.name != "*" {
 			f, ok := info.tagToField[t.name]
 			if !ok {
-				return nil, nil, fmt.Errorf(`type %s has no %q db tag`, info.structType.Name(), t.name)
+				return nil, nil, fmt.Errorf(`type %s has no %q db tag`, info.typ.Name(), t.name)
 			}
 			// For a none star expression we record output destinations here.
```
solmyrcan, line 109: For a non-star expression, we record output destinations here.

internal/expr/info.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // Field represents a single field from a struct type.
```
solmyrcan: Field represents reflected information about a single field from a struct type.

internal/expr/info.go
```diff
@@ -6,7 +6,7 @@ import (
...
 type field struct {
-	fieldType reflect.Type
+	typ reflect.Type
 
 	// Name is the name of the struct field.
```
solmyrcan: Name of the struct field.

Aflynn50: Changed.

internal/expr/info.go
Aflynn50: Changed.j

internal/expr/prepare.go
Aflynn50: Ah I should probably change that somewhere else, I think it doesn't come under the scope of this PR.

niemeyer (approved):
Looks reasonable. :+1:

PR #40, opened 2023-02-14
Raw user input in error's

Aflynn50:
This PR adds an extra field to `inputPart` and `outputPart` called `rawString`. This contains the section of the input that was parsed by the `parseInput` or `parseOutput`. This is then used in error messages in the prepare phase.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -450,12 +450,17 @@ func (p *Parser) parseTargets() ([]fullName, bool, error) {
 // parseOutputExpression requires that the ampersand before the identifiers must
 // be followed by a name byte.
 func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
+	start := p.pos
 
 	// Case 1: There are no columns e.g. "&Person.*".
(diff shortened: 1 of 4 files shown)


internal/expr/parser.go
```diff
@@ -451,11 +451,16 @@ func (p *Parser) parseTargets() ([]fullName, bool, error) {
 // be followed by a name byte.
 func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
 
+	start := p.pos
```
TheSignPainter98, line 453: I think the newline would look nicer just after this line than just before :)

internal/expr/parts.go
```diff
@@ -32,7 +32,8 @@ func (fn fullName) String() string {
 // inputPart represents a named parameter that will be sent to the database
 // while performing the query.
 type inputPart struct {
-	source fullName
+	source    fullName
+	rawString string
```
TheSignPainter98: The 'String' suffix on a string-type field seems redundant, perhaps this could just be `raw`?

PR #42, opened 2023-03-02
Add Scan

Aflynn50:
This PR adds the `scan.go` file. This contains all the functions to take the values returned by the database and insert them into the structs mentioned in the output expressions.

It also fixes a bug the named input parameters. These previously had been implemented but with `?` in the query rather than `@name`. This bug was not found in testing because it was not possible to test with a complete database at that point.


internal/expr/prepare.go
```diff
@@ -183,7 +183,10 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 	}
 
 	var sql bytes.Buffer
+
+	// n counts inputs and m counts outputs.
 	var n int
+	var m int
```
TheSignPainter98: I think that `n` and `m` needing a comment to explain what they are implies that their names should be more descriptive, perhaps. How about calling them `inputIdx` and `outputIdx`?

Aflynn50: Fixed!

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+func (re *ResultExpr) Decode(dests ...any) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot decode expression: %s", err)
+		}
+	}()
+
+	for _, dest := range dests {
+		if dest == nil {
+			return fmt.Errorf("need valid sturct, got nil")
```
solmyrcan: sturct -> struct

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	return nil
+}
+
+func (re *ResultExpr) scan() ([]any, error) {
+	cols, err := re.rows.Columns()
+	if err != nil {
+		return []any{}, fmt.Errorf("cannot scan: %s", err)
+	}
+
+	// rows.Scan takes a list of pointers to where it should scan the values
```
solmyrcan: I thought of putting it this way, but please ignore if it doesn't improve the message :) 
_rows.Scan takes a list of pointers, each pointing to a value where a scanned result from the result row will be written into._

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	// from the result row.
+	// We call rows.Scan(ptrs) and the results are stored in vals.
```
solmyrcan: Another possible way to put it:
_Calling rows.Scan(ptrs) stores the results in vals._

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	vals := make([]any, len(cols))
+	ptrs := make([]any, len(cols))
+	for i, _ := range cols {
+		ptrs[i] = &vals[i]
+	}
+	re.rows.Scan(ptrs...)
+
+	// offset is the distace between column X in the row and the position of the
```
solmyrcan: I thought maybe we could rephrase this in a different way, but again, ignore if you think it's not clearer :)

sqlair columns are named as _sqlair_X, where X = 0, 1, 2, .... 
offset is the difference between X and the list index where _sqlair_X is located in the list of columns of the result row.

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+		} else {
+			offset++
+		}
+	}
+
+	return sqlairVals, nil
+}
+
+// decodeValue sets the fields in the reflected struct "v" which have tags
+// corrosponding to columns in current row of the query results.
```
solmyrcan: corresponding to columns in the current row 
:)

internal/expr/complete.go
```diff
@@ -41,6 +41,15 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 		}
 		v := reflect.ValueOf(arg)
 		t := v.Type()
+
+		if t.Kind() == reflect.Pointer {
+			return nil, fmt.Errorf("need struct, got pointer to %s", t.Elem().Kind())
```
TheSignPainter98: I think it's simpler just to handle this case in the check below? It would also make the error messaging slightly more consistent. Currently, the messages are different if this function gets a `*int` compared to an `int` for example, which feels odd

internal/expr/db_test.go
```diff
@@ -0,0 +1,622 @@
+package expr_test
+
+import (
+	"database/sql"
+
+	"github.com/canonical/sqlair/internal/expr"
+	_ "github.com/mattn/go-sqlite3"
```
TheSignPainter98, line 7: I'm curious, is this necessary?

internal/expr/db_test.go
```diff
@@ -0,0 +1,622 @@
...
+	 `
+
+	inserts := []string{
+		"INSERT INTO person VALUES ('Fred', 30, 1000, 'fred@email.com');",
+		"INSERT INTO person VALUES ('Mark', 20, 1500, 'mark@email.com');",
+		"INSERT INTO person VALUES ('Mary', NULL, 3500, 'mary@email.com');",
+		"INSERT INTO person VALUES ('James', 35, 4500, 'james@email.com');",
+		"INSERT INTO address VALUES (1000, 'Happy Land', 'Main Street');",
+		"INSERT INTO address VALUES (1500, 'Sad World', 'Church Road');",
+		"INSERT INTO address VALUES (3500, 'Ambivilent Commons', 'Station Lane');",
```
TheSignPainter98:
```suggestion
		"INSERT INTO address VALUES (3500, 'Ambivalent Commons', 'Station Lane');",
```
:)

internal/expr/db_test.go
```diff
@@ -0,0 +1,622 @@
...
+		resultExpr.Close()
+	}
+
+	_, err = db.Exec(drop)
+	if err != nil {
+		c.Fatal(err)
+	}
+}
+
+func (s *DBSuite) TestAll(c *C) {
```
TheSignPainter98: Somehow the name `TestAll` feels misleading at first glance, as to me it unfortunately implied that it somehow controlled the other `Test*` functions. Perhaps `TestAll` and `TestOne` could be renamed to something like `TestResultAll` and `TestResultOne`?

Also is it worth adding a `TestAllError` function?

internal/expr/db_test.go
```diff
@@ -0,0 +1,622 @@
...
+
+	db, err := createExampleDB(create, inserts)
+	if err != nil {
+		return "", nil, err
+	}
+	return drop, db, nil
+
+}
+
+func (s *DBSuite) TestJujuStore(c *C) {
```
TheSignPainter98: This seems like a nice real-world example, but I'm not sure what it tests which isn't covered by the cases above?

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+		} else if !ok {
+			break
+		}
+
+		rs := []any{}
+		for _, t := range ts {
+			rp := reflect.New(t)
+			// We need to unwrap the struct inside the interface{}.
+			r := rp.Elem()
+			err := re.decodeValue(r)
```
TheSignPainter98: Could this be a call to `re.Decode(...)`? It would be very nice if both `All` and `Once` called the same function to decode the results

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+			ts = append(ts, t)
+		}
+	}
+	return ts
+}
+
+// All returns a slice containing all rows returned in the query. Each row is
+// a slice of any types that contains all structs mentioned in the output expressions
+// of the query.
+func (re *ResultExpr) All() ([][]any, error) {
```
TheSignPainter98, line 50: I think this contains some details which are present in the signature and which don't necessarily help the user figure out why they would need this function. How about something shorter?
```
// All returns every row resulting from a query, parsed using that query's output expressions.
```
I'm not 100% sure about the use of 'parsed' in this suggestion though as it already has a meaning elsewhere in SQLAir...

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+
+		s = append(s, rs)
+	}
+
+	re.Close()
+	return s, nil
+}
+
+// Next calls sql.rows.Next() and checks if it has thrown an error.
+func (re *ResultExpr) Next() (bool, error) {
```
TheSignPainter98, line 87: The `sql.rows.Next()` feels like an implementation detail.

Honestly I'm not sure whether this function even needs a doc, the signature doesn't leave too much room for interpretation. Although perhaps it could benefit from the return types being named `(ok bool, err error)`?

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	ok := re.rows.Next()
+	if !ok && re.rows.Err() != nil {
+		return false, re.rows.Err()
+	}
+	return ok, nil
+}
+
+// Decode copies the columns in the current row into the fields of the structs in dests.
+// dests can contain any subset of the structs mentioned in the query.
+func (re *ResultExpr) Decode(dests ...any) (err error) {
```
TheSignPainter98, line 97: The term 'copies' is quite specific, it feels like the something the user of this function must think carefully about. Perhaps 'puts' is more suitably vague.

Also on the second like, it should be 'dests can _be_ any subset,' otherwise, considering the empty set, `dests` could be _any_ set!

internal/expr/scan.go
TheSignPainter98: Also `s/distace/distance/` :)

internal/expr/scan.go
TheSignPainter98: I don't think it's customary to put quote marks around parameter names. (Personally I think this doesn't look great but it is at least convention.)

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+
+		}
+	}
+	if !typeFound {
+		return fmt.Errorf("no output expression of type %s", v.Type().Name())
+	}
+	return nil
+}
+
+func setValue(dest reflect.Value, fInfo field, val any) error {
```
TheSignPainter98: Perhaps `fInfo` could be renamed to something like `dest`? Ideally I'd suggest calling it `field`, but I'm not sure shadowing a type name is a good idea

internal/expr/scan.go
Aflynn50: So I actually ended up rewriting quite a lot after discovering the `Addr` function is the reflect package that allows us to get the address of a field of a struct and set it directly. This means we don't even need this scan function at all since we just pass `rows.Scan` a list of the `Addr()` of each field and it can set them directly itself. This has some benefits including that it will automatically convert from an `int64` to an `int` if needed (before it was decoding into something of type `any` but now it know the fields type since the `Addr()` gives us a pointer with a type so the scan function can do the conversions it was always meant to do!)

internal/expr/scan.go
Aflynn50: See above

internal/expr/scan.go
Aflynn50: Changed

internal/expr/scan.go
Aflynn50: This function has changed as well. I've written a new comment for it (chock full of spelling mistakes probably)

internal/expr/complete.go
Aflynn50: Changed (in a different PR that is now in!)

internal/expr/db_test.go
Aflynn50: Its good to test it with a real db, otherwise we'll probably write something useless.

Aflynn50: I think this starts it in the background or something, not entirely sure but it means the `sql.Open("sqlite3", ":memory:")` works.

internal/expr/db_test.go
Aflynn50: Good catch!

internal/expr/db_test.go
Aflynn50: I'll change it to TestValidAll, we use that pattern elsewhere

Aflynn50: Theres no `TestAllError` because all the errors that All throws are passed up from functions tested in `TestDecodeError`

internal/expr/db_test.go
Aflynn50: I think its good to keep in there just as a pr piece if anyone from Juju comes nosing around.

internal/expr/scan.go
Aflynn50: It's slightly more efficient this way because we dont do all the checks of Decode, also `reflect.New` returns a `reflect.Value`. We would also have to get the `Interface` and make it a first class value only for it to be reflected on again in Decode which seems like a lot of wasted expensive reflection.

internal/expr/scan.go
Aflynn50: I've got rid of the second sentence.

internal/expr/scan.go
Aflynn50: Changed

internal/expr/scan.go
Aflynn50: This comment has changed with the update as well.

internal/expr/scan.go
Aflynn50: This has all dissapered as well :)

internal/expr/scan.go
Aflynn50: Changed

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	"database/sql"
+	"fmt"
+	"reflect"
+	"strconv"
+)
+
+// ResultExpr represents the result of a database query.
+type ResultExpr struct {
+	outputs []field
+	rows    *sql.Rows
```
niemeyer, line 13: Per our last call, the API for this type is what the spec defines as `Query`. This is breaking away the notion that this package is about expresssions (`expr`), and introducing concepts which are about handling the database conenction and interactions.

Is this what we want?  What alternatives were considered?

internal/expr/scan.go
niemeyer: Per the spec and our last call where covered this, this shoud be `Next() bool`, without the error result which is cached for the `Close` time.

internal/expr/scan.go
niemeyer: _copies_ is still there.  Per comment from Ed, we can probably generalize it a bit:
```
// Decode decodes the current result into the dests struct values.
```
No need to define further constraints here, which should have been explained on the way in where we document the statement. Worst case, the developer will get a nice error message pointing out the actual mistake.

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	return ok, nil
+}
+
+// Decode copies the columns in the current row into the fields of the structs in dests.
+// dests must contain all the structs mentioned in the query.
+func (re *ResultExpr) Decode(dests ...any) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot decode expression: %s", err)
```
niemeyer, line 100: s/expression/result/ - It's not the expression that is being decoded here.

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		destVal := reflect.ValueOf(dest)
+		if destVal.Kind() != reflect.Pointer {
+			return fmt.Errorf("need pointer to struct, got non-pointer of kind %s", destVal.Kind())
+		}
+
+		destVal = reflect.Indirect(destVal)
+
+		if destVal.Kind() != reflect.Struct {
```
niemeyer, line 118: Minor nitpick: there seems to be a bit of excessive spacing all around, and I'm picking this point as a good example: all the lines above are tightly related to one another, and yet they are almost all split apart.

internal/expr/scan.go
```diff
@@ -0,0 +1,196 @@
...
+
+	ptrs := []any{}
+
+	// 	sqlair columns are named as _sqlair_X, where X = 0, 1, 2,...
+	//  offset is the difference between X and the index i and where column _sqlair_X is located in cols.
+	//  It allows non-sqlair columns to be returned in the results.
+	offset := 0
+
+	for i, col := range cols {
+		if col == "_sqlair_"+strconv.Itoa(i-offset) {
```
niemeyer, line 167: This seems more complex/expensive than expected. We were the ones putting the index in, which should match the actual index that we have internally. Why are we doing such manual string building comparisons instead of actually using the index that we've put in ourselves directly?

niemeyer (changes requested):
Thanks Alastair. Looks quite nice. Just some organizational details mostly.

internal/expr/scan.go
Aflynn50: True, I've implemented it as such in the `sqlair.go` file of #44 but theres no reason it shouldn't match here.

internal/expr/scan.go
Aflynn50: That's a good point, I think this scanning side of it would be better suited in a separate package.

internal/expr/scan.go
Aflynn50: Changed

internal/expr/scan.go
Aflynn50: Changed

internal/expr/scan.go
Aflynn50: Changed to make it more concise

internal/expr/scan.go
Aflynn50: We only know about the columns that we parsed as part of an SQLair expression. We don't know if the user might have put other columns into the query that could displace the ordering, e.g. `SELECT col1, &Person.*, col2 ...`. Of course this simple case would probably not be something the user does since the `col1` and `col2` can't be read with SQLair (until we add the `M` type), but with more complex queries we might have issues.

PR #43, opened 2023-03-02
Fix missing named input parameters and add complete error checks

Aflynn50:
This PR fixes a few issues that were found with complete when writing PR #42.

The first issue is that we were still using `?` in the query but in complete were generating named parameters of the form `@sqlair_0`. We use named parameters to avoid issues with users add their own `?` to the query and displacing the parameters passed to the DB.

The second issue is that the error messages thrown when a user passes a pointer to a struct instead of a copy the error message was confusing. I found this issue when I made that mistake. Similarly if they passed a none struct type. Both of these problems have now been fixed with explicit checks of the `Kind` that throw the appropriate errors.

The error tests for Complete in `expr_test.go` have also been refactored into a table driven format.

### DIFF
--- a/internal/expr/complete.go
+++ b/internal/expr/complete.go
@@ -41,6 +41,15 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 		}
 		v := reflect.ValueOf(arg)
 		t := v.Type()
+
+		if t.Kind() == reflect.Pointer {
+			return nil, fmt.Errorf("need struct, got pointer to %s", t.Elem().Kind())
+		}
+
+		if t.Kind() != reflect.Struct {
+			return nil, fmt.Errorf("need struct, got %s", t.Kind())
+		}
+
 		typeValue[t] = v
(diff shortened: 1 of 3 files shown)


internal/expr/complete.go
```diff
@@ -41,6 +41,14 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 		}
 		v := reflect.ValueOf(arg)
 		t := v.Type()
+
+		if t.Kind() != reflect.Struct {
+			if t.Kind() != reflect.Pointer {
```
niemeyer: This won't work when the arg is a poitner to a struct.

We need something similar to `if v.Kind() == reflect.Pointer { v = v.Elem() }` beforehand, and then we don't need to worry about Pointers in the test here.

niemeyer (changes requested):
Nice, just one detail about pointer handling and it's good to go.

niemeyer (approved):
Thanks!

PR #44, opened 2023-03-08
Add Library Interface

Aflynn50:
This PR adds the `sqlair.go` file. This file contains all the public methods that will be exposed as the library API.

It also contains PR #42.

The public interface contains a `Statement`, a `DB` and a `Query` type. The statement represents a prepared SQLair statement. The `DB` is a database that can execute queries that include SQLair DML and the `Query` represents the query and iterates over the results.

Following the spec we use a pattern of iteration over the `Query` where errors are not immediately returned. Instead, `Next` and `Decode` return a boolean and if an error is encountered. The error can then be checked with `Query.Err()`.

`Close` must be called once iteration over the rows has finished. It is called implicitly if `Next` or `Decode` return false.


CONTRIBUTING.md
```diff
@@ -2,15 +2,15 @@ Code formatting
 ===============
 
 Go provides a tool, `go fmt`, which facilitates a standardised format to Go 
-source code. SQLAir has two additional policies:
+source code. SQLair has two additional policies:
```
TheSignPainter98: Perhaps these spelling changes belong in another PR? There are a couple elsewhere too

internal/expr/complete.go
```diff
@@ -8,6 +8,14 @@ import (
 	"strings"
 )
 
+func GetCompletedSQL(ce *CompletedExpr) string {
```
TheSignPainter98: This function and `GetCompletedArgs` feel like getters, so I think they could benefit from being methods? If so, they should probably be `ce.CompletedSql()` and `ce.CompletedArgs()` respectively

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	id integer,
+	address_id integer,
+	email text
+);
+CREATE TABLE address (
+	id integer,
+	district text,
+	street text
+);
+
```
TheSignPainter98: I think this extra whitespace can be removed

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return db, nil
+}
+
+func personAndAddressDB() (string, *sql.DB, error) {
+	create := `
```
TheSignPainter98: The name `create` on its own feels ambiguous, perhaps `create_tables` would be more precise?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	id integer,
+	district text,
+	street text
+);
+
+`
+	drop := `
+	 drop table person;
+	 drop table address;
+	 `
```
TheSignPainter98: A few minor things here:
1. The indentation is a little strange, it's probably best to keep it consistent with `create` (either indenting or dedenting both)
2. If this is to remain indented, there's an extra space after the leading tabs, so the query is mis-aligned, probably best to make this another tab to clearly indent one level further
3. The 'drop' and 'table' keywords aren't uppercase like the other queries
4. Somehow calling a string `drop` feels strange when it appears later, perhaps `drop_tables` to match `create_tables` would be clearer?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	db, err := createExampleDB(create, inserts)
+	if err != nil {
+		return "", nil, err
+	}
+	return drop, db, nil
+}
+
+func (s *DBSuite) TestValidDecode(c *C) {
+	var tests = []struct {
+		summery  string
```
TheSignPainter98: `s/summery/summary/g` :)

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query    string
+		types    []any
+		inputs   []any
+		outputs  [][]any
+		expected [][]any
+	}{{
+		summery:  "double select with name clash (first 4 results)",
```
TheSignPainter98: It's not immediately clear what the 'first 4 results' here refers to, and as there are no other summaries which share the same prefix, could this be removed?

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+		if t := out.structType; !isDup[t] {
+			isDup[t] = true
+			ts = append(ts, t)
+		}
+	}
+	return ts
+}
+
+// All iterates over the query and decodes all the rows.
+// It fabricates all struct instansiations needed.
```
TheSignPainter98: `s/instansiations/instantiations/` :)

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+			return [][]any{}, err
+		} else if !ok {
+			break
+		}
+
+		newStructs := make([]reflect.Value, len(ts))
+
+		for i, t := range ts {
+			rp := reflect.New(t)
+			// We need to unwrap the struct inside the interface{}.
```
TheSignPainter98: This comment feels quite long for such a simple section of the logic? I think it's worth highlighting why `.Elem()` is important, but currently the comment just describes what the code is already doing rather than why. Also, what it currently says isn't true, `t` is a `reflect.Type` rather than an `interface{}`. If in future a kind other than structs were supported, this comment would also quietly go out of date.

I think this loop would benefit from being simplified, perhaps something like this would be cleaner?
```go
for i, t := range ts {
    newStructs[i] = reflect.New(t).Elem() // t is a pointer
}
```

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	return ts
+}
+
+// All iterates over the query and decodes all the rows.
+// It fabricates all struct instansiations needed.
+func (re *ResultExpr) All() ([][]any, error) {
+	var rss [][]any
+
+	ts := getTypes(re.outputs)
```
TheSignPainter98: Perhaps this could be called `output_types`? I only see it used once

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := re.decodeRow(vals...); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// decodeRow copies a result row into the fields of the reflect.Values in dests.
+// All the structs mentioned in the query must be in dests.
+// The reflect.Value must be of Kind reflect.Struct and must be addressable and settable.
```
TheSignPainter98: As there are multiple `reflect.Value` values passed, I think this last line should be plural, to avoid misleading by making the type name plural, how about the following?
```
// All dests must be of kind reflect.Struct and must be addressable and settable.
```

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// One is shorthand for q.Next(); q.Decode(outputStructs...); q.Close().
+func (q *Query) One(outputStructs ...any) error {
+	return q.re.One(outputStructs...)
+}
+
+// All iterates over the query and decodes all the rows.
+// It fabricates all struct instansiations needed.
+func (q *Query) All() ([][]any, error) {
```
TheSignPainter98: I think more precisely:
```
// All collects the query results, decodes and returns them in fresh instances.
```
Otherwise, `s/instansiations/instantiations/` :)

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+		// We must close the rows if an error occours.
+		// The error, if any, from Rows.Close is ignored.
+		_ = q.re.Close()
+		return false
+	}
+	return true
+}
+
+// Close finishes iteration of the results.
+func (q *Query) Close(outputStructs ...any) error {
```
TheSignPainter98: Are the `outputStructs` parameters required here?

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return res, nil
+}
+
+// Next prepares the next row for decoding.
+// The first call to Next will execute the query.
+// If an error occours it can be checked with Query.Err(), Next will return false and Close will be called implicitly.
+// If an error has previously occoured Next will return false.
+func (q *Query) Next() bool {
```
TheSignPainter98, line 93: Would it be good to merge the `Next` and `Decode` functionality? It seems natural to me that `Next` would take addresses to fill with its outputs. Are there any situations where someone would want to know whether `Next` were successful but they don't care about decoding the results?

I think the last two lines can be simplified, how about something like this? (avoids `s/occours/occurs/`)
```
// On an error, returns false and closes the query.
```
I'm not 100% sure about implicitly closing the query, it seems a little magical, it might make it a little too easy to accidentally double-`Close` the query?

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	err := q.re.Close()
+	if err != nil {
+		return err
+	}
+	q.re = nil
+	return nil
+}
+
+// One is shorthand for q.Next(); q.Decode(outputStructs...); q.Close().
```
TheSignPainter98: The sequence of operations feels like an implementation detail and doesn't really highlight _why_ a user would want to call this method. Perhaps something like this would be more apparent?
```
// One runs a query, writes its sole result into the passed structs and closes the query.
```
Out of curiosity, does this method check that exactly one result is returned? The previous doc implies no, but I think it could be a useful feature

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+type Query struct {
+	pe  *expr.PreparedExpr
+	re  *expr.ResultExpr
+	err error
+	q   func() (*sql.Rows, error)
+}
+
+// Err returns the error, if any, that was encountered during iteration.
+// It should only be called after Close, or after Next has returned false (implicit close).
+func (q *Query) Err() error {
```
TheSignPainter98: The sentence-based line-wrapping here is inconsistent with other functions, it might be worth going through making sure everything here wraps at some _n_ characters

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+	return s
+}
+
+// Query takes a prepared SQLair Statement and returns a Query object for
+// iterating over the results.
+func (db *DB) Query(s *Statement, inputStructs ...any) (*Query, error) {
+	return db.QueryContext(s, context.Background(), inputStructs...)
+}
+
+func (db *DB) QueryContext(s *Statement, ctx context.Context, inputStructs ...any) (*Query, error) {
```
TheSignPainter98: It's probably worth noting why a user would want to pass a context to this function, perhaps something like
```
// QueryContext takes a prepared SQLAir statement, and returns a cancellable Query object for iterating over the results.
```
Also `QueryContext` feels like a getter, perhaps `QueryWithContext` would be clearer?

The above can also apply to `ExecContext` below.

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+	preparedExpr, err := parsedExpr.Prepare(typeInstantiations...)
+	if err != nil {
+		return nil, err
+	}
+	return &Statement{pe: preparedExpr}, nil
+}
+
+func MustPrepare(query string, typeInstantiations ...any) *Statement {
```
TheSignPainter98: I think this function could benefit from a doc. As it panics, perhaps it's worth mentioning it explicitly in addition to using the `Must` prefix? Maybe something like this could work?
```
// MustPrepare prepares a query with the given type instantiations. Panics on error.
```

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+
+// sqlDB returns the underlying database object.
+func (db *DB) sqlDB() *sql.DB {
+	return db.db
+}
+
+// Prepare checks that SQLair DML statements in a SQL query are well formed.
+// typeInstantiations must contain an instance of every type mentioned in the
+// SQLair DML of the query. These are used only for type information.
+func Prepare(query string, typeInstantiations ...any) (*Statement, error) {
```
TheSignPainter98: I'm not sure _checks_ is the right word here as it implies that the function returns a `bool`. I think instead, using some form of 'try' would fit better?
```
// Prepare tries to parse SQLair DML statements into SQL and extract type
// information from the passed instances of the types mentioned in query.
```
Also, is it possible to avoid 'DML'? I have a personal phobia of acronyms :P

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// Statement is a sql statement with valid sqlair DML.
+type Statement struct {
+	pe *expr.PreparedExpr
+}
+
+// Query stores a database query and iterates over the results.
+type Query struct {
```
TheSignPainter98: Really minor thing, I think query _holds_ a database query? 'Stores' when dealing with database things might have the wrong implication. Also, as this is a type, I don't think that it 'iterates,' perhaps it 'allows iteration over results'?

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// Statement is a sql statement with valid sqlair DML.
+type Statement struct {
```
TheSignPainter98, line 13: I think this comment doesn't tell the whole story, perhaps
```
// Statement represents a SQLair DML expression, parsed from a
// SQL query and with relevant type information.
```
Also, is it worth making the capitalisation of mentions of SQL consistent with SQLair?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	}, {
+		summery:  "select into star struct",
+		query:    "SELECT (name, address_id) AS &Person.* FROM person WHERE address_id IN ( $Manager.address_id, $Address.district )",
+		types:    []any{Person{}, Address{}, Manager{}},
+		inputs:   []any{Manager{PostalCode: 1000}, Address{ID: 2000}},
+		outputs:  [][]any{{&Person{}}},
+		expected: [][]any{{&Person{Fullname: "Fred", PostalCode: 1000}}},
+	}}
+
+	// A Person struct with different int types
```
solmyrcan: This line reads a bit ambiguous as it mentions "different int types". Perhaps we could combine the two sentences as:
"A Person struct that shadows the one in the tests above"

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+	var testsWithOtherPerson = []struct {
+		summery  string
+		query    string
+		types    []any
+		inputs   []any
+		outputs  [][]any
+		expected [][]any
+	}{{
+		summery:  "mistyped person",
```
solmyrcan: maybe the test summary string could be: "shadow person"

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+	// A Person struct with different int types
+	// Person shadows the Person struct in the tests above
+	type Person struct {
+		ID         int32  `db:"id"`
+		Fullname   string `db:"name"`
+		PostalCode int32  `db:"address_id"`
+	}
+
+	var testsWithOtherPerson = []struct {
```
solmyrcan: maybe we could call this as `testsWithShadowPerson`

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query:    "SELECT * AS &Person.* FROM person",
+		types:    []any{Person{}},
+		inputs:   []any{},
+		outputs:  [][]any{{&Person{}}, {&Person{}}, {&Person{}}, {&Person{}}},
+		expected: [][]any{{&Person{30, "Fred", 1000}}, {&Person{20, "Mark", 1500}}, {&Person{40, "Mary", 3500}}, {&Person{35, "James", 4500}}},
+	}}
+
+	tests = append(tests, testsWithOtherPerson...)
+
+	drop, db, err := personAndAddressDB()
```
solmyrcan: `drop` could be renamed as `dropTables` here, if we change the name of the `drop` variable elsewhere.

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			c.Errorf("test %q failed (Exec):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			continue
+		}
+
+		// Get as many rows as we have lists of structs to output into.
+		for i, outs := range t.outputs {
+			if ok, err := resultExpr.Next(); err != nil {
+				c.Fatalf("test %q failed (Next):\ninput: %s\nerr: %s", t.summery, t.query, err)
+			} else if !ok {
+				c.Fatalf("test %q failed (Next):\ninput: %s\nerr: no more rows in query", t.summery, t.query)
```
solmyrcan: Should this be rephrased as `no rows in query results`, rather than `no more rows in query`, dropping the word `more` ?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			}
+			err := resultExpr.Decode(outs...)
+			if err != nil {
+				c.Fatalf("test %q failed (Decode):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			}
+			c.Assert(outs, HasLen, len(t.expected[i]),
+				Commentf("test %q failed:\ninput: %s", t.summery, t.query))
```
solmyrcan: Would this `Assert` `Commentf` line need some text on the reason of the assertion failure, like `err: insufficient number of outputs in query results` or similar, or does the assertion already output a relevant message when getting the `HasLen` marker?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			for j, out := range outs {
+				c.Assert(out, DeepEquals, t.expected[i][j],
+					Commentf("test %q failed (Decode):\ninput: %s", t.summery, t.query))
```
solmyrcan: Similar to above, would this line need a text to indicate the error reason?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	}, {
+		summery: "not a struct",
+		query:   "SELECT * AS &Person.* FROM person",
+		types:   []any{Person{}},
+		inputs:  []any{},
+		outputs: [][]any{{&map[string]any{}}},
+		err:     "cannot decode expression: need struct, got map",
+	}}
+
+	drop, db, err := personAndAddressDB()
```
solmyrcan: Similar to above - perhaps rename `drop` to `dropTables`

CONTRIBUTING.md
Aflynn50: Fair enough, that's probably true.

internal/expr/complete.go
Aflynn50: Changed.

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+	_, err = db.Exec(drop)
+	if err != nil {
+		c.Fatal(err)
+	}
+}
+
+func (s *DBSuite) TestDecodeErrors(c *C) {
+	var tests = []struct {
+		summery string
```
solmyrcan: summary - need to modify all in the file :)

internal/expr/db_test.go
Aflynn50: Changed

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			c.Errorf("test %q failed (Exec):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			continue
+		}
+
+		// Get as many rows as we have lists of structs to output into.
+		for _, outs := range t.outputs {
+			if ok, err := resultExpr.Next(); err != nil {
+				c.Fatalf("test %q failed (Next) :\ninput: %s\nerr: %s", t.summery, t.query, err)
+			} else if !ok {
+				c.Fatalf("test %q failed (Next) :\ninput: %s\nerr: no more rows in query", t.summery, t.query)
```
solmyrcan: Similar to the comment, perhaps say `no rows in query result`

internal/expr/db_test.go
Aflynn50: Removed.

internal/expr/db_test.go
Aflynn50: Changed!

internal/expr/db_test.go
Aflynn50: Haha, 57 changes, nice.

internal/expr/db_test.go
Aflynn50: It only checks the first 4 rows of the output, I've changed to rows instead of results.

internal/expr/scan.go
Aflynn50: Changed.

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			c.Errorf("test %q failed (Complete):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			continue
+		}
+		resultExpr, err := sqlairDB.Query(completedExpr)
+		if err != nil {
+			c.Errorf("test %q failed (Exec):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			continue
+		}
+
+		// Get as many rows as we have lists of structs to output into.
```
solmyrcan: Should this be rephrased as:
"Fill in the output structs with results from the query." 
or something similar.

internal/expr/scan.go
Aflynn50: Changed to outputTypes

internal/expr/db_test.go
Aflynn50: (to camel case version)

internal/expr/db_test.go
Aflynn50: (also to camel case)

internal/expr/scan.go
Aflynn50: Good point, that's clearer and simpler

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+func (s *DBSuite) TestValidAll(c *C) {
+	var tests = []struct {
+		summery  string
+		query    string
+		types    []any
+		inputs   []any
+		expected [][]any
+	}{{
+		summery:  "double select with name clash",
```
solmyrcan: Maybe change the test summary as:
"select person and address id combinations"

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query:    "SELECT p.id AS &Person.*, a.id AS &Address.* FROM person AS p, address AS a",
+		types:    []any{Person{}, Address{}},
+		inputs:   []any{},
+		expected: [][]any{{Person{ID: 30}, Address{ID: 1000}}, {Person{ID: 30}, Address{ID: 1500}}, {Person{ID: 30}, Address{ID: 3500}}, {Person{ID: 20}, Address{ID: 1000}}, {Person{ID: 20}, Address{ID: 1500}}, {Person{ID: 20}, Address{ID: 3500}}, {Person{ID: 40}, Address{ID: 1000}}, {Person{ID: 40}, Address{ID: 1500}}, {Person{ID: 40}, Address{ID: 3500}}, {Person{ID: 35}, Address{ID: 1000}}, {Person{ID: 35}, Address{ID: 1500}}, {Person{ID: 35}, Address{ID: 3500}}},
+	}, {
+		summery:  "simple select person",
```
solmyrcan: Maybe summary as: "select all columns into struct" ?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query:    "SELECT * AS &Person.* FROM person",
+		types:    []any{Person{}},
+		inputs:   []any{},
+		expected: [][]any{{Person{30, "Fred", 1000}}, {Person{20, "Mark", 1500}}, {Person{40, "Mary", 3500}}, {Person{35, "James", 4500}}},
+	}, {
+		summery:  "single line returned",
```
solmyrcan: "single result with input conditions" ? or similar maybe

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		expected: [][]any{{Person{30, "Fred", 1000}, Address{1000, "Happy Land", "Main Street"}, Manager{30, "Fred", 1000}}},
+	}, {
+		summery:  "nothing returned",
+		query:    "SELECT &Person.* FROM person WHERE id = $Person.id",
+		types:    []any{Person{}},
+		inputs:   []any{Person{ID: 1243321}},
+		expected: [][]any{},
+	}}
+
+	drop, db, err := personAndAddressDB()
```
solmyrcan: Same as above, perhaps rename as `dropTables`

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+func (s *DBSuite) TestValidOne(c *C) {
+	var tests = []struct {
+		summery  string
+		query    string
+		types    []any
+		inputs   []any
+		outputs  []any
+		expected []any
+	}{{
+		summery:  "double select with name clash",
```
solmyrcan: Maybe change the test summary as:
"select person and address id combinations"

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query:    "SELECT p.id AS &Person.*, a.id AS &Address.* FROM person AS p, address AS a",
+		types:    []any{Person{}, Address{}},
+		inputs:   []any{},
+		outputs:  []any{&Person{}, &Address{}},
+		expected: []any{&Person{ID: 30}, &Address{ID: 1000}},
+	}, {
+		summery:  "select many",
```
solmyrcan: "select into multiple structs, with input conditions" ?

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+		query:    "SELECT p.* AS &Person.*, a.* AS &Address.*, p.* AS &Manager.* FROM person AS p, address AS a WHERE p.id = $Person.id AND a.id = $Address.id ",
+		types:    []any{Person{}, Address{}, Manager{}},
+		inputs:   []any{Address{ID: 1000}, Person{ID: 30}},
+		outputs:  []any{&Person{}, &Address{}, &Manager{}},
+		expected: []any{&Person{30, "Fred", 1000}, &Address{1000, "Happy Land", "Main Street"}, &Manager{30, "Fred", 1000}},
+	}}
+
+	drop, db, err := personAndAddressDB()
```
solmyrcan: `dropTables`

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+			c.Errorf("test %q failed (One):\ninput: %s\nerr: %s\n", t.summery, t.query, err)
+			continue
+		}
+		for i, e := range t.expected {
+			c.Assert(t.outputs[i], DeepEquals, e,
+				Commentf("test %q failed:\ninput: %s", t.summery, t.query))
+		}
+	}
+
+	_, err = db.Exec(drop)
```
solmyrcan: `dropTables`

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+	}{{
+		summery: "no rows",
+		query:   "SELECT * AS &Person.* FROM person WHERE id=12312",
+		types:   []any{Person{}},
+		inputs:  []any{},
+		outputs: []any{&Person{}},
+		err:     "cannot return one row: no results",
+	}}
+
+	drop, db, err := personAndAddressDB()
```
solmyrcan: `dropTables`

internal/expr/db_test.go
```diff
@@ -0,0 +1,507 @@
...
+func (s *DBSuite) TestOneErrors(c *C) {
+	var tests = []struct {
+		summery string
+		query   string
+		types   []any
+		inputs  []any
+		outputs []any
+		err     string
+	}{{
+		summery: "no rows",
```
solmyrcan: "no result row, with input condition" ?

internal/expr/scan.go
Aflynn50: Changed.

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	err := re.Decode(args...)
+	if err != nil {
+		return err
+	}
+	re.Close()
+	return nil
+}
+
+// getTypes returns the types in the order they appear in the query
```
solmyrcan: ...returns the field types...

internal/expr/scan.go
```diff
@@ -0,0 +1,197 @@
...
+func getTypes(outs []field) []reflect.Type {
```
solmyrcan: Do you think we could call the function as `getFieldTypes`, as it expects fields specifically?

sqlair.go
Aflynn50: Changed.

sqlair.go

sqlair.go
Aflynn50: It was like that before but in the go style guide it recommends keeping things on one line as long as they don't look too weird, even if they're longer than the standard _n_ characters.

sqlair.go
Aflynn50: I've replaces "DML" with "expression" and revised the comment

sqlair.go
Aflynn50: Added a comment.

package_test.go
```diff
@@ -0,0 +1,156 @@
...
+	lease_type_id text
+);
+CREATE TABLE lease_type (
+	id text,
+	type text
+);
+
+`
+	drop := `
+	 drop table lease;
```
solmyrcan: As in the above comment, perhaps we could align the tabs according to those of `create` above here as well.

package_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+	return drop, db, nil
+
+}
+
+func (s *PackageSuite) TestJujuStore(c *C) {
+	var tests = []struct {
+		summery  string
```
solmyrcan: summary :)

sqlair.go
Aflynn50: I've based these functions on these https://pkg.go.dev/database/sql#DB.Query . I've changed the comments to match the ones in the link a bit more.

sqlair.go
Aflynn50: The design of this is based on the spec. I find it a little strange but I think its what the library users will be used to.

sqlair.go
Aflynn50: No they are not.

sqlair.go
Aflynn50: Comment changed. I think the idea is that it doesn't matter if more than one result is returned, you just want any of them.

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+
+	return res, nil
+}
+
+// Next prepares the next row for decoding.
+// The first call to Next will execute the query.
+// If an error occours it can be checked with Query.Err(), Next will return false and Close will be called implicitly.
+// If an error has previously occoured Next will return false.
```
solmyrcan: occurred :)

sqlair.go

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		q.err = err
+		return false
+	}
+	return ok
+}
+
+// Decode stores a result row from the query into the structs specified in its SQLair DML.
+// outputStructs must contains all the structs mentioned in the query.
+// If an error occours during decode it will return false and Close will be called implicitly.
```
solmyrcan: Maybe this is a bit too much detail to mention the call to `Close()` on `re`, as it's an internal operation of `Decode()`, what do you think? I guess we would need to mention it if the caller would have to do something like `Open()` etc explicitly before calling `Decode()`.

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+// The error can be checked with Query.Err().
```
solmyrcan: How about: "If Decode returns false, an error check can be done with Query.Err()"

sqlair.go
```diff
@@ -0,0 +1,172 @@
...
+// If an error has previously occoured Decode will return false.
```
solmyrcan: occurred :)
Also maybe rephrase as this?:
If input query already has an error, Decode returns false.

internal/expr/db_test.go
TheSignPainter98: Ah good point

sqlair.go
TheSignPainter98: Ah I had a feeling this was the case, fair enough

sqlair.go
TheSignPainter98: Okay sounds good

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: The more is there because there might be a few rows present but we might be expecting 4 and only get 3. I've added the word results.

internal/expr/db_test.go
Aflynn50: With the HasLen marker its already pretty clear whats going on.

internal/expr/db_test.go
Aflynn50: I've added a int to indicate which row of the output this failed on.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed. I'm adding summerys faster then you can spot them haha.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Made clearer.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/db_test.go
Aflynn50: Changed.

internal/expr/scan.go
Aflynn50: Changed comment.

internal/expr/scan.go
Aflynn50: It doesn't get the type of the field though, it gets the type of the structs in the query.

package_test.go

package_test.go

sqlair.go

sqlair.go
Aflynn50: This bit is following the API from the spec.

sqlair.go

sqlair.go

internal/expr/scan.go
solmyrcan: aha! thanks, didn't read it well :)

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+// It is ready to be run on a SQLair DB.
+type Statement struct {
+	pe *expr.PreparedExpr
+}
+
+// Prepare expands the types mentioned in the SQLair expressions and checks
+// the SQLair parts of the query are well formed.
+// typeInstantiations must contain an instance of every type mentioned in the
+// SQLair expressions of the query. These are used only for type information.
+func Prepare(query string, typeInstantiations ...any) (*Statement, error) {
```
niemeyer: How about typeSample?

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+	db *sql.DB
+}
+
+func NewDB(db *sql.DB) *DB {
+	return &DB{db: db}
+}
+
+// sqlDB returns the underlying database object.
+func (db *DB) sqlDB() *sql.DB {
+	return db.db
```
niemeyer, line 53: I think this method was meant to be public, otherwise we just have two different ways to access it privately.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+	return true
+}
+
+// Close closes the query and returns any errors encountered during iteration.
+func (q *Query) Close() error {
+	if q.err != nil {
+		_ = q.rows.Close()
+		return q.err
+	}
+	err := q.rows.Close()
```
niemeyer, line 148: This can be simplified a bit:
```
err := q.rows.Close()
q.rows = nil
if q.err != nil {
        retrun q.err
}
return err
```

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+		return false
+	}
+	defer func() {
+		if !ok {
+			q.err = fmt.Errorf("cannot decode result: %s", q.err)
+		}
+	}()
+
+	if q.rows == nil {
+		q.err = fmt.Errorf("Decode called without calling Next")
```
niemeyer: `iteration ended or not started` seems more appropriate, as we might have reached the last element in which case this shoud fail as well.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		q.rows = rows
+	}
+	return q.rows.Next()
+}
+
+// Decode decodes the current result into the structs in dests.
+// dests must contain all the structs mentioned in the query.
+// If an error occurs it will be returned with Query.Close().
+func (q *Query) Decode(dests ...any) (ok bool) {
```
niemeyer: `outputStructs` from `One` feels nicer. In either case, we should use matching terms.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	q.Decode(outputStructs...)
+	return q.Close()
+}
+
+// All iterates over the query and decodes all the rows.
+// It fabricates all struct instantiations needed.
+// It returns a list of rows, each row is a list of pointers to structs.
+// The structs appear in the syntactic order they appear in the query.
+func (q *Query) All() ([][]any, error) {
```
niemeyer: As we discussed, we proably want to pass values into here similar to One and Decode, but let's please reduce a bit the size of the PR for now so we have less to figure out and can merge the basics before we figure the entire API out.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// sqlDB returns the underlying database object.
+func (db *DB) sqlDB() *sql.DB {
+	return db.db
+}
+
+// Exec executes an SQLair Statement without returning any rows.
+// Exec uses ExecContext with context.Background internally.
+func (db *DB) Exec(s *Statement, inputStructs ...any) (sql.Result, error) {
```
niemeyer: As we talked, can we please leave the Exec functions out for now?  We may end up doing them in terms of the Query API itself too, so let's focus on makign that one really nice and then we figure how to do the one shot calls.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+	return res, nil
+}
+
+// Query takes a prepared SQLair Statement and returns a Query object for
+// iterating over the results.
+// Query uses QueryContext with context.Background internally.
+func (db *DB) Query(s *Statement, inputStructs ...any) (*Query, error) {
+	return db.QueryContext(s, context.Background(), inputStructs...)
```
niemeyer: Again per the conversation, let's avoid returning an error here so that this kind of logic is possible:
```
err := db.Query(...).One(&result)
```
That tends to be nice to use, and it's going to be easy as we alredy have logic around q.err.

sqlair.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// QueryContext takes a prepared SQLair Statement and returns a Query object for
+// iterating over the results.
+func (db *DB) QueryContext(s *Statement, ctx context.Context, inputStructs ...any) (*Query, error) {
```
niemeyer: Same here.

niemeyer (changes requested):
Looks nice!

Here is a first pass per our conversation, only on the visible API side.

sqlair.go
Aflynn50: Changed.

sqlair.go

sqlair.go

sqlair.go
Aflynn50: Changed.

sqlair.go

sqlair.go
Aflynn50: Changed to `outputs` since when we introduce the `M` type it will no longer be `outputStructs` and `outputValues` becomes confusing once you have the `reflect.Value` of it in the mix.

sqlair.go
Aflynn50: Removed the All function.

sqlair.go
Aflynn50: Removed the Exec function.

sqlair.go

internal/expr/info.go
```diff
@@ -4,9 +4,13 @@ import (
...
 )
 
+type Outputs struct {
+	structFields []field
+}
+
 // field represents reflection information about a field from some struct type.
 type field struct {
-	fieldType reflect.Type
+	typ reflect.Type
```
niemeyer: This seems to be going back and forth in the last few PRs.  fieldType seems like a nice partner to structType below. Lacking the prefix, it's unclear which one it is.

internal/expr/prepare.go
```diff
@@ -15,6 +15,10 @@ type PreparedExpr struct {
 	sql     string
 }
 
+func (pe *PreparedExpr) Outputs() Outputs {
+	return Outputs{structFields: pe.outputs}
```
niemeyer: Names seem a bit mixed up and hard to follow.  We have `[]field` becoming `pe.outputs` becoming `structFields` becoming `Outputs`, and then we use them as `outputs` in the `Query`, except we then get an `outputs` parameter again which is not this.  We're not really helping code readers to figure what is meant where.  I'll make a more concrete proposal below.

internal/expr/decode.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+)
+
+// OutputAddrs returns list of pointers to struct fields specified in outputs.
+// All the structs mentioned in the query must be in dests.
+// All dests must be of kind reflect.Struct and must be addressable and settable.
+func OutputAddrs(cols []string, outputs Outputs, dests []reflect.Value) ([]any, error) {
```
niemeyer: The confusion is visible here again: `dests` is actually `outputs` in the code calling this function, but we have also `outputs` that comes from the other side. The names need to be more clear and be able to stick with the data that it represents throughout the code base.

sqlair.go
```diff
@@ -0,0 +1,176 @@
...
+// If an error occurs it will be returned with Query.Close().
+func (db *DB) QueryContext(s *Statement, ctx context.Context, inputStructs ...any) *Query {
+	q := &Query{outputs: s.pe.Outputs()}
+
+	ce, err := s.pe.Complete(inputStructs...)
+	if err != nil {
+		q.err = err
+	}
+	q.q = func() (*sql.Rows, error) {
+		return db.db.QueryContext(ctx, ce.CompletedSQL(), ce.CompletedArgs()...)
```
niemeyer: Okay, here is a general renaming suggestion:

- PreparedExpr.Complete(...) => PreparedExpr.Query(...) resulting QueryExpr
- CompletedExpr => QueryExpr
- CompletedExpr.CompletedSQL => QueryExpr.QuerySQL
- CompletedExpr.CompletedArgs => QueryExpr.QueryArgs
- Outputs.structFields => QueryExpr.outputs (aligns with other "outputs")
- Outputs => Dropped
- PreparedExpr.Outputs => Dropped, QueryExpr holds the data.
- inputStructs => inputArgs
- outputs (in Decode) => outputArgs
- outputs (in One) => outputArgs
- OutputAddrs(cols, outputs, dest) => QueryExpr.ScanArgs(columns, outputArgs) resulting scanArgs (outputArgs here is the parameter itself, reflect.Value processing goes inside ScanArgs)

The end result of all of this is that we have QueryExpr.{QuerySQL,QueryArgs,ScanArgs}, inputArgs, and outputArgs. The QueryExpr is generated at Query time, and carried over in the upcoming Iter type for decoding.

In the future, I think we'll need to chage outputs and inputs from being related to []field, as we add support for maps, etc, but let's leave that for another day so we can polish that public layer and the interaction with the expr package until it's all reading cleanly.

I of course may be missing details here, but hopefully the general proposal is somewhat consistent and cleans up the naming a bit.

internal/expr/decode.go
```diff
@@ -0,0 +1,52 @@
...
+		if col == "_sqlair_"+strconv.Itoa(i-offset) {
+			field := outputs.structFields[i-offset]
+			dest, ok := typeDest[field.structType]
+			if !ok {
+				return []any{}, fmt.Errorf("type %s found in query but not passed to decode", field.structType.Name())
+			}
+
+			val := dest.FieldByIndex(field.index)
+			if !val.CanSet() {
+				return []any{}, fmt.Errorf("cannot set field %s of struct %s", field.name, field.structType.Name())
```
niemeyer: These two error cases above all seems to be missing tests exploring the issues. It's nice to make sure we continue to keep these problems in mind as we refactor this logic in the future.

Also, the first parameter in those cases shoud be nil.

internal/expr/decode.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		typeDest[dest.Type()] = dest
+	}
+
+	//  SQLair columns in the results are named as _sqlair_X, where X = 0, 1, 2,...
+	//  offset is the difference between X and the index i of cols where the column _sqlair_X is located.
+	var offset = 0
+	var ptrs = []any{}
+	for i, col := range cols {
+		if col == "_sqlair_"+strconv.Itoa(i-offset) {
```
niemeyer: This logic is allocating a new string for every iteration of the loop, even when that column is the same as last, and even when it doesn't look like a usable column. We can optimize this with a couple of simple functions, which also have the benefit of unifying the logic when creating and parsing:

 ```
...
        for i, col := range cols {
                idx, ok := markerIndex(col)
                if !ok {
                        continue
                }
                ...
        }
...

const markerPrefix = "_sqlair_"

func markerName(n int) string {
        return markerPrefix + strconv.Itoa(n)
}

func markerIndex(s string) (int, bool) {
        if strings.HasPrefix(s, markerPrefix) {
                n, err := strconv.Atoi(s[len(markerPrefix):])
                if err == nil {
                        return n, true
                }
        }
        return 0, false
}
```

File: internal/expr/decode.go [hunk 66]
```diff
@@ -0,0 +1,52 @@
...
+			}
+
+			val := dest.FieldByIndex(field.index)
+			if !val.CanSet() {
+				return []any{}, fmt.Errorf("cannot set field %s of struct %s", field.name, field.structType.Name())
+			}
+			ptrs = append(ptrs, val.Addr().Interface())
+		} else {
+			// Columns not mentioned in output expressions are scanned into x.
+			// TODO: Add M type and save these columns in the map.
```
niemeyer: Can't we provide nil into the scan function to just drop them?

Also, I'm not sure if the M type comment holds. Our vague discussions in the early planning sections was to make that selection somewhat explicit too. We'll see when we get there.

niemeyer (changes requested):
Thanks for the changes, Alastair.

Here is a second pass, now focusing more on the internal organization.

internal/expr/decode.go
Aflynn50: `Rows.Scan` doesn't let you drop anything ([stackoverflow post discussing it](https://stackoverflow.com/questions/52679761/can-gos-rows-scan-ignore-fields-from-sql-query)), theres a nil pointer panic for `nil`. As far as I can work out this is the simplest way to do it.

There was a suggestion that the M type could return uncaptured columns but it was just a suggestion, I'll remove the comment.

internal/expr/decode.go
Aflynn50: I've added this in.

internal/expr/decode.go
Aflynn50: I've added a test for the first one.
The second one however I'm not sure that its actually reachable. I had left the error check in just because in the [laws of reflection](https://go.dev/blog/laws-of-reflection) it suggests you check that something is settable before setting, but as far as I can tell it will only be unsettable if it is an unexported field or a copy of a value (we require they pass a pointer). I can't work out any way to trigger the error message.
I'll remove it for now and I guess we can add it back in with some context if someone finds a way to trigger a cannot set panic.

sqlair.go
Aflynn50: Changed, I think I've managed to change all the comments mentioning these things too.

internal/expr/decode.go
Aflynn50: Made changes below to address this.

internal/expr/prepare.go
Aflynn50: Made changes below to address this.

internal/expr/info.go
Aflynn50: I've just realised that we don't even need to know the field type. In previous versions of Scan I had used it but now if there is a problem with the type returned from the database not matching the field type it will be caught by `Rows.Scan` rather than us. I've removed the field.

internal/expr/decode.go
Aflynn50: On second thoughts I've changed the `CanSet` one to an internal error.

internal/expr/decode.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+// markerIndex returns the int X from the string "_sqlair_X".
+func markerIndex(s string) (int, bool) {
```
niemeyer: This and markerName/markerPrefix are two sides of the same logic and belog together (here or there).

internal/expr/decode.go
```diff
@@ -0,0 +1,63 @@
...
+			return n, true
+		}
+	}
+	return 0, false
+}
+
+// ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
+// All the structs mentioned in the query must be in outputVals.
+// All outputVals must be of kind reflect.Struct.
+func (qe *QueryExpr) ScanArgs(columns []string, outputVals []reflect.Value) ([]any, error) {
```
niemeyer: In the previous review, ScanArgs actually takes outputArgs in by itself. The reason for this was that all the reflection manipulation is being done inside the expr package currently. It seems awkward to have that one detail being done outside, when even the values themselves are not used there but here.

internal/expr/decode.go
niemeyer, line 1: This entire file is pretty much a single function of QueryExpr, which has a file for itself (query.go). Would it make sense to move it there?

niemeyer (approved):
Thanks, last few nitpicks and LGTM!

internal/expr/decode.go
Aflynn50: Moved to `prepare.go`.

internal/expr/decode.go
Aflynn50: True, I've moved it inside `ScanArgs`

internal/expr/decode.go
Aflynn50: Makes sense, its been moved.

niemeyer (approved):
Thanks, nice to see this going in!

PR #45, opened 2023-03-08
Better errors

Aflynn50:
This PR changes the wording of a few errors to make it clearer. These improvements are mainly based on me encountering the errors due to silly mistakes and realising they are not as useful as they could be.

The changes are:
- In `parser.go`:
  - Add example of how a io type should be qualified when it is not
  - Include the raw strings in some error messages (this PR builds on #40)
  - Mention the type when its tag is malformed
  - Reword a few things
- In `prepare.go`
  - Include the raw section of string from the query that contains the error
  - Put quotes round struct types since we can have anonymous structs in a list which have an empty string for a type name (e.g. as can be see in the `db_test.go` file in PR #42).
  - Separate error if the user doesn't pass any types at all to prepare (and there is at least one type needed)
  - Seperate error if the user passes a pointer/list and rewording of an error if its anything else that is not a struct
- In `complete.go`
  - Rewording for consistency
  - Separate error if the user doesn't pass any inputs an we need at least one
  - Error for multiple of the same struct
- In `expr_test.go`
  - Removed/combine superfluous tests in the main table driven section
  - Refactor some tests into a table driven format 
  - Made testing struct names and fields consistent across the file


internal/expr/complete.go
```diff
@@ -37,16 +37,24 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
...
+				return nil, fmt.Errorf("need struct, got nil pointer")
+			}
+			v = reflect.Indirect(v)
+		default:
+			return nil, fmt.Errorf("need struct, got %s", v.Kind())
 		}
 
+		t := v.Type()
+		if _, ok := typeValue[t]; ok {
+			return nil, fmt.Errorf("more than one instance of type %q. To input different instances of the same struct a type alias must be used", t.Name())
```
niemeyer: The new error is nice, but maybe a bit verbose.. perhaps:
```
type %q provided more than once, rename one of them
```
We don't need to provide advice on how to rename. They're Go developers and can choose how to do it.

internal/expr/complete.go
```diff
@@ -37,16 +37,24 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
...
-
-		if t.Kind() != reflect.Struct {
-			return nil, fmt.Errorf("need struct, got %s", t.Kind())
+		switch v.Kind() {
+		case reflect.Struct:
+		case reflect.Pointer:
+			if v.IsNil() {
+				return nil, fmt.Errorf("need struct, got nil pointer")
+			}
+			v = reflect.Indirect(v)
```
niemeyer: Now we need to check again what's v's kind. The original logic seemed simpler and was addressing this properly.

internal/expr/complete.go
```diff
@@ -69,7 +75,11 @@ func (pe *PreparedExpr) Complete(args ...any) (ce *CompletedExpr, err error) {
 	for i, in := range pe.inputs {
 		v, ok := typeValue[in.structType]
 		if !ok {
-			return nil, fmt.Errorf(`type %s not found, have: %s`, in.structType.Name(), strings.Join(typeNames, ", "))
+			if len(typeNames) == 0 {
+				return nil, fmt.Errorf(`type %q not found, no input structs were found`, in.structType.Name())
+			} else {
+				return nil, fmt.Errorf(`type %q not found, have: %s`, in.structType.Name(), strings.Join(typeNames, ", "))
```
niemeyer: The original prefix seems vague too: type not found where?  The new 0-case also suffers similarly: no input structs were found where?

How about:

- Zero case: `type %q missing from expression`
- Other case: `type %q missing from expression, have: %`

internal/expr/parser.go
```diff
@@ -359,12 +359,12 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 
 	if id, ok := p.parseIdentifier(); ok {
 		if !p.skipByte('.') {
-			return fullName{}, false, fmt.Errorf("column %d: type not qualified", p.pos)
+			return fullName{}, false, fmt.Errorf("column %d: type %q not qualified. Types can be qualified with a db tag or an asterisk. e.g. &P.col1 or &P.*", p.pos, id)
```
niemeyer: Similar to the earlier point, this seems too verbose. Definitely happy to brainstorm on how to make the error messages more clear, but that shoudn't feel like we're writing documentation on error messages.  Perhaps:
```
column %d: unqualified type, expected %s.* or %s.<name> (db tag)
```

internal/expr/parser.go
```diff
@@ -359,12 +359,12 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
...
 		}
 
 		idField, ok := p.parseIdentifierAsterisk()
 		if !ok {
-			return fullName{}, false, fmt.Errorf("column %d: invalid identifier", p.pos)
+			return fullName{}, false, fmt.Errorf("column %d: invalid identifier following %q", p.pos, id)
```
niemeyer: This message seems bogus as it's saying that the identifier following the identifier is invalid. Maybe:
```
column %d: invalid identifier suffix for %q
```

internal/expr/prepare.go
```diff
@@ -62,11 +63,16 @@ func starCheckOutput(p *outputPart) error {
 func prepareInput(ti typeNameToInfo, p *inputPart) (field, error) {
 	info, ok := ti[p.source.prefix]
 	if !ok {
-		return field{}, fmt.Errorf(`type %s unknown, have: %s`, p.source.prefix, strings.Join(getKeys(ti), ", "))
+		ts := getKeys(ti)
+		if len(ts) == 0 {
+			return field{}, fmt.Errorf(`type %q not found, no types were passed to Prepare`, p.source.prefix)
+		} else {
+			return field{}, fmt.Errorf(`type %q not found, have: %s`, p.source.prefix, strings.Join(ts, ", "))
```
niemeyer: Same as earlier case, so would suggest a similar message. Except, the other error was prefixed with "invalid input parameter:" instead, but here the error prefix includes "expression" but not "parameter", right? So maybe:
```
parameter with type %q missing from expression
parameter with type %q missing from expression, have %s
```

internal/expr/prepare.go
```diff
@@ -175,6 +181,16 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 
 	// Generate and save reflection info.
 	for _, arg := range args {
+		if arg == nil {
+			return nil, fmt.Errorf("need struct, got nil")
+		}
+		switch k := reflect.TypeOf(arg).Kind(); k {
+		case reflect.Struct:
+		case reflect.Pointer:
+			return nil, fmt.Errorf("need struct, got pointer to %s. Prepare takes structs by value as they are only used for their type information", reflect.TypeOf(arg).Elem().Kind())
```
niemeyer: Same comments as similar logic early on about type-checking after the point dereferencing and about error message style.

internal/expr/prepare.go
```diff
@@ -50,10 +51,10 @@ func starCheckOutput(p *outputPart) error {
 
 	if targetStars > 1 || sourceStars > 1 || (sourceStars == 1 && targetStars == 0) ||
 		(starTarget && numTargets > 1) || (starSource && numSources > 1) {
-		return fmt.Errorf("invalid asterisk in output expression: %s", p)
+		return fmt.Errorf("invalid asterisk in output expression: %s", p.raw)
 	}
 	if !starTarget && (numSources > 0 && (numTargets != numSources)) {
-		return fmt.Errorf("mismatched number of cols and targets in output expression: %s", p)
+		return fmt.Errorf("mismatched number of cols and targets in output expression: %s", p.raw)
```
niemeyer: I missed this before: cols here is columns, right?  If so: s/cols/columns/

internal/expr/expr_test.go
```diff
@@ -293,281 +233,195 @@ var tests = []struct {
...
-		sql      string
-		structs  []any
-		errorstr string
-	}{{
-		sql:      "SELECT (&Person.*, &Person.*) FROM t",
-		structs:  []any{Address{}, Person{}},
-		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[] [Person.* Person.*]]",
+		query:       "SELECT (name, p.*) AS (&Person.id, &Person.*) FROM t",
+		prepareArgs: []any{Address{}, Person{}},
+		err:         "cannot prepare expression: invalid asterisk in output expression: (name, p.*) AS (&Person.id, &Person.*)",
```
niemeyer, line 346: These messages are a lot nicer, thank you.  I wonder if they'll get a bit out of hand when the expressions get larger, but that's a problem we already had and we can always turn the expression into a suffix+"..." after a given length if it becomes awkward.

internal/expr/expr_test.go
```diff
@@ -293,281 +233,195 @@ var tests = []struct {
...
-		sql      string
-		structs  []any
-		errorstr string
-	}{{
-		sql:      "SELECT (&Address.*, &Address.id) FROM t",
-		structs:  []any{Address{}, Person{}},
-		errorstr: "cannot prepare expression: invalid asterisk in output expression: Output[[] [Address.* Address.id]]",
+		query:       "SELECT (p.*, t.*) AS &Address.* FROM t",
+		prepareArgs: []any{Address{}},
+		err:         "cannot prepare expression: invalid asterisk in output expression: (p.*, t.*) AS &Address.*",
```
niemeyer, line 354: This improves it significantly, but this error is still not great. No need to fix it now, but the error message gives no clue about what's actually wrong. Imagine this with an expression that is 5 lines long.

niemeyer (changes requested):
Thank you, lots of good improvements here.

Some suggestions below.

internal/expr/complete.go
Aflynn50: Ah true, we do need to check the kind again after the Indirect, I've added in a check.

The problem with the original logic was that we never checked for a nil pointer. The `arg == nil` check only finds `nil` of type `nil`. Complete accepts either a pointer to a struct or a struct so if we get a pointer we will necessarily need to first check that it is a pointer and then that it was pointing to a struct (unless there is a better way to do it I don't know about)

internal/expr/complete.go
Aflynn50: This case is actually for a type missing from the arguments. I could put `type %q missing from parameters` instead?

internal/expr/parser.go
Aflynn50: This certainly seems clearer and more concise, I'd be happy with this.

internal/expr/complete.go
Aflynn50: Following discussion on mattermost I've changed this to check `v.Kind() == reflect.Invalid || v.IsNil() ` after taking the `reflect.ValueOf` instead.

internal/expr/parser.go
Aflynn50: How about `invalid identifier suffix following %q`?

internal/expr/prepare.go
Aflynn50: As with the earlier case, this is the error for when the user doesn't pass the type to `Prepare`, so maybe `type %q not passed to Prepare` or `type %q not passed as a parameter`.

internal/expr/complete.go
Aflynn50: Or maybe `type %q not passed as a parameter`

internal/expr/prepare.go
Aflynn50: This case is slightly different since we only deal with the type of `arg` rather than the value. Even though getting the type a second time with `reflect.TypeOf(arg).Elem().Kind()` is pretty inefficient I was thinking its ok since it only happens when the error is thrown.

internal/expr/prepare.go

internal/expr/expr_test.go
Aflynn50: Hopefully nobody will be writing a 5 line long output expression, but yea, point taken.

internal/expr/expr_test.go
Aflynn50: True, though hopefully most the time the separate output structs will be in seperate output expressions and we shouldn't end up with massive chunks of text here.

PR #46, opened 2023-03-09
Change SQLAir to SQLair

Aflynn50:
This PR changes SQLAir -> SQLair. The new version matches the capitalisation of SQLite/DQLite and makes more sense as a pun ("Lair" sounds like "layer" and this library is a compatibility layer).


SimonRichardson:
Interesting, this was closer to why I called it SQLair in the first place.

> SQLite Query Layer

Which was meant as a backronym for SQL as well. That never stuck though.

niemeyer (approved):
Sounds good. :)

PR #48, opened 2023-03-20
Add support for map inputs and outputs

Aflynn50:
This PR allows users to pass map arguments to the query. The maps must have a key type of kind `reflect.String`. Maps a treated in a similar way to structs.

For inputs one can write in the query  `x = $M.key` and then pass a map called M with key `key` and the value stored there will be passed to the database as the argument.

For outputs one can either write `(col1, col2) AS &M.*` or `&M.col1`. Here, the result columns will be stored with key `col1`/`col2` in the map passed to `Decode`. 

We also add a type `sqlair.M = map[string]any` that is of a valid type to be used with `sqlair`.


sqlair.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/canonical/sqlair/internal/expr"
 )
 
+// sqlair-provided M-type map.
+type M = map[string]any
+
 // Statement represents a SQL statemnt with valid SQLair expressions.
```
solmyrcan: statemnt -> statement

internal/expr/info.go
```diff
@@ -4,6 +4,18 @@ import (
 	"reflect"
 )
 
+type typeElement interface {
```
TheSignPainter98: I'm not sure 'element' has the right connotation here, it feels a map or slice element but `field` also implements this. Perhaps `typeInfo` would work? (Which would require changing the existing `typeInfo` function, perhaps to `typeInfoOf` matching `reflect.TypeOf`?)

internal/expr/prepare.go
```diff
@@ -192,20 +214,33 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+			}
+			name := info.typ.Name()
+			if name == mapName {
+				return nil, fmt.Errorf("name %q is reserved for maps", mapName)
+			}
+			ti[name] = info
+		case reflect.Map:
+			return nil, fmt.Errorf("got map, maps do not need to be passed as parameters")
+		case reflect.Pointer:
+			return nil, fmt.Errorf("need struct, got pointer. Prepare takes structs by value as they are only used for their type information")
```
TheSignPainter98: I think this pointer error case can be absorbed into the case below it. The extra information in the error message is nice, but someone could read this, get frustrated and think 'but you know it's a pointer, why can't you dereference it to get the type info?' Either way, the first part of the message tells them what went wrong and how to fix it

internal/expr/query.go
```diff
@@ -36,93 +36,142 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		}
 	}
 
 	var typeValue = make(map[reflect.Type]reflect.Value)
 	var typeNames []string
+	var m map[string]any
 	for _, arg := range args {
 		if arg == nil {
-			return nil, fmt.Errorf("need valid struct, got nil")
+			return nil, fmt.Errorf("need a map or struct, got nil")
```
TheSignPainter98: Tiny point, I think the 'a' should be removed here to make the article usage consistent

internal/expr/query.go
```diff
@@ -36,93 +36,142 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-
-		if !typeInQuery[t] {
-			// Check if we have a type with the same name from a different package.
-			for _, in := range pe.inputs {
-				if t.Name() == in.structType.Name() {
-					return nil, fmt.Errorf("type %s not found, have %s", in.structType.String(), t.String())
+		switch t.Kind() {
+		case reflect.Map:
+			if m != nil {
+				return nil, fmt.Errorf(`multiple maps`)
```
TheSignPainter98: I think this error message needs a little more context, how about "query received multiple maps"?

internal/expr/query.go
```diff
@@ -36,93 +36,142 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			}
+			switch mtype := arg.(type) {
+			case map[string]any:
+				m = mtype
+			case *map[string]any:
```
TheSignPainter98: I'm not sure I follow, why are both the `map[string]any` and `*map[string]any` cases required? Would it be cleaner to enforce just one to be consistent with `*struct{}` being forbidden elsewhere? It also feels a little strange to see the pointer case guarded with a `.Kind() == reflect.Map`

internal/expr/query.go
```diff
@@ -36,93 +36,142 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+							return nil, fmt.Errorf("type %s not found, have %s", te.structType.String(), t.String())
+						}
+					}
 				}
+				return nil, fmt.Errorf("%s not referenced in query", t.Name())
 			}
-
-			return nil, fmt.Errorf("%s not referenced in query", t.Name())
+		default:
+			return nil, fmt.Errorf("need struct, got %s", t.Kind())
```
TheSignPainter98: I think this needs amending to `"need struct or map, got %s"`

internal/expr/query.go
```diff
@@ -36,93 +36,142 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			}
+			typeDest[outputVal.Type()] = outputVal
+		case reflect.Map:
+			if m != nil {
+				return nil, nil, fmt.Errorf(`found multiple map types`)
+			}
+			switch arg := outputArg.(type) {
+			case map[string]any:
+				m = arg
+			case *map[string]any:
```
TheSignPainter98: If only one of `map[string]any` and `*map[string]any` were supported (as above), this would also need changing

sqlair.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/canonical/sqlair/internal/expr"
 )
 
+// sqlair-provided M-type map.
+type M = map[string]any
```
TheSignPainter98: I'm not sure what this doc adds, perhaps it would be helpful to hint at its use with something like this?
```
// M represents an arbitrary number of arbitrarily-typed values. Sqlair can
// populate it with query results.
```

internal/expr/query.go
```diff
@@ -132,19 +181,43 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
+			keyIndex[te.name] = len(mapPtrs) - 1
 		}
-		ptrs = append(ptrs, val.Addr().Interface())
 	}
-	return ptrs, nil
+	return ptrs, &MapDecodeInfo{mapPtrs: mapPtrs, keyIndex: keyIndex, m: m}, nil
+}
+
+// PopulateMap enters the scanned values into the map.
+func (mapDecodeInfo *MapDecodeInfo) PopulateMap() {
```
TheSignPainter98: This is really small, but somehow 'puts' feels more natural to me? Also should the 'the' be removed?

Will `PopluateMap` is the only method on this type, perhaps just `Populate` would be cleaner?

internal/expr/info.go
Aflynn50: I feel like with `typeInfo` you wouldn't expect there to be multiple of them for one struct or map, also the main thing it stores is the "element" of the type, i.e. the field or key of the type that we are interested in. The other options that I was considering were:
- typeElement
- typeMember
- typeField
- typeIndex 

Its an annoyingly hard thing to name. Do you think any of those are better, or can think of any others?

internal/expr/prepare.go
Aflynn50: I've gotten rid of the longer one.

internal/expr/query.go

internal/expr/query.go
Aflynn50: The newest changes mean this isnt an error anymore!

internal/expr/query.go
Aflynn50: In query we allow both `struct` and `*struct`. With map you can get away with only passing by value since the map object itself is already a pointer, but I decided for consistency it'd be best to allow people to pass a pointer to a map since it does no harm.

internal/expr/query.go

sqlair.go
Aflynn50: Updated this comment to include an example.

internal/expr/query.go

internal/expr/query.go
TheSignPainter98: Ah okay thanks for the explanation. Consistency sounds good

internal/expr/info.go
TheSignPainter98: I spent a bit thinking about this too, you're right `typeInfo` also has the wrong vibe. I quite like `typeMember`, the others feel specific to a particular kind

internal/expr/expr_test.go
```diff
@@ -33,6 +34,10 @@ type District struct{}
 
 type M map[string]any
 
+type OtherMap map[string]any
+
+type StringMap map[string]string
+
 var tests = []struct {
 	summery          string
```
solmyrcan: summary :P

internal/expr/expr_test.go
```diff
@@ -33,6 +34,10 @@ type District struct{}
 
 type M map[string]any
 
+type OtherMap map[string]any
+
```
solmyrcan, line 45: Do you think we could add one more, as `map[string]int`, to add a bit of more variety ?

internal/expr/expr_test.go
```diff
@@ -150,6 +155,12 @@ var tests = []struct {
 	"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[, ] Output[[a.*] [Address.*]] Bypass[ FROM person, address a WHERE name = 'Fred']]",
 	[]any{Person{}, Address{}},
 	"SELECT address_id AS _sqlair_0, id AS _sqlair_1, name AS _sqlair_2, a.district AS _sqlair_3, a.id AS _sqlair_4, a.street AS _sqlair_5 FROM person, address a WHERE name = 'Fred'",
+}, {
+	"map output",
```
solmyrcan: Do you think we could call this "multiple map star outputs, single map input", or something similar, just to be more specific, in case we add tests with different combinations later on. 
It's because outputs are `.*` kind, and there are more than one, and there is also a map as input.

internal/expr/expr_test.go
```diff
@@ -150,6 +155,12 @@ var tests = []struct {
...
+	"SELECT (p.name, a.id) AS &M.*, street AS &StringMap.*, &OtherMap.street FROM person, address a WHERE name = $M.name",
+	"[Bypass[SELECT ] Output[[p.name a.id] [M.*]] Bypass[, ] Output[[street] [StringMap.*]] Bypass[, ] Output[[] [OtherMap.street]] Bypass[ FROM person, address a WHERE name = ] Input[M.name]]",
+	[]any{sqlair.M{}, OtherMap{}, StringMap{}},
+	"SELECT p.name AS _sqlair_0, a.id AS _sqlair_1, street AS _sqlair_2, street AS _sqlair_3 FROM person, address a WHERE name = @sqlair_0",
 }, {
 	"multicolumn output v1",
```
solmyrcan, line 172: Now that we have maps, maybe we could call this one something like "multicolumn output with structs v1", or something similar.

internal/expr/expr_test.go
```diff
@@ -240,6 +251,12 @@ var tests = []struct {
 	"[Bypass[INSERT INTO person (name) VALUES ] Input[Person.name]]",
 	[]any{Person{}},
 	`INSERT INTO person (name) VALUES @sqlair_0`,
+}, {
+	"insert with map",
```
solmyrcan: How about "insert with map inputs" ?

internal/expr/expr_test.go
```diff
@@ -564,7 +584,58 @@ func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
...
+	tests := []struct {
+		summary string
+		input   string
+		args    []any
+		expect  string
+	}{{
+		"all output into map star",
+		"SELECT &M.* FROM person WHERE name = 'Fred'",
+		[]any{sqlair.M{}},
+		"cannot prepare expression: &M.* cannot be used with a map when no column names are specified or column name is *",
```
solmyrcan: Do you think we could slightly alter this error message to:
... cannot be used as map output ...
instead of
... cannot be used with a map ...
This would apply to the two others below, and the code line where it's returned.

internal/expr/expr_test.go
```diff
@@ -564,7 +584,58 @@ func (s *ExprSuite) TestPrepareMismatchedColsAndTargs(c *C) {
...
+		"SELECT * AS &CustomMap.* FROM person WHERE name = 'Fred'",
+		[]any{CustomMap{}},
+		"cannot prepare expression: &CustomMap.* cannot be used with a map when no column names are specified or column name is *",
+	}, {
+		"invalid map",
+		"SELECT * AS &invalidMap.* FROM person WHERE name = 'Fred'",
+		[]any{invalidMap{}},
+		"cannot prepare expression: map type invalidMap must have key type string, found type int",
+	}, {
+		"clashing names",
```
solmyrcan: Maybe we could say "clashing map names" ?

internal/expr/info.go
solmyrcan: I also think `typeMember` would be slightly better. Another one I can think of is `typeContent`, where field is content of struct, and key is content of map, maybe?

internal/expr/prepare.go
```diff
@@ -127,34 +137,47 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []field, error
...
 		if len(p.source) == 0 || p.source[0].name == "*" {
-			pref := ""
-
-			// Prepend table name. E.g. "t" in "t.* AS &P.*".
-			if len(p.source) > 0 {
-				pref = p.source[0].prefix
-			}
+			switch info := info.(type) {
+			case *mapInfo:
+				return nil, nil, fmt.Errorf(`&%s.* cannot be used with a map when no column names are specified or column name is *`, info.typ().Name())
```
solmyrcan: I think this is the line where we could change the string to "...cannot be used as map output..."

internal/expr/prepare.go
```diff
@@ -127,34 +137,47 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []field, error
...
+				for _, tag := range info.tags {
+					outCols = append(outCols, fullName{pref, tag})
+					typeElements = append(typeElements, info.tagToField[tag])
+				}
+				return outCols, typeElements, nil
 			}
-			return outCols, fields, nil
 		}
 
 		// Case 1.2: Explicit columns e.g. "(col1, t.col2) AS &P.*".
```
solmyrcan, line 170: Would we want to remove the case annotations, `Case 1.2` etc, from the comments ?

internal/expr/query.go
```diff
@@ -36,92 +36,120 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 
 	var typeValue = make(map[reflect.Type]reflect.Value)
 	var typeNames []string
 	for _, arg := range args {
 		if arg == nil {
-			return nil, fmt.Errorf("need valid struct, got nil")
+			return nil, fmt.Errorf("need map or struct, got nil")
 		}
 		v := reflect.ValueOf(arg)
 		v = reflect.Indirect(v)
```
solmyrcan, line 55: Could these two be combined in a single line, as in: 
`v := reflect.Indirect(reflect.ValueOf(arg))`

internal/expr/query.go
```diff
@@ -36,92 +36,120 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		v, ok := typeValue[typeElement.outerType()]
 		if !ok {
-			return nil, fmt.Errorf(`type %s not found, have: %s`, in.structType.Name(), strings.Join(typeNames, ", "))
+			return nil, fmt.Errorf(`type %s not found, have: %s`, typeElement.outerType().Name(), strings.Join(typeNames, ", "))
+		}
+		switch te := typeElement.(type) {
+		case field:
+			qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(i), v.FieldByIndex(te.index).Interface()))
+		case mapKey:
+			// MapIndex returns a zero value of the key is not in the map so we
```
solmyrcan: "... a zero value if the key ..."

internal/expr/query.go
```diff
@@ -36,92 +36,120 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 
+type MapDecodeInfo struct {
+	m         reflect.Value
+	valuePtrs []any
+	keyIndex  map[string]int
+}
+
 // ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
 // All the structs mentioned in the query must be in outputArgs.
 // All outputArgs must be structs.
```
solmyrcan: I guess we  now need to also mention maps in the above three lines of comments.

internal/expr/query.go
```diff
@@ -36,92 +36,120 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			if k != reflect.Pointer {
+				return nil, nil, fmt.Errorf("need map or pointer to struct, got %s", k)
+			}
+			if outputVal.IsNil() {
+				return nil, nil, fmt.Errorf("got nil pointer")
+			}
+			outputVal = outputVal.Elem()
+			k = outputVal.Kind()
+			if k != reflect.Struct && k != reflect.Map {
+				return nil, nil, fmt.Errorf("need map or pointer to struct, got pointer to %s", k)
```
solmyrcan, line 140: Should this instead say "need pointer to map or struct, got pointer to %s" ?

internal/expr/query.go
```diff
@@ -36,92 +36,120 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 	outputVals := []reflect.Value{}
 	for _, outputArg := range outputArgs {
 		if outputArg == nil {
-			return nil, fmt.Errorf("need pointer to struct, got nil")
+			return nil, nil, fmt.Errorf("need map or pointer to struct, got nil")
 		}
 		outputVal := reflect.ValueOf(outputArg)
-		if outputVal.Kind() != reflect.Pointer {
-			return nil, fmt.Errorf("need pointer to struct, got %s", outputVal.Kind())
+		k := outputVal.Kind()
```
solmyrcan, line 129: Would it be a good idea to have a switch block for Map and Pointer types?

```Go
switch k := outputVal.Kind() {
case reflect.Pointer:
    // do Pointer-specific error stuff
    // then do Pointer stuff
case reflect.Map:
    // do Map-specific error stuff
    // then do Map stuff
default:
    // do other error stuff
}
```

internal/expr/query.go
```diff
@@ -132,19 +160,45 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
+			mapDecodeInfo, ok := mapDecodeInfos[te.mapType]
+			if !ok {
+				return nil, nil, fmt.Errorf("type %q found in query but not passed to decode", te.mapType.Name())
+			}
+			// Scan in to a new variable with the type of the maps value
+			val := reflect.New(te.mapType.Elem()).Elem()
+			addr := val.Addr().Interface()
+			ptrs = append(ptrs, addr)
+			mapDecodeInfo.valuePtrs = append(mapDecodeInfo.valuePtrs, addr)
+			mapDecodeInfo.keyIndex[te.name] = len(mapDecodeInfo.valuePtrs) - 1
```
solmyrcan: Could we have a map of type `map[string]any`  instead? (not talking about the M type here :P ) 
i.e. instead of keeping an index to the slice `valuePtrs`, which is a slice of `Interface{}`, could we directly have the map to keep the `Interface{}`s ? Then we wouldn't need to the `valuePtrs` slice I think.

With that, we would have:
```Go
type MapDecodeInfo struct {
	m         reflect.Value
	kv        map[string]any
}
```
Then later, we could do:

```Go
mapDecodeInfo.m.SetMapIndex(reflect.ValueOf(k), reflect.ValueOf(mapDecodeInfo.kv[k])
```

internal/expr/query.go
```diff
@@ -132,19 +160,45 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
-		if !val.CanSet() {
-			return nil, fmt.Errorf("internal error: cannot set field %s of struct %s", field.name, field.structType.Name())
-		}
-		ptrs = append(ptrs, val.Addr().Interface())
+// Populate puts scanned values into the map.
+func (mapDecodeInfo *MapDecodeInfo) Populate() {
+	for k, i := range mapDecodeInfo.keyIndex {
+		// Is there a better way to pass the set values here than their
+		// pointers?
+		mapDecodeInfo.m.SetMapIndex(reflect.ValueOf(k), reflect.ValueOf(mapDecodeInfo.valuePtrs[i]).Elem())
```
solmyrcan: As in the previous comment, maybe we could instead do (please see previous comment for `kv` here):

```Go
mapDecodeInfo.m.SetMapIndex(reflect.ValueOf(k), reflect.ValueOf(mapDecodeInfo.kv[k])
```

internal/expr/typeinfo.go
```diff
@@ -42,46 +42,51 @@ func typeInfo(value any) (*info, error) {
...
 
-	sort.Strings(tags)
-	info.tags = tags
+		sort.Strings(tags)
+		structInfo.tags = tags
 
-	return &info, nil
+		return &structInfo, nil
+	default:
+		return nil, fmt.Errorf("internal error: attempted to obtain struct information for something that is not a struct: %s.", t)
```
solmyrcan: I guess maps need to be mentioned here as well:
"internal error: attempted to obtain type information for something that is neither a struct nor a map: %s"

sqlair.go

internal/expr/expr_test.go
Aflynn50: Ah, this is because the better errors PR #45 isn't in yet, it gets changed there.

internal/expr/expr_test.go
Aflynn50: Changed it to that.

internal/expr/expr_test.go
Aflynn50: True, I'll change it to "map input and output" to keep it relatively concise

internal/expr/expr_test.go
Aflynn50: True, though these summary names are really just for finding the tests when one fails, I'm not sure its worth adding a ton of name changes to this PR.

internal/expr/expr_test.go

internal/expr/expr_test.go
Aflynn50: Changed

internal/expr/expr_test.go
Aflynn50: Changed to clashing map and struct names.

internal/expr/prepare.go

internal/expr/prepare.go
Aflynn50: I think they should still be applicable

internal/expr/query.go
Aflynn50: They certainly can!

internal/expr/query.go

internal/expr/query.go

internal/expr/query.go
Aflynn50: Maps can be passed as a value, since they are themsevles a pointer under the hood, so they can pass one as a pointer if they want, but also by value.

internal/expr/query.go
Aflynn50: I was thinking of doing this, but we have the outputVal.Elem() in the middle that changes the Kind of outputVal half way though and makes it all a bit messy.

internal/expr/query.go
Aflynn50: True, thats a good point, I'll have a go at implementing that.

Aflynn50: I've changed it! I used two slices rather than a map since in the end all we need to do is cycle through all key/value pairs rather than actually using it like a map.

internal/expr/query.go
Aflynn50: Changed!

internal/expr/typeinfo.go
Aflynn50: Changed.

internal/expr/query.go
solmyrcan: Do you think we could do `n = outputVal.Elem().Kind()` there and then:
`if n != reflect.Struct && n != reflect.Map`
If I'm reading it correctly, this might would avoid that shadowing, and still happily have a switch block.

internal/expr/query.go
solmyrcan: Ah I see, thanks for pointing that out :)

internal/expr/info.go
Aflynn50: That would be good too, its currently `typeMember`

internal/expr/info.go
```diff
@@ -19,12 +32,31 @@ type field struct {
...
 }
 
-// Info represents reflected information about a struct type.
-type info struct {
-	typ reflect.Type
+func (f structField) outerType() reflect.Type {
+	return f.structType
+}
+
+type info interface {
```
niemeyer: Let's please take the chance to fix this overly generic name. Maybe `typeInfo` and rename the function to `getTypeInfo`?

internal/expr/prepare.go
```diff
@@ -77,54 +77,63 @@ func starCheckOutput(p *outputPart) error {
...
+	case *mapInfo:
+		return mapKey{name: p.source.name, mapType: info.typ()}, nil
+	case *structInfo:
+		f, ok := info.tagToField[p.source.name]
+		if !ok {
+			return nil, fmt.Errorf(`type %q has no %q db tag`, info.typ().Name(), p.source.name)
+		}
+		return f, nil
+	default:
+		return nil, fmt.Errorf(`internal error: unknow info type: %T`, info)
```
niemeyer: "unknown"

internal/expr/prepare.go
```diff
@@ -77,54 +77,63 @@ func starCheckOutput(p *outputPart) error {
...
+			switch info := info.(type) {
+			case *mapInfo:
+				typeMembers = append(typeMembers, mapKey{name: t.name, mapType: info.typ()})
+			case *structInfo:
+				f, ok := info.tagToField[t.name]
+				if !ok {
+					return nil, nil, fmt.Errorf(`type %q has no %q db tag`, info.typ().Name(), t.name)
+				}
+				typeMembers = append(typeMembers, f)
+			}
```
niemeyer, line 138: Missing default case.

internal/expr/prepare.go
```diff
@@ -133,34 +142,47 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []field, error
...
 		if len(p.source) == 0 || p.source[0].name == "*" {
-			pref := ""
-
-			// Prepend table name. E.g. "t" in "t.* AS &P.*".
-			if len(p.source) > 0 {
-				pref = p.source[0].prefix
-			}
+			switch info := info.(type) {
+			case *mapInfo:
+				return nil, nil, fmt.Errorf(`&%s.* cannot be used with a map output when no column names are specified or column name is *`, info.typ().Name())
```
niemeyer: The first case is implied on the second case (if column name is * then no column name was specified). So:
```
&%s.* cannot be used for maps when no source column names are specified
```
We might flex that by attempting to read all columns, but seems fine for now.

internal/expr/prepare.go
```diff
@@ -133,34 +142,47 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []field, error
...
 
-			for _, tag := range info.tags {
-				outCols = append(outCols, fullName{pref, tag})
-				fields = append(fields, info.tagToField[tag])
+				for _, tag := range info.tags {
+					outCols = append(outCols, fullName{pref, tag})
+					typeMembers = append(typeMembers, info.tagToField[tag])
+				}
+				return outCols, typeMembers, nil
 			}
```
niemeyer, line 167: Missing default case.

internal/expr/prepare.go
```diff
@@ -133,34 +142,47 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []field, error
...
+					if !ok {
+						return nil, nil, fmt.Errorf(`type %q has no %q db tag`, info.typ().Name(), c.name)
+					}
+					outCols = append(outCols, c)
+					typeMembers = append(typeMembers, f)
 				}
-				outCols = append(outCols, c)
-				fields = append(fields, f)
+				return outCols, typeMembers, nil
 			}
```
niemeyer, line 191: Missing default case.

internal/expr/query.go
```diff
@@ -40,34 +40,33 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		inQuery[typeMember.outerType()] = true
 	}
 
 	var typeValue = make(map[reflect.Type]reflect.Value)
 	var typeNames []string
 	for _, arg := range args {
 		v := reflect.ValueOf(arg)
 		if v.Kind() == reflect.Invalid || (v.Kind() == reflect.Pointer && v.IsNil()) {
-			return nil, fmt.Errorf("need struct, got nil")
+			return nil, fmt.Errorf("need map or struct, got nil")
```
niemeyer: This is pretty minor, but can we please reorder all of those messages so they say "struct or map" rather than the "map or struct"? We want "struct" to be the norm, and map to be the exception.

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 
 	// Query parameteres.
 	qargs := []any{}
-
-	for i, in := range pe.inputs {
-		v, ok := typeValue[in.structType]
+	for i, typeMember := range pe.inputs {
+		v, ok := typeValue[typeMember.outerType()]
```
niemeyer: Instead of repeating this method call so many times, we can just save the value once and use it:
```
outerType := typeMember.outerType()
```

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		}
+		switch tm := typeMember.(type) {
+		case structField:
+			qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(i), v.FieldByIndex(tm.index).Interface()))
+		case mapKey:
+			// MapIndex returns a zero value if the key is not in the map so we
+			// need to check the MapKeys().
+			var val reflect.Value
+			for _, key := range v.MapKeys() {
+				if key.String() == tm.name {
```
niemeyer: That's doing a linear iteration over a structure that is heavily optimized for key searches to that it can.... search for a key?  Doesn't it smell awkward? :-)

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-		qargs = append(qargs, named)
 	}
-
 	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
+// MapDecodeInfo stores a map and the results to go in it.
+// Once values have been scanned into the values slice Populate() must be called
+// to set the values in the map.
+type MapDecodeInfo struct {
```
niemeyer: Most of the adapted logic in this PR feels pretty clean, but this part looks like a bolted on hack on what used to be a pretty neat API.  We would definitely not have arrived here if we had considered the whole picture upfront, so it needs a step back and reconsideration of what's the underlying idea and the proper abstraction.

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-// All outputArgs must be structs.
-func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error) {
+// All the structs and maps mentioned in the query must be in outputArgs.
+func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, []*MapDecodeInfo, error) {
+	var typesInQuery = []string{}
+	var inQuery = make(map[reflect.Type]bool)
+	for _, typeMember := range qe.outputs {
+		if ok := inQuery[typeMember.outerType()]; !ok {
+			inQuery[typeMember.outerType()] = true
+			typesInQuery = append(typesInQuery, typeMember.outerType().Name())
```
niemeyer: How many `typeMember.outerType()` calls are too many? :-)

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			for _, key := range v.MapKeys() {
+				if key.String() == tm.name {
+					val = v.MapIndex(reflect.ValueOf(tm.name))
+					break
+				}
+			}
+			if !val.IsValid() {
+				return nil, fmt.Errorf(`map does not contain key %q`, tm.name)
 			}
+			qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(i), val.Interface()))
```
niemeyer: This call is almost exactly the same across the two branches. Note that even the `.Interface()` call is the same. So let's just save the reflect.Value and do the rest commonly following the switch.

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-		outputVal = reflect.Indirect(outputVal)
-		if outputVal.Kind() != reflect.Struct {
-			return nil, fmt.Errorf("need pointer to struct, got pointer to %s", outputVal.Kind())
+		if outputVal.Kind() == reflect.Map {
+			mapDecodeInfos[outputVal.Type()] = &MapDecodeInfo{m: outputVal}
+		} else {
+			if _, ok := typeDest[outputVal.Type()]; ok {
+				return nil, nil, fmt.Errorf("type %q provided more than once, rename one of them", outputVal.Type().Name())
+			}
+			typeDest[outputVal.Type()] = outputVal
```
niemeyer: Again, in the spirit of the above comment, it's not clear why the branches above are different. They are doing essentially the same thing, tracking the type with `typeDest[outputVal.Type()] = outputVal` for maps looks fine, and the error seems equally valid for both.

internal/expr/query.go
```diff
@@ -147,19 +174,41 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
+		case structField:
+			outputVal, ok := typeDest[tm.structType]
+			if !ok {
+				return nil, nil, fmt.Errorf("type %q found in query but not passed to get", tm.structType.Name())
+			}
+			val := outputVal.FieldByIndex(tm.index)
+			if !val.CanSet() {
+				return nil, nil, fmt.Errorf("internal error: cannot set field %s of struct %s", tm.name, tm.structType.Name())
+			}
+			ptrs = append(ptrs, val.Addr().Interface())
```
niemeyer, line 177: Similarly here, the logic above is more similar than it is different, and indeed it's being somewhat repeated below. I'd try to unify them further.

internal/expr/query.go
```diff
@@ -147,19 +174,41 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
+		case mapKey:
+			mapDecodeInfo, ok := mapDecodeInfos[tm.mapType]
+			if !ok {
+				return nil, nil, fmt.Errorf("type %q found in query but not passed to get", tm.mapType.Name())
+			}
+			// Scan in to a new value with the type of the maps value
+			mapDecodeInfo.keys = append(mapDecodeInfo.keys, tm.name)
+			mapDecodeInfo.values = append(mapDecodeInfo.values, reflect.New(tm.mapType.Elem()).Elem())
+			ptrs = append(ptrs, mapDecodeInfo.values[len(mapDecodeInfo.values)-1].Addr().Interface())
```
niemeyer: The last line is the same as above too. THe only difference is where val comes from, and ideally we'd have a handle for it instead of losing the reference and then digging back into the slice.

internal/expr/query.go
```diff
@@ -76,67 +75,95 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-	var typeDest = make(map[reflect.Type]reflect.Value)
-	for _, outputVal := range outputVals {
-		t := outputVal.Type()
-		if !inQuery[t] {
-			return nil, fmt.Errorf("type %q does not appear in query, have: %s", t.Name(), strings.Join(typesInQuery, ", "))
-		}
-		if _, ok := typeDest[t]; ok {
-			return nil, fmt.Errorf("type %q provided more than once, rename one of them", t.Name())
-		}
-		typeDest[t] = outputVal
```
niemeyer, line 137: Following on the above, worth noting how maps fit right in with the logic here, with no changes other than field.structType to member.outerType.

internal/expr/query.go
```diff
@@ -147,19 +174,41 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
+			mapDecodeInfo.keys = append(mapDecodeInfo.keys, tm.name)
+			mapDecodeInfo.values = append(mapDecodeInfo.values, reflect.New(tm.mapType.Elem()).Elem())
+			ptrs = append(ptrs, mapDecodeInfo.values[len(mapDecodeInfo.values)-1].Addr().Interface())
 		}
+	}
+	var mapInfoSlice = []*MapDecodeInfo{}
+	for _, v := range mapDecodeInfos {
+		mapInfoSlice = append(mapInfoSlice, v)
+	}
+	return ptrs, mapInfoSlice, nil
```
niemeyer: Perhaps as a possible way out of the API issue here, we can define this and return as the second result here. This logic agrees with the other comments above, in that most of the logic is shared between structs and maps.
```
onSuccess := func() {
	if !hasMap {
		return // Save time for the common case.
	}
	for pos, column := range columns {
		idx, ok := markerIndex(column)
		if !ok {
			continue
		}
		typeMember, ok := qe.outputs[idx]
		if !ok {
			continue
		}
		tm, ok := typeMember.(mapKey); ok {
		if !ok {
			continue
		}
		outputVal := typeDest[tm.mapType]
		outputVal.SetMapIndex(reflect.ValueOf(tm.name, ptrs[pos])
	}
}
```

internal/expr/typeinfo.go
```diff
@@ -41,47 +41,52 @@ func typeInfo(value any) (*info, error) {
...
-	tags := []string{}
-
-	for i := 0; i < t.NumField(); i++ {
-		f := t.Field(i)
-		// Fields without a "db" tag are outside of SQLair's remit.
-		tag := f.Tag.Get("db")
-		if tag == "" {
-			continue
+// reflect.Value that is specifically required for SQLAir operation.
+func generate(t reflect.Type) (info, error) {
```
niemeyer: This also deserves a less generic name (generateTypeInfo, for example).

internal/expr/typeinfo.go
```diff
@@ -41,47 +41,52 @@ func typeInfo(value any) (*info, error) {
...
+	switch t.Kind() {
+	case reflect.Map:
+		if t.Key().Kind() != reflect.String {
+			return nil, fmt.Errorf(`map type %s must have key type string, found type %s`, t.Name(), t.Key().Kind())
 		}
-		if !f.IsExported() {
-			return nil, fmt.Errorf("field %q of struct %s not exported", f.Name, t.Name())
+		return &mapInfo{mapType: t}, nil
+	case reflect.Struct:
+		structInfo := structInfo{
```
niemeyer: This is clobbering the type name with the variable name which makes the code unnecessarily confusing.

niemeyer (changes requested):
This looks very nice overall, Alastair. There's only one piece that seems worth refactoring a bit, and the rest is just trivial tuning.

niemeyer (changes requested):
This looks very nice overall, Alastair. There's only one piece that seems worth refactoring a bit, and the rest is just trivial tuning.

niemeyer (changes requested):
This looks very nice overall, Alastair. There's only one piece that seems worth refactoring a bit, and the rest is just trivial tuning.

internal/expr/prepare.go
By reading all columns do you mean put in a wildcard e.g. `SELECT * FROM ...`? I would be hesitant to do this considering that its generally bad practice to use wildcards, and if we removed this error and selected everything instead then there's a chance the user is doing by mistake (since it would be perfectly valid to do `&P.*` if `P` was a struct)

internal/expr/prepare.go
Aflynn50: Added cases.

internal/expr/prepare.go

internal/expr/info.go
Aflynn50: Changed.

internal/expr/prepare.go
Aflynn50: Added.

internal/expr/prepare.go
Aflynn50: Added.

internal/expr/query.go

internal/expr/query.go

internal/expr/query.go
Aflynn50: Added var.

internal/expr/query.go

internal/expr/typeinfo.go
Aflynn50: Changed.

internal/expr/typeinfo.go
Aflynn50: Changed, also merge info.go and typeinfo.go since there is no reason for them to be separate files

internal/expr/query.go
Aflynn50: For the case of the first branch
i.e.:
```Go
if k != reflect.Map {
	if k != reflect.Pointer {
		...
```
This makes a special case for maps that have been passed directly in (not as a reference). With Maps this causes no problems and it would be an artificial constraint to only allow them to be passed as pointers. We also allow them to be passed as pointers in which case the branch code will be triggered and the error checks will be carried out (since k will be `reflect.Pointer`)

For the branch in that block I have removed the special case for map. You're right that that error seems equally valid for both.

internal/expr/query.go
Aflynn50: Done, see comment below.

internal/expr/query.go

internal/expr/query.go
Aflynn50: I've used the idea of closures but instead of repeating the logic for checking the columns I've created a closure for each `SetMapIndex`, put these in a list and then had `onSuccess` iterate through and call each of these closures. This works pretty well because it really minimises the repeated code.

Let me know if you would still rather it done the way you've given here though.

internal/expr/query.go
Aflynn50: It does, but as far as I've been able to work out this is the only way to ensure that a key is in the map.

As explained in the comment above this loop, if we use `MapIndex` on the reflected value a zero value is returned when the key is not in the map which is indistinguishable from a key that is in the map and is set to zero.

It would be good if we could simply use it as a map, rather than a reflected value but then we run into the problem of not knowing the maps value type. We allow people to pass `map[string]any` or `map[string]int` ect.. We cant index the map without knowing its type and as far as I know there is no way to cast to a specific concrete type at runtime with reflect (without enumerating all possibilities).

We could of course simple set the argument to the zero value, but to me that would feel like creating an easy way to introduce errors (for example a spelling mistake in the key name would go undetected).

Aflynn50: Changed to use `MapIndex`. It turned out it returns a `reflect.Invalid` if it cannot find the map key instead of a zero value the map value type.

internal/expr/query.go
Aflynn50: Changed to save the information in a struct rather than a list of closures

niemeyer (approved):
Thanks! Didn't expect to have maps working so soon. Great to have it.

PR #49, opened 2023-03-21
Add All

Aflynn50:
This PR adds the `All` method to `Query`. This method takes pointers to slices that contain each type in the query, these are then populated with the rows returned from the query.

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -89,7 +89,7 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
-// ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
+// ScanArgs returns list of pointers to the struct fields outputArgs that are listed in qe.outputs.
 // All the structs mentioned in the query must be in outputArgs.
 // All outputArgs must be structs.
 func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error) {
@@ -128,7 +128,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
 			return nil, fmt.Errorf("type %q does not appear in query, have: %s", t.Name(), strings.Join(typesInQuery, ", "))
 		}
 		if _, ok := typeDest[t]; ok {
-			return nil, fmt.Errorf("type %q provided more than once, rename one of them", t.Name())
+			return nil, fmt.Errorf("type %q provided more than once", t.Name())
 		}
 		typeDest[t] = outputVal
 	}
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -89,7 +89,7 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
-// ScanArgs returns list of pointers to the struct fields outputArgs that are listed in qe.outputs.
+// ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
 // All the structs mentioned in the query must be in outputArgs.
 // All outputArgs must be structs.
 func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error) {
@@ -128,7 +128,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
 			return nil, fmt.Errorf("type %q does not appear in query, have: %s", t.Name(), strings.Join(typesInQuery, ", "))
 		}
 		if _, ok := typeDest[t]; ok {
-			return nil, fmt.Errorf("type %q provided more than once", t.Name())
+			return nil, fmt.Errorf("type %q provided more than once, rename one of them", t.Name())
 		}
 		typeDest[t] = outputVal
 	}
--- a/sqlair.go
+++ b/sqlair.go
@@ -165,19 +165,27 @@ func (q *Query) One(outputArgs ...any) error {
 	return iter.Close()
 }
 
-// All iterates over the query and decodes all the rows.
-// resultColumns must contain pointers to slices of each of the query arguments e.g.
-// > q.All(&[]*Person{}, &[]*Address{}) or q.All(&[]Address{}, &[]Person{})
-// It fabricates all structs as needed.
-func (q *Query) All(slicePointers ...any) error {
+// All iterates over the query and decodes all rows into the provided slices.
+//
+// For example:
+//
+//	var pslice []Person
+//	var aslice []*Address
+//	err := query.All(&pslice, &aslice)
+func (q *Query) All(sliceArgs ...any) (err error) {
 	if q.err != nil {
 		return q.err
 	}
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot populate slice: %s", err)
+		}
+	}()
 
 	// Check slice inputs
 	var slicePtrVals = []reflect.Value{}
 	var sliceVals = []reflect.Value{}
-	for _, ptr := range slicePointers {
+	for _, ptr := range sliceArgs {
 		ptrVal := reflect.ValueOf(ptr)
 		if ptrVal.Kind() != reflect.Pointer {
 			return fmt.Errorf("need pointer to slice, got %s", ptrVal.Kind())
@@ -198,20 +206,31 @@ func (q *Query) All(slicePointers ...any) error {
 		var outputArgs = []any{}
 		for _, sliceVal := range sliceVals {
 			elemType := sliceVal.Type().Elem()
-			if elemType.Kind() != reflect.Pointer {
-				return fmt.Errorf("need pointer to slice of pointers, got pointer to slice of %s", elemType.Kind())
+			var outputArg reflect.Value
+			switch elemType.Kind() {
+			case reflect.Pointer:
+				outputArg = reflect.New(elemType.Elem())
+			case reflect.Struct:
+				outputArg = reflect.New(elemType)
+			default:
+				iter.Close()
+				return fmt.Errorf("need slice of struct, got slice of %s", elemType.Kind())
 			}
-			outputArg := reflect.New(elemType.Elem())
 			outputArgs = append(outputArgs, outputArg.Interface())
 		}
 		if !iter.Decode(outputArgs...) {
 			break
 		}
 		for i, outputArg := range outputArgs {
-			sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg))
+			switch sliceVals[i].Type().Elem().Kind() {
+			case reflect.Pointer:
+				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg))
+			case reflect.Struct:
+				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg).Elem())
+			}
 		}
 	}
-	err := iter.Close()
+	err = iter.Close()
 	if err != nil {
 		return err
 	}
--- a/sqlair.go
+++ b/sqlair.go
@@ -197,7 +197,6 @@ func (q *Query) All(slicePointers ...any) error {
 	for iter.Next() {
 		var outputArgs = []any{}
 		for _, sliceVal := range sliceVals {
-			// Can we combine these lines?
 			elemType := sliceVal.Type().Elem()
 			if elemType.Kind() != reflect.Pointer {
 				return fmt.Errorf("need pointer to slice of pointers, got pointer to slice of %s", elemType.Kind())
@@ -217,7 +216,6 @@ func (q *Query) All(slicePointers ...any) error {
 		return err
 	}
 
-	// Do we need this?
 	for i, ptrVal := range slicePtrVals {
 		ptrVal.Elem().Set(sliceVals[i])
 	}
--- a/sqlair.go
+++ b/sqlair.go
@@ -224,14 +224,14 @@ func (q *Query) All(sliceArgs ...any) (err error) {
 			break
 		}
 		for i, outputArg := range outputArgs {
-			switch sliceVals[i].Type().Elem().Kind() {
+			switch k := sliceVals[i].Type().Elem().Kind(); k {
 			case reflect.Pointer:
 				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg))
 			case reflect.Struct:
 				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg).Elem())
 			default:
 				iter.Close()
-				return fmt.Errorf("internal error: output arg has unexpected kind %s", sliceVals[i].Type().Elem().Kind())
+				return fmt.Errorf("internal error: output arg has unexpected kind %s", k)
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


sqlair.go
```diff
@@ -140,6 +140,35 @@ func (q *Query) Decode(outputArgs ...any) (ok bool) {
 	return true
 }
 
+// All iterates over the query and decodes all the rows.
+// It fabricates all structs as needed.
+// It returns a list of rows, each row is a list of pointers to structs.
+// The structs appear in the syntactic order they appear in the query.
```
TheSignPainter98: I think the word 'syntactic' might be superfluous here

internal/expr/query.go
```diff
@@ -114,12 +141,18 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
...
 	for _, outputVal := range outputVals {
 		if !inQuery[outputVal.Type()] {
 			return nil, fmt.Errorf("type %q does not appear as an output type in the query", outputVal.Type().Name())
 		}
-		typeDest[outputVal.Type()] = outputVal
+	}
+	return qe.scanArgs(columns, outputVals)
+}
+
+func (qe *QueryExpr) scanArgs(columns []string, outputValues []reflect.Value) ([]any, error) {
```
TheSignPainter98: Does this need a doc?

internal/expr/query.go
```diff
@@ -86,6 +86,73 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		}
+		a.sliceVals[i] = reflect.Append(a.sliceVals[i], newVal)
+		outputVals = append(outputVals, reflect.Indirect(newVal))
+	}
+	return scanArgs(columns, outputVals, a.outputs)
+}
+
+func (a *AllIter) SetSlices() {
+	for i := range a.sliceVals {
+		a.slicePtrs[i].Elem().Set(a.sliceVals[i])
```
solmyrcan: Would it be good to:
- Check if `a.slicePtrs[i]` is not nil, before calling `Elem()`, to avoid a panic ?
- Check if `Set()` would work first here with a call to `CanSet()` to the return value of `Elem()` to avoid a panic ?

This may seem unnecessary, as I believe an `AllIter` is currently generated/populated by the `NewAllIter()` method, but if later another method gets added, returning an `AllIter`, `SetSlices()` would always make sure it can set the slices without panicking no matter where `AllIter` comes from.

internal/expr/query.go
```diff
@@ -86,6 +86,73 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
+type AllIter struct {
+	outputs   []field
+	slicePtrs []reflect.Value
+	sliceVals []reflect.Value
+	elemTypes []reflect.Type
+	isPtrElem []bool
```
solmyrcan: `slicePtrs`, `sliceVals`, `elemTypes`, and `isPtrElem` seem to have their elements 1-1 associated at each index, i.e. they are same-length slices where elements at index `i` are related. Having gone through the `NewAllIter()` method, I believe each element of a slice here is related with a slice as well. Would it be a good idea to define a struct to encapsulate the metadata for a slice, which is currently kept at `slicePtrs`, `sliceVals`, `elemTypes`, `isPtrElem`, like so:

``` Go
struct sliceInfo {
    ptr          reflect.Value
    val          reflect.Value
    elemType     reflect.Type
    isPtrElem    bool
}
```
It feels like this would make it convenient to work with a single struct, rather than multiple slices. For example, at the end of `AllIter()`, there are four `append()` lines which would become a single one by doing so.

We would then have something like:

``` Go
struct Allter {
    outputs    [] field
    sInfos     [] sliceInfo
}
```
And then within the methods of type `AllIter`, we would do, e.g.:

`a.sInfos[i].ptr` -> To get the `reflect.Pointer` for the i^th slice
`a.sInfos[i].val` -> To get the `reflect.Slice` for the i^th slice

Just an idea :)

PS: Not sure about naming of `sInfos` and `sliceInfo`, but I'm sure you could name them better than me :P

internal/expr/query.go
```diff
@@ -86,6 +86,73 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+func (a *AllIter) SetSlices() {
+	for i := range a.sliceVals {
+		a.slicePtrs[i].Elem().Set(a.sliceVals[i])
+	}
+}
+
+func (qe *QueryExpr) NewAllIter(slicePointers []any) (*AllIter, error) {
+	var inQuery = make(map[reflect.Type]bool)
+	for _, field := range qe.outputs {
+		inQuery[field.structType] = true
```
solmyrcan: Is the map type not considered yet in this PR ?

internal/expr/query.go
```diff
@@ -86,6 +86,73 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+	}
+
+	var allIter = AllIter{outputs: qe.outputs}
+	for _, ptr := range slicePointers {
+		var isPtrElem = false
```
solmyrcan: Maybe we could move this var down to right before the if block at line 141 below, which would make it easier to associate it with the context there.

internal/expr/query.go
```diff
@@ -86,6 +86,73 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			return nil, fmt.Errorf("need pointer to slice, got pointer to %s", sliceType.Kind())
+		}
+		elemType := sliceType.Elem()
+		if elemType.Kind() == reflect.Pointer {
+			isPtrElem = true
+			elemType = elemType.Elem()
+		}
+		if !inQuery[elemType] {
+			return nil, fmt.Errorf("type %q does not appear as an output type in the query", elemType.Name())
+		}
```
solmyrcan: I wonder if we could perform the sequence of operations that we currently do in a single iteration of this for loop within a separate function instead, and then call that function in the loop at each iteration. This function would be something like:

```Go
func generateSliceInfo (ptr any, inQuery map[reflect.Type]bool) sliceInfo {
    // operations done in a single iteration of the current for loop to appear here
}
```
And then we would call it in the for loop as:
```Go
for _, ptr := range slicePointers {
   info := generateSliceInfo(ptr, inQuery)
   allIter.sInfos = append(allIter.sInfos, info)
}
```

sqlair.go

internal/expr/query.go
Aflynn50: Added.

internal/expr/query.go
Aflynn50: Thats a good point I've changed it to your suggestion! When I was writing this I was really wishing Go had tuples, it feels like they would be very natural here, but I guess this way we are forced to name everything which may ultimately be a good thing.

internal/expr/query.go
Aflynn50: Not yet, I anticipate a messy rebase :)

internal/expr/query.go
Aflynn50: Good point, I've moved it

internal/expr/query.go
Aflynn50: I think we could, but since the whole function fits on one page as it is this would probably end up adding unnecessary overhead. It would certainly be worth considering if it has to grow a lot when maps are added here.

internal/expr/query.go
```diff
@@ -89,7 +89,87 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
-// ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
+// sliceInfo contains information about a column slice passed to All.
```
niemeyer: I may be off somehow, but I was expecting the implementation to be much simpler.

Can't we have something along those lines (scratch code):

```
func (q *Query) All(sliceArgs []any) error {
	var sliceValues []reflect.Value
	for i, sliceArg := range sliceArgs {
		silceValuePtr := reflect.ValueOf(sliceArg)
		if sliceValuePtr.Kind() != reflect.Pointer { /* error */ }
		sliceValue := sliceValuePtr.Elem()
		if sliceValue.Kind() != reflect.Slice { /* error */ }
		sliceValues = append(sliceValues, sliceValue)
	}

	iter := q.Iter()
	for iter.Next() {
		var outputArgs []any
		for sliceValue := range sliceValues {
			outputArg := reflect.New(sliceValue.Elem().Type()).Elem()
			outputArgs := append(outputArgs, outputArg.Interface())
		}
		​​if !iter.Decode(outputArgs) {
			break
		}
		for i, outputArg := range outputArgs {
			sliceValues[i] = reflect.Append(sliceValues[i], reflect.ValueOf(outputArg))
		}
	}
	return iter.Close()
}

**Review by niemeyer — CHANGES_REQUESTED** [review]

File: internal/expr/query.go (hunk 8 above)
Comment (Aflynn50) [reply internal/expr/query.go]: We certainly could do it like that I believe. The main reasons for having the bulk of the code in the `query.go` and using the `AllInfo` are:
- We avoid the unnecessary type checks at the start of `ScanArgs` by not going through `Decode`
- The AllInfo struct is useful for saving info about the slices types that we then don't need to then generate again
- We keep the `All` in `sqlair.go` short and keep the use of `reflect` confined to `expr`

Ultimately the current implementation pretty much does the same things in the same order as above but tries to avoid lots of extra reflect.ValueOf/type checking.

Comment (niemeyer) [reply internal/expr/query.go]: It's not worth buying into logic that is significantly more complex before we can even measure the actual impact of the two implementations, and once we do get to the point of optimizing, we should focus on optimizing Iterator itself instead of All.

Comment (Aflynn50) [reply internal/expr/query.go]: Ok, I'll have a go at implementing as above.

Comment (Aflynn50) [reply internal/expr/query.go]: Changes made!

File: sqlair.go [hunk 9]
```diff
@@ -163,3 +164,61 @@ func (q *Query) One(outputArgs ...any) error {
 	iter.Decode(outputArgs...)
 	return iter.Close()
 }
+
+// All iterates over the query and decodes all the rows.
+// resultColumns must contain pointers to slices of each of the query arguments e.g.
+// > q.All(&[]*Person{}, &[]*Address{}) or q.All(&[]Address{}, &[]Person{})
```
niemeyer: This is slightly confusing because these values are being immediately discarded, which is something we do for real in the case of Prepare, but here it would make no sense as the whole point is retrieving those results. Maybe:
```
// All iterates over the query and decodes all rows into the provided slices.
//
// For example:
//
//     var pslice []Person
//     var aslice []*Address{}
//     err := query.All(&pslice, &aslice)
// 
func ...
```

We should really take []Person here rather than []*Person, otherwise we'll be allocating a slice value that points to individually allocated Person values, instead of having one slice of contiguous values which is significantly more efficient.

sqlair.go
```diff
@@ -163,3 +164,61 @@ func (q *Query) One(outputArgs ...any) error {
...
+// It fabricates all structs as needed.
+func (q *Query) All(slicePointers ...any) error {
```
niemeyer: I'd stick to the eariler theme and call them `sliceArgs` (we don't call them `outputPointers`, or `scanPointers`).

sqlair.go
```diff
@@ -163,3 +164,61 @@ func (q *Query) One(outputArgs ...any) error {
...
+		return q.err
+	}
+
+	// Check slice inputs
+	var slicePtrVals = []reflect.Value{}
+	var sliceVals = []reflect.Value{}
+	for _, ptr := range slicePointers {
+		ptrVal := reflect.ValueOf(ptr)
+		if ptrVal.Kind() != reflect.Pointer {
+			return fmt.Errorf("need pointer to slice, got %s", ptrVal.Kind())
```
niemeyer, line 193: I'm not sure about how we phrase this in Decode, but maybe something like this would be helpful:

`cannot decode result: need ...`

Otherwise this error might be a bit mysterious depending on where it pops up.

Same for the similar cases below.

sqlair.go
```diff
@@ -163,3 +164,61 @@ func (q *Query) One(outputArgs ...any) error {
...
+		sliceVals = append(sliceVals, sliceVal)
+	}
+
+	iter := q.Iter()
+	for iter.Next() {
+		var outputArgs = []any{}
+		for _, sliceVal := range sliceVals {
+			elemType := sliceVal.Type().Elem()
+			if elemType.Kind() != reflect.Pointer {
+				return fmt.Errorf("need pointer to slice of pointers, got pointer to slice of %s", elemType.Kind())
```
niemeyer: This is a bit unclear, but more fundamentally, should this even be an error?  If we get `[]People`, that's not a slice of pointers, and we do want to support it as it's significantly more efficient.

niemeyer (changes requested):
Thanks!  Last pass I think:

sqlair.go
Aflynn50: Thats a good point, should we allow both `[]*Person` and `[]Person` or just `Person[]`? I feel like it might be best to give the user the option, but steer them towards `[]Person` in the example.

Aflynn50: I'll assume from the latter comments that we want to allow both.

Aflynn50: Done.

sqlair.go

sqlair.go
Aflynn50: Added `cannot populate slice:`.

sqlair.go

sqlair.go
```diff
@@ -163,3 +164,80 @@ func (q *Query) One(outputArgs ...any) error {
...
 }
+
+// All iterates over the query and decodes all rows into the provided slices.
+//
+// For example:
+//
+//	var pslice []Person
+//	var aslice []*Address
+//	err := query.All(&pslice, &aslice)
+func (q *Query) All(sliceArgs ...any) (err error) {
```
niemeyer, line 177: Can we please have the additional `//` between the commented and uncommented code?  It looks significantly more clear for cases like this.

sqlair.go
```diff
@@ -163,3 +164,80 @@ func (q *Query) One(outputArgs ...any) error {
...
+		}
+		if !iter.Decode(outputArgs...) {
+			break
+		}
+		for i, outputArg := range outputArgs {
+			switch sliceVals[i].Type().Elem().Kind() {
+			case reflect.Pointer:
+				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg))
+			case reflect.Struct:
+				sliceVals[i] = reflect.Append(sliceVals[i], reflect.ValueOf(outputArg).Elem())
```
niemeyer, line 231: We need a `default:` here similar to the case above, but prefixed with `internal error:` as this should never happen if the prior logic is correct. This prevents awkward bugs passing through as successes and being harder to diagnose.

niemeyer (changes requested):
This looks good, with only one relevant issue below, and one nitpick:

sqlair.go
Aflynn50: `go fmt` removes it!
I'd like to put one there but I cant see a elegant way to do it apart from adding back in a comment below, which I have just done.

sqlair.go
Aflynn50: Changed.

sqlair.go
niemeyer: Ouch, thanks.

sqlair.go
```diff
@@ -163,3 +164,85 @@ func (q *Query) One(outputArgs ...any) error {
...
+			default:
+				iter.Close()
+				return fmt.Errorf("internal error: output arg has unexpected kind %s", sliceVals[i].Type().Elem().Kind())
```
niemeyer: The parameter here is the argument being switched on.
```
switch kind := <long expression>; kind {
default:
    // ... use kind
}
```

niemeyer (approved):
Thanks!

PR #50, opened 2023-03-21
No rows error

Aflynn50:
This PR adds an error to the public API that indicates no rows were found in the query results. This is returned by `One` if the results contain no rows.


niemeyer (approved):
Thanks!

PR #51, opened 2023-03-21
Add Iterator to Query

Aflynn50:
This PR adds a new function `Query.Iter()` this returns an `Iterator` that can be used to iterate over the results. The iterator has the method `Next`, `Decode` and `Close`. The query is run on the database when `Query.Iter()` is called.

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -53,11 +53,19 @@ func (db *DB) Unwrap() *sql.DB {
 	return db.db
 }
 
-// Query holds a database query and is used to iterate over the results.
+// Query holds the results of a database query.
 type Query struct {
 	qe   *expr.QueryExpr
 	q    func() (*sql.Rows, error)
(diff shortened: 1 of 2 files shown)


sqlair.go
```diff
@@ -78,77 +86,78 @@ func (db *DB) QueryContext(ctx context.Context, s *Statement, inputArgs ...any)
...
+		return iter.err
 	}
 	return err
 }
+
+// One runs a query and decodes the first row into outputArgs.
+func (q *Query) One(outputArgs ...any) error {
+	iter := q.Iter()
+	if !iter.Next() {
+		return fmt.Errorf("cannot return one row: no results")
```
niemeyer, line 159: This needs to be ErrNoRows, but a nice fix for a follow up. Super nice PR in general, thanks!

niemeyer (approved):

PR #52, opened 2023-03-28
Fix typo

Aflynn50:
Change annoymous to anonymous in test.


niemeyer (approved):
Let's please enable the tests to run automatically here.  I hadn't realized we didn't have that hooked in.

PR #54, opened 2023-03-29
Add Outcome

Aflynn50:
This PR adds a new struct, Outcome that can be used to access the outcome of the query (number of rows affected and similar things). The struct needs to be passed as the first of the `outputArgs` to `Get` and it will be populated with any information available once the query is run.

It also merges `One` and `Run` into `Get` and renames:
`All` -> `GetAll`
`Iter.Decode` -> `Iter.Get`

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -8,12 +8,16 @@ import (
 	"strings"
 )
 
-func (ce *QueryExpr) QuerySQL() string {
-	return ce.sql
+func (qe *QueryExpr) QuerySQL() string {
+	return qe.sql
 }
 
-func (ce *QueryExpr) QueryArgs() []any {
-	return ce.args
+func (qe *QueryExpr) QueryArgs() []any {
+	return qe.args
+}
+
+func (qe *QueryExpr) HasOutputs() bool {
+	return len(qe.outputs) > 0
 }
 
 type QueryExpr struct {
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -1,6 +1,7 @@
 package expr
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"reflect"
@@ -16,6 +17,10 @@ func (qe *QueryExpr) QueryArgs() []any {
 	return qe.args
 }
 
+func (qe *QueryExpr) QueryContext() context.Context {
+	return qe.ctx
+}
+
 func (qe *QueryExpr) HasOutputs() bool {
 	return len(qe.outputs) > 0
 }
@@ -24,6 +29,8 @@ type QueryExpr struct {
 	sql     string
 	args    []any
 	outputs []field
+
+	ctx context.Context
 }
 
 // Query returns a query expression ready for execution, using the provided values to
@@ -33,7 +40,7 @@ type QueryExpr struct {
 //	type Person struct {
 //	        Name string `db:"fullname"`
 //	}
-func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
+func (pe *PreparedExpr) Query(ctx context.Context, args ...any) (ce *QueryExpr, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("invalid input parameter: %s", err)
@@ -90,7 +97,7 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 		qargs = append(qargs, named)
 	}
 
-	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
+	return &QueryExpr{ctx: ctx, outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
 // ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -1,7 +1,6 @@
 package expr
 
 import (
-	"context"
 	"database/sql"
 	"fmt"
 	"reflect"
@@ -17,16 +16,10 @@ func (qe *QueryExpr) QueryArgs() []any {
 	return qe.args
 }
 
-func (qe *QueryExpr) QueryContext() context.Context {
-	return qe.ctx
-}
-
 type QueryExpr struct {
 	sql     string
 	args    []any
 	outputs []field
-
-	ctx context.Context
 }
 
 // Query returns a query expression ready for execution, using the provided values to
@@ -36,7 +29,7 @@ type QueryExpr struct {
 //	type Person struct {
 //	        Name string `db:"fullname"`
 //	}
-func (pe *PreparedExpr) Query(ctx context.Context, args ...any) (ce *QueryExpr, err error) {
+func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("invalid input parameter: %s", err)
@@ -93,7 +86,7 @@ func (pe *PreparedExpr) Query(ctx context.Context, args ...any) (ce *QueryExpr, 
 		qargs = append(qargs, named)
 	}
 
-	return &QueryExpr{ctx: ctx, outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
+	return &QueryExpr{outputs: pe.outputs, sql: pe.sql, args: qargs}, nil
 }
 
 // ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -152,7 +152,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
 		field := qe.outputs[idx]
 		outputVal, ok := typeDest[field.structType]
 		if !ok {
-			return nil, fmt.Errorf("type %q found in query but not passed to decode", field.structType.Name())
+			return nil, fmt.Errorf("type %q found in query but not passed to get", field.structType.Name())
 		}
 
 		val := outputVal.FieldByIndex(field.index)
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -21,10 +21,6 @@ func (qe *QueryExpr) QueryContext() context.Context {
 	return qe.ctx
 }
 
-func (qe *QueryExpr) HasOutputs() bool {
-	return len(qe.outputs) > 0
-}
-
 type QueryExpr struct {
 	sql     string
 	args    []any
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -16,6 +16,10 @@ func (qe *QueryExpr) QueryArgs() []any {
 	return qe.args
 }
 
+func (qe *QueryExpr) HasOutputs() bool {
+	return len(qe.outputs) > 0
+}
+
 type QueryExpr struct {
 	sql     string
 	args    []any
--- a/package_test.go
+++ b/package_test.go
@@ -583,9 +583,7 @@ func (s *PackageSuite) TestAllErrors(c *C) {
 
 func (s *PackageSuite) TestRunAndOutcome(c *C) {
 	dropTables, sqldb, err := personAndAddressDB()
-	if err != nil {
-		c.Fatal(err)
-	}
+	c.Assert(err, IsNil)
 
 	var jim = Person{
 		ID:         70,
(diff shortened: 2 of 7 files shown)


sqlair.go
```diff
@@ -51,17 +51,19 @@ func NewDB(db *sql.DB) *DB {
...
-	qe   *expr.QueryExpr
-	q    func() (*sql.Rows, error)
-	rows *sql.Rows
-	err  error
+	qe      *expr.QueryExpr
+	outcome *Outcome
+	closed  bool
+
+	q func() (*sql.Rows, error)
+	e func() (sql.Result, error)
```
niemeyer: That's the opposite of what we discussed recently. We should remove the closure altogether, rather than adding one more.

Also, how can a query be closed when it was never executed?  The Close method only exists on the Iterator.

How can it have an outcome?  Similar rationale..

Something seems quite off here, or I misunderstand is the idea.

niemeyer (changes requested):

sqlair.go
Aflynn50: After trying out the alternatives I stuck with the closures because it seemed like a really nice way to make the Query object agnostic to whether it was running on a transaction or straight on the database. I didn't find any other options that felt as good but if there are any better ways that would be good.

The close `bool` is to stop people accidentally running the query twice. Once the underlying `Rows` are closed then there is no way to open then again and if we were to allow them to run `One` again this would have to run the query again, which might not be expected. Once a method has been run on the query it gets closed and nothing else can be run on it.

It seems I may have misunderstood exactly what you were hoping for with the `Outcome`.

package_test.go
```diff
@@ -595,10 +593,21 @@ func (s *PackageSuite) TestRun(c *C) {
 
 	db := sqlair.NewDB(sqldb)
 
+	var outcome = sqlair.Outcome{}
+
 	insertStmt := sqlair.MustPrepare("INSERT INTO person VALUES ( $Person.name, $Person.id, $Person.address_id, 'jimmy@email.com');", Person{})
-	err = db.Query(nil, insertStmt, &jim).Run()
+	err = db.Query(nil, insertStmt, &outcome, &jim).Run()
```
niemeyer: This is mixing an output parameter together with the input parameters, so feels like a slightly odd place to hang it. Also, this makes it tempting to store the value in the Query, as is being currently done in the PR, but that doesn't seem entirely right either. The easiest way to see it is imaginging the same query executed multiple times, each with a different outcome. It'd be odd to have that same value reused over and over. Instead, I think we want to pair this with `One` and `Decode`, which are the places where the output parameters are provided, and these are already independent for each query.

niemeyer (changes requested):

internal/expr/query.go
```diff
@@ -148,7 +152,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) ([]any, error)
 		field := qe.outputs[idx]
 		outputVal, ok := typeDest[field.structType]
 		if !ok {
-			return nil, fmt.Errorf("type %q found in query but not passed to decode", field.structType.Name())
+			return nil, fmt.Errorf("type %q found in query but not passed to get", field.structType.Name())
```
niemeyer, line 155: This sentence is now a bit awkward... how about:
```
type %q found in query but missing from Get call
```
I actually wonder if this should really be an error, but let's keep it for now.

package_test.go
```diff
@@ -601,20 +599,82 @@ func (s *PackageSuite) TestRun(c *C) {
...
+		ID:         70,
+		Fullname:   "Jim",
+		PostalCode: 500,
+	}
+
+	db := sqlair.NewDB(sqldb)
+
+	var outcome = sqlair.Outcome{}
+
+	insertStmt := sqlair.MustPrepare("INSERT INTO person VALUES ( $Person.name, $Person.id, $Person.address_id, 'jimmy@email.com');", Person{})
```
niemeyer: Is the space after `(` still required here?  If so, we need this fixed, otherwise, let's please drop it from the query.

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
 	}
 
 	qe, err := s.pe.Query(inputArgs...)
-	return &Query{qs: db.db, qe: qe, err: err, ctx: ctx}
+	return &Query{qs: db.db, qe: qe, ctx: ctx, err: err}
 }
 
-// Run will execute the query.
-// Any rows returned by the query are ignored.
+// Run is an alias for Get that takes no arguments.
```
niemeyer, line 95:
```
// Run runs the query and ignores results, if any.
```

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+	defer iter.Close()
+	if outcome != nil {
+		err := iter.Get(outcome)
+		if err != nil {
+			return err
+		}
+	}
+	if !q.qe.HasOutputs() {
+		if len(outputArgs) > 0 {
+			return fmt.Errorf("cannot get results: query does not return any results")
```
niemeyer: Are we not handling this error inside the expression package already?  And if not, isn't it handled by the underlying logic handling Get already?

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
 }
 
-// Run will execute the query.
-// Any rows returned by the query are ignored.
+// Run is an alias for Get that takes no arguments.
 func (q *Query) Run() error {
+	return q.Get()
+}
+
+// Query.Get will run the query.
```
niemeyer: A bit more compact:
```
// Get runs the query and decodes the first result into the provided output arguments.
// It returns ErrNoRows if output arguments were provided but no results were found.
// An &Outcome{} variable may be provided as the first output variable.
```

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+//
+//	iter := q.Iter()
+//	err := iter.Get(&outcome)
+//	if iter.Next() {
+//		...
+//
 // outputArgs must contain all the structs mentioned in the query.
-// If an error occurs it will be returned with Iter.Close().
-func (iter *Iterator) Decode(outputArgs ...any) (err error) {
+// A pointer to an Outcome struct can be passed as the first argument which
```
niemeyer: This feels quite long in comparison to the first Get, when they are pretty much the same, and the reason for that is because it's trying to document way more than its own scope. Also, the last sentence is unfinished.

Making it similar to the first Get:
```
// Get decodes the result from the previous Next call into the provided output arguments.
// An &Outcome{} variable may be provided as the single output variable before
// the first call to Next.
```

sqlair.go
```diff
@@ -170,29 +231,27 @@ func (iter *Iterator) Close() error {
...
-	err := ErrNoRows
-	if iter.Next() {
-		err = iter.Decode(outputArgs...)
-	}
-	if cerr := iter.Close(); cerr != nil {
-		return cerr
-	}
-	return err
+// A pointer to an outcome can be passed as the first variadic argument to Run.
+// It will be populated with the outcome of the query execution.
```
niemeyer: Run doesn't take arguments anymore..
```
// Outcome holds metadata about executed queries, and can be provided as the
// first output argument to any of the Get methods.
```

sqlair.go
```diff
@@ -170,29 +231,27 @@ func (iter *Iterator) Close() error {
...
 // For example:
 //
 //	var pslice []Person
 //	var aslice []*Address
-//	err := query.All(&pslice, &aslice)
+//	err := query.GetAll(&pslice, &aslice)
 //
 // sliceArgs must contain pointers to slices of each of the output types.
-func (q *Query) All(sliceArgs ...any) (err error) {
+// A pointer to an Outcome struct can be passed as the first argument which will be populated with the outcome of the query.
```
niemeyer: Let's please reuse the &Outcome{} line suggested earlier.

sqlair.go
```diff
@@ -202,6 +261,16 @@ func (q *Query) All(sliceArgs ...any) (err error) {
...
 
+	var outcome *Outcome
+	if len(sliceArgs) > 0 {
+		if oc, ok := sliceArgs[0].(*Outcome); ok {
+			outcome = oc
+			sliceArgs = sliceArgs[1:]
+		}
+	}
+	if outcome != nil {
+		outcome.result = nil
```
niemeyer: The data flow here feels a bit confusing. E.g.:
```
if len(sliceArgs) > 0 {
        if outcome, ok := sliceArgs[0].(*Outcome); ok {
                outcome.result = nil
                sliceArgs = sliceArgs[1:]
        }
}
```

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+		cols, err = rows.Columns()
+		if err != nil {
+			return &Iterator{err: err}
+		}
+	} else {
+		result, err = q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
+		if err != nil {
+			return &Iterator{err: err}
+		}
 	}
```
niemeyer, line 155: Alternative shorter spelling:
```
(...)
if q.qe.HasOutputs() {
        rows, err = q.qs.QueryContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
        if err == nil {
                cols, err = rows.Columns()
        }
} else {
        result, err = q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
}
return &Iterator{qe: q.qe, rows: rows, cols: cols, err: err, result: result}
```

sqlair.go
```diff
@@ -87,64 +89,122 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+		}
+		return iter.Close()
+	}
+	if !iter.Next() {
+		if err := iter.Close(); err != nil {
+			return err
+		}
+		return ErrNoRows
+	}
+	return iter.Get(outputArgs...)
```
niemeyer: This function as a whole feels somewhat heavy for being the simpler version for a one shot operation. It may be time to decouple it from the iterator and abstract away any functionality that would be duplicated, but I assume most of it has already been abstracted out.

If you'd prefer to do this later, we could at least rephrase the function to streamline it a bit:
```
iter := q.Iter()
if outcome != nil {
        err = iter.Next(outcome)
}
if err == nil && !iter.Next() {
        err = ErrNoRows
}
if err == nil {
        err = iter.Get(outpuArgs...)
}
if cerr := iter.Close(); err = nil || err == ErrNoRows {
        err = cerr
}
return err
```

Note that this version accounts for the potential error on the final iter.Close, which is being ignored at the moment.

niemeyer (changes requested):
Thanks, Alastair. Some further feedback:

internal/expr/query.go
Aflynn50: True, in previous versions I had implemented it as such, where you could pass a subset of the output structs rather than all of them. Though the way that `Get` works now makes it a little harder.

We pass the addresses of the fields of the structs directly to the `Rows.Scan` function which _does_ require values for all the columns that appear in the query.

To allow subsets of the outputs it would require adding in dummy values to pass to scan, or creating new values of each type so that we could scan into these and then move them to the output structs when needed.

package_test.go
Aflynn50: Its not required anymore :) fixed.

sqlair.go
Aflynn50: Changed.

sqlair.go

sqlair.go
Aflynn50: This is the case where they have run a DML statement like "INSERT" or "UPDATE" and they have asked for outputs. We don't actually call `Get` (apart from before `Next` to get the `Outcome`) because `Next` returns false. So there is no case where `Get` is called with these `outputArgs`.

This is a mistake from the user where their query contains no output expressions but they have passed some structs to the results into.

On further consideration I think this error should be moved to before the query is actually executed, since if the error is being triggered the user has misunderstood something.

sqlair.go

sqlair.go

sqlair.go

sqlair.go

sqlair.go
Aflynn50: Changed.

sqlair.go
Aflynn50: In this version we would return `ErrNoRows` when they run a query with no output expressions, I've edited it a bit to make it more concise though.

The reason it ends up being so complicated is because we're effectively trying to handle two quite different control flows at the same time, we have the case where there are no output arguments and the case where they want one row. Even by removing the iter and doing the operations directly we will still have this conflict.

sqlair.go
niemeyer: The question was actually whether this error would be picked up by the expression parser itself. It should be, I think: the expression says there are not outputs, and yet we've passed data types in that are missing from the expression.

Aflynn50: The error here is that the expression contains no output expressions. When parsing the expression that is fine, its only here when they try and give a struct to output into that they have made an error. There is nothing wrong with the query not having any output expressions. The error message is a bit unclear here, I'll change it.

niemeyer (approved):
Thank you!

PR #55, opened 2023-03-31
Remove QueryContext

Aflynn50:
This PR removes `DB.QueryContext` and adds `ctx` as the first argument to `Query`. If `ctx` is `nil` then `context.Background()` will be used.


niemeyer (approved):
:+1:

PR #56, opened 2023-03-31
Replace Unwrap with PlainDB

Aflynn50:
This PR replaces Unwrap with PlainDB as the accessor for the underlying database. It also updates the naming of the databases in `package_test.go`.


niemeyer (approved):
Let's see for how long it survives. :)

PR #57, opened 2023-03-31
Add Transactions

Aflynn50:
This PR adds support for DB transactions. A new TX struct is added which encapsulates a transaction with Commit and Roleback methods.

This PR includes #58.

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -204,7 +204,7 @@ func (s *PackageSuite) TestValidDecode(c *C) {
 
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	if err != nil {
 		c.Fatal(err)
 	}
@@ -301,7 +301,7 @@ func (s *PackageSuite) TestDecodeErrors(c *C) {
 			Commentf("\ntest %q failed:\ninput: %s\noutputs: %s", t.summary, t.query, t.outputs))
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	if err != nil {
 		c.Fatal(err)
 	}
@@ -357,7 +357,7 @@ func (s *PackageSuite) TestValidOne(c *C) {
 		}
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	if err != nil {
 		c.Fatal(err)
 	}
@@ -399,7 +399,7 @@ func (s *PackageSuite) TestOneErrors(c *C) {
 			Commentf("\ntest %q failed:\ninput: %s\noutputs: %s", t.summary, t.query, t.outputs))
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	if err != nil {
 		c.Fatal(err)
 	}
@@ -421,7 +421,7 @@ func (s *PackageSuite) TestErrNoRows(c *C) {
 		c.Errorf("test failed, error %q not the same as %q", err, sql.ErrNoRows)
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	if err != nil {
 		c.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


sqlair.go
```diff
@@ -246,3 +246,44 @@ func (q *Query) All(sliceArgs ...any) (err error) {
...
+
+type TX struct {
+	tx *sql.Tx
+}
+
+func (db *DB) NewTX(tx *sql.Tx) *TX {
+	return &TX{tx: tx}
+}
+
+func (db *DB) Begin(ctx context.Context, opts *sql.TxOptions) (*TX, error) {
```
niemeyer: This should be our own TXOptions type, not the underlying ones. Otherwise we're stuck with that forever.

package_test.go
```diff
@@ -581,6 +581,80 @@ func (s *PackageSuite) TestAllErrors(c *C) {
...
+	if err != nil {
+		c.Fatal(err)
+	}
+	c.Assert(derek, Equals, derekCheck)
+	err = tx.Commit()
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	_, err = db.Unwrap().Exec(dropTables)
```
niemeyer: We should now be using our own implementation instead of unwrapping it. Let's get used to our own API. :)

You'll probably start wishing for something like Query.Run(), which again is the sort of reason why we want to dogfood it.

package_test.go
```diff
@@ -581,6 +581,80 @@ func (s *PackageSuite) TestAllErrors(c *C) {
...
+
+	// Check derek isnt in db, insert derek; commit.
+	tx, err = db.Begin(nil, nil)
+	if err != nil {
+		c.Fatal(err)
+	}
+	var derekCheck = Person{}
+	err = tx.Query(nil, selectStmt, &derek).One(&derekCheck)
+	if !errors.Is(err, sqlair.ErrNoRows) {
+		c.Fatalf("test failed, got err %s, expected %s", err, sqlair.ErrNoRows)
```
niemeyer: No need to say "test failed" on test failures. Not sure if we have other similar cases?

In this particular case, though:
```
c.Assert(err, Equals, sql.ErrNoRows)
```

package_test.go
```diff
@@ -581,6 +581,80 @@ func (s *PackageSuite) TestAllErrors(c *C) {
 	}
 }
 
+func (s *PackageSuite) TestTransactions(c *C) {
+	dropTables, sqldb, err := personAndAddressDB()
+	if err != nil {
+		c.Fatal(err)
```
niemeyer: All of these `if err != nil .. Fatal` are better spelled as `c.Assert(err, IsNil)`.

package_test.go
```diff
@@ -581,6 +581,80 @@ func (s *PackageSuite) TestAllErrors(c *C) {
...
+	db := sqlair.NewDB(sqldb)
+	tx, err := db.Begin(nil, nil)
+	if err != nil {
+		c.Fatal(err)
+	}
+	// Insert derek then rollback.
+	q := tx.Query(nil, insertStmt, &derek)
+	// TODO: replace this with q.Run() or equivlent.
+	iter := q.Iter()
+	iter.Next()
```
niemeyer: That really doesn't read so nicely as a way to do insert. Can we please have `Run()`?

sqlair.go
```diff
@@ -246,3 +246,44 @@ func (q *Query) All(sliceArgs ...any) (err error) {
...
+	}
+
+	qe, err := s.pe.Query(inputArgs...)
+	q := func() (*sql.Rows, error) {
+		if err != nil {
+			return nil, err
+		}
+		return tx.tx.QueryContext(ctx, qe.QuerySQL(), qe.QueryArgs()...)
+	}
+	return &Query{qe: qe, q: q, err: err}
```
niemeyer: This logic can be abstracted out since it's the same between the Query and the Transaction. We're just doing QueryContext on something else, which can be easily abstracted out into a private interface.

Also, as we discussed before, we should be executing the query right here instead of delaying it further via a closure unnecessarily.

niemeyer (changes requested):
Thanks, here is a first pass:

package_test.go
Aflynn50: Ah yea, I had already replaced these with Query.Run in the other one, I was waiting until that was available first from the other PR.

sqlair.go

package_test.go

package_test.go

package_test.go
Aflynn50: Once Run is in I'll add to the tests in this PR.

sqlair.go
Aflynn50: Ah ok, we hadn't discussed that it was an option to execute it here. I was under the impression we were always going to wait to execute the query until Iter(). Which is why I believed that closures were the only sensible option.

I have no problem with doing this way so I'll change it.

package_test.go

niemeyer (approved):
Thanks for the changes. :+1:

PR #58, opened 2023-04-03
Add Run

Aflynn50:
This PR adds the method Run which can be used to run the query when we do not care about the results. This is useful for Inserts, Updates and other DML statements.

This PR contains PR https://github.com/canonical/sqlair/pull/56 and PR https://github.com/canonical/sqlair/pull/55 as well.

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -8,12 +8,16 @@ import (
 	"strings"
 )
 
-func (ce *QueryExpr) QuerySQL() string {
-	return ce.sql
+func (qe *QueryExpr) QuerySQL() string {
+	return qe.sql
 }
 
-func (ce *QueryExpr) QueryArgs() []any {
-	return ce.args
+func (qe *QueryExpr) QueryArgs() []any {
+	return qe.args
+}
+
+func (qe *QueryExpr) HasOutputs() bool {
+	return len(qe.outputs) > 0
 }
 
 type QueryExpr struct {
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -1,6 +1,7 @@
 package expr
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"reflect"
@@ -16,6 +17,10 @@ func (qe *QueryExpr) QueryArgs() []any {
 	return qe.args
 }
 
+func (qe *QueryExpr) QueryContext() context.Context {
+	return qe.ctx
+}
+
 func (qe *QueryExpr) HasOutputs() bool {
 	return len(qe.outputs) > 0
 }
@@ -24,6 +29,8 @@ type QueryExpr struct {
 	sql     string
 	args    []any
 	outputs []field
+
+	ctx context.Context
 }
 
 // Query returns a query expression ready for execution, using the provided values to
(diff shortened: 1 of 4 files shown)


internal/expr/query.go
```diff
@@ -1,25 +1,32 @@
...
-func (ce *QueryExpr) QuerySQL() string {
-	return ce.sql
+func (qe *QueryExpr) QuerySQL() string {
+	return qe.sql
 }
 
-func (ce *QueryExpr) QueryArgs() []any {
-	return ce.args
+func (qe *QueryExpr) QueryArgs() []any {
+	return qe.args
```
niemeyer, line 16: Thanks for fixing those.

internal/expr/query.go
```diff
@@ -1,25 +1,32 @@
...
+func (qe *QueryExpr) QueryContext() context.Context {
+	return qe.ctx
 }
 
 type QueryExpr struct {
 	sql     string
 	args    []any
 	outputs []field
+
+	ctx context.Context
```
niemeyer: This seems misplaced here, considering that nothing in this type, or even this package, require the context as it's all expression manipulation.

package_test.go
```diff
@@ -575,12 +575,121 @@ func (s *PackageSuite) TestAllErrors(c *C) {
 			Commentf("\ntest %q failed:\ninput: %s\nslices: %s", t.summary, t.query, t.slices))
 	}
 
-	_, err = db.PlainDB().Exec(dropTables)
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
```
niemeyer, line 578: This makes me think we'll probably want some sort of QueryString at some point, but let's wait until we have more data about this.

sqlair.go
```diff
@@ -156,6 +176,9 @@ func (iter *Iterator) Close() error {
 func (q *Query) One(outputArgs ...any) error {
 	iter := q.Iter()
 	if !iter.Next() {
+		if iter.err != nil {
```
niemeyer: We should be able to do this without breaking the Iter abstraction and depending on details of its implementation here, or calling Close twice:
```
err := ErrNoRows
iter := q.Iter()
if iter.Next() {
    iter.Decode(outputArgs...)
    err = nil
}
if cerr := iter.Close(); cerr != nil {
        return cerr
}
return err
```

niemeyer (changes requested):
Thanks, Alastair. Just a few trivials and should be ready to go.

package_test.go
Aflynn50: I agree, it would be good to be to do a DML in one command, we could even accept inputs and pass the argument structs to `Prepare` behind the scenes.

internal/expr/query.go
Aflynn50: I agree, I've changed this. I had put it here based on what was said in the meeting yesterday.

sqlair.go

package_test.go
niemeyer: Indeed, that's probably uncontroversial enough for us to support right away.

sqlair.go
```diff
@@ -56,12 +56,19 @@ func (db *DB) PlainDB() *sql.DB {
...
 // Query holds the results of a database query.
 type Query struct {
-	qe   *expr.QueryExpr
-	q    func() (*sql.Rows, error)
-	rows *sql.Rows
-	err  error
+	qe  *expr.QueryExpr
+	qs  querySubstrate
+	ctx context.Context
+	err error
```
niemeyer, line 71: These four fields make it look solid. :+1:

niemeyer (approved):

PR #59, opened 2023-04-03
Fix bug with One not checking for errors and others

Aflynn50:
I found a few bugs while testing. These were:
- If there is an error with the SQLair syntax in the query run with `One()` it will return ErrNoRows rather than that error, there was also no check for a query error at the start of `One`, this has been fixed by adding in an error check to the `q` function in `Query`.
- There was also a possible error with `Decode` not checking for `iter.rows == nil`. I've added a check for this and tests for the method order.

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -94,7 +94,7 @@ DROP TABLE address;
 	return dropTables, db, nil
 }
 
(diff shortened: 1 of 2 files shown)


niemeyer:
As we discussed, let's please drop the use of closures in the query. This bug will be fixed as a side effect.

PR #60, opened 2023-04-05
Change decode to return an error.

Aflynn50:
This PR now changes Decode to return an error instead of a bool so that issues can be dealt with straight away. It now doesn't close the rows since an error could be handled and iteration could continue.

Old description:
> This PR adds two lines in `Decode` to close the underlying `sql.Rows` if an error occurs in execution. This closes the underlying database connection.
> 
> If a call to `sql.Rows.Next` returns `false` the rows are automatically closed. Since we are stuck with this behaviour when making use of the underlying `sql.Rows` it makes sense that we are consistent with it between our `Iter.Next` and `Iter.Decode` which both return `bool`.
> 
> Moreover, if decode throws an error and the user _does not_ do `Iter.Close` then they will get obscure errors from subsequent actions due to the leaking of the underlying a database connection. In the interest of making the users life as pain-free as possible it makes sense to close this connection proactivally where possible.
> 
> It is very much good practice to do `Iter.Close` after `Next` or `Decode` has returned false not least to check for the error, regardless of this change.

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -180,6 +180,7 @@ func (s *PackageSuite) TestValidDecode(c *C) {
 		}
 
 		iter := db.Query(nil, stmt, t.inputs...).Iter()
+		defer iter.Close()
 		i := 0
 		for iter.Next() {
 			if i >= len(t.outputs) {
@@ -285,6 +286,7 @@ func (s *PackageSuite) TestDecodeErrors(c *C) {
 		}
 
 		iter := db.Query(nil, stmt, t.inputs...).Iter()
+		defer iter.Close()
 		i := 0
 		for iter.Next() {
 			if i >= len(t.outputs) {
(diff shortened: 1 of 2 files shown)


package_test.go
```diff
@@ -669,8 +673,9 @@ func (s *PackageSuite) TestQueryMultipleRuns(c *C) {
 		if i >= len(iterOutputs) {
 			c.Fatalf("expected %d rows, got more", len(iterOutputs))
 		}
-		if !iter.Decode(iterOutputs[i]) {
-			break
+		if err := iter.Decode(iterOutputs[i]); err != nil {
+			iter.Close()
+			c.Fatal(err)
```
niemeyer, line 680: The `iter.Close` here seems a bit arbitrary, in the sense that there's a different fatal call a couple of lines above and the iterator is not being closed there. I'd suggest instead doing a `defer iter.Close()` at the top.

Also, perhaps not for this PR, but all of these Fatal calls can be changed to:
```
err := .... do whatever ...
c.Assert(err, IsNil)
```

niemeyer (approved):
Half of the bottom comment is out of date as you fixed it just now. The other half still applies but can be done in the future.

PR #62, opened 2023-04-06
Add RunQuery

Aflynn50:
This PR adds a method to `DB` called `RunQuery`. This method is shorthand for
```Go
stmt, err := Prepare(query, args...)
if err != nil {
    ...
}
err := db.Query(ctx, stmt, args...).Run()
```
For example instead of the above, you would write:
```Go
err := db.RunQuery(ctx, query, args...)
```

This PR is waiting on #63

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -103,6 +103,20 @@ func (q *Query) Run() error {
 	return nil
 }
 
+func (db *DB) RunQuery(ctx context.Context, query string, inputArgs ...any) error {
+	// Deference input args to use as type samples in prepare
+	var prepArgs = []any{}
+	for _, inputArg := range inputArgs {
+		prepArgs = append(prepArgs, reflect.Indirect(reflect.ValueOf(inputArg)).Interface())
+	}
+
+	stmt, err := Prepare(query, prepArgs...)
+	if err != nil {
+		return err
+	}
+	return db.Query(ctx, stmt, inputArgs...).Run()
+}
+
 // Iter returns an Iterator to iterate through the results row by row.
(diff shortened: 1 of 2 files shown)


sqlair.go
```diff
@@ -103,6 +103,21 @@ func (q *Query) Run() error {
 	return nil
 }
 
+// RunQuery is shorthand for preparing a statement, building a query, then doing Query.Run.
+func (db *DB) RunQuery(ctx context.Context, query string, inputArgs ...any) error {
```
niemeyer, line 107: Per previous conversation, I believe this should be:
```
db.QueryString(ctx, str, inputArgs).Run()
```
Otherwise you'll have to replicate that method for every single alternative found inside the Query type:

- db RunQuery
- db.IterQuery
- db.AllQuery

etc

niemeyer (changes requested):

sqlair.go
Aflynn50: This issue with `Iter`, `All` and `One` is that they have output expressions as well as inputs, and `Prepare` needs all the types mentioned in the query, including the outputs.
If we have `QueryString` which returns a `Query` then we don't see the output types until they call `All`/`Iter`/`One`.

We could have them passed to `QueryString` next to the input arguments but this would be a little messy, you would pass input types only once but output types twice.

If we did have separate replicates of each method that would work, but I agree that this isnt ideal either. 

Since `QueryString` doesn't return an error it could be feasible to somehow delay the `Prepare` until we have all the types (i.e. they call `All`/`Iter`/`One`) but I'm not sure what the best mechanism to do this might be.

PR #66, opened 2023-04-19
Factor out QueryContext and ExecContext to seperate functions

Aflynn50:
This PR creates internal functions that are a part of `Query` to pass through to the `database/sql` functions.

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -96,11 +96,19 @@ func (q *Query) Run() error {
 	if q.err != nil {
 		return q.err
 	}
-	_, err := q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
+	_, err := q.exec()
 	if err != nil {
 		return err
 	}
 	return nil
+}
+
+func (q *Query) exec() (sql.Result, error) {
+	return q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
+}
+
+func (q *Query) query() (*sql.Rows, error) {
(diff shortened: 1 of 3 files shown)


sqlair.go
```diff
@@ -96,20 +96,28 @@ func (q *Query) Run() error {
...
 	}
 	return nil
 }
 
+func (q *Query) exec() (sql.Result, error) {
+	return q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
+}
+
+func (q *Query) query() (*sql.Rows, error) {
+	return q.qs.QueryContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
```
manadart: It is [generally discouraged](https://go.dev/blog/context-and-structs) to store a context as a struct member.

There is no guard against calling `Run` twice, which could result in unpredictable behaviour.

I would prefer for methods that materialise an action against the database, whether DML or SQL, to accept a context.

manadart: Another issue, is that if you pass a `context.WithTimeout` you're counting down starting at the instantiation rather than at the execution.

Aflynn50: Good point, I've added a PR (#67) to move the context to the `Get/Run/GetAll` stage.

niemeyer (changes requested):
This doesn't feel like an improvement. We have "query args", "scan args", "output args", etc, and in that same object we already have QueryArgs and ScanArgs. Renaming one of them to Args feels like just making things less clear. Then, moving these one-liners into a method by themselves when they're only used once again feels like making code less readable for no benefit.

PR #68, opened 2023-05-02
Remove NewTX

Aflynn50:
~~Remove the unnecessary dependency on the DB for a `NewTX` to allow from a transaction to be wrapped without the DB.~~

Remove `NewTX`.


niemeyer (approved):
Sounds like the wise choice for now, thanks.

PR #69, opened 2023-05-17
Add support for input expressions in INSERT statements

Aflynn50:
Add explicit support for input expressions in `INSERT` statements.
For example:
- `INSERT INTO table (*) VALUES ($Person.*)`
- `INSERT INTO table (col1, col2, col3) VALUES ($Person.*)`
- `INSERT INTO table (renamed_col1, renamed_col2) VALUES ($Person.col1, $Person.col2)`

The tests are updated to use this new syntax for creating test databases. The test `TestJujuStore` is removed as it was added as a proof of concept and is no longer needed. The test would need a significant rewrite to work with the new method for creating a test DB.


package_test.go
```diff
@@ -900,6 +900,84 @@ func (s *PackageSuite) TestTransactionErrors(c *C) {
...
+		inputs:   []any{},
+		outputs:  [][]any{{&Person{}}, {&Person{}}, {&Person{}}, {&Person{}}},
+		expected: [][]any{{&Person{20, "Mark", 1500}}, {&Person{40, "Mary", 3500}}, {&Person{35, "James", 4500}}},
+	}, {
+		summary:  "insert person",
+		query:    "INSERT INTO person (*) VALUES ($Person.*)",
+		types:    []any{Person{}},
+		inputs:   []any{Person{50, "Jim", 6000}},
+		outputs:  [][]any{},
+		expected: [][]any{},
```
SimonRichardson: SQLite as of 3.7.11 can also do [bulk inserts of rows](https://www.sqlite.org/lang_insert.html), we should check that this syntax is also available.

```sql
INSERT INTO person (name, age) VALUES ("fred", 12), ("matt", 32), ("alice", 42), ("tim", 14);
```

It might be worth figuring out what `$Person` would be in this case i.e. do we need to index them, or treat them like a stack?

```sql
INSERT INTO person (*) VALUES ($Person.*), ($Person.*), ($Person.*), ($Person.*);
```
or via the index syntax, which does allow for repetition (useful?). This could be backward compatible, with `$Person` being an alias to `$Person[0]`?

```sql
INSERT INTO person (*) VALUES ($Person[0].*), ($Person[1].*), ($Person[2].*), ($Person[0].*);
```

SimonRichardson: If we do want to support bulk row updates, then the alternative syntax for bulk updates should have a test.

```sql
INSERT INTO person
          SELECT 'fred' AS "name", 12 AS "age"
UNION ALL SELECT 'alice', 14
UNION ALL SELECT 'bob', 15;
```

Aflynn50: I have actually been thinking this would be a pretty cool feature. The idea I had was that you would write your query as normal:
```sql
INSERT INTO person (*) VALUES ($Person.*)
```
And then when you do the query you could either pass a single `Person` OR a slice of Person: `db.Query(ctx, stmt, &personSlice)`. This would then expand the insert behind the scenes to be a bulk insert for all the Person objects in the slice. How does this sounds as an interface?

I also was thinking it would be a seperate PR after this got in.

Aflynn50: Since we are going to do the database prepare behind the scenes it will be perfectly possible for one SQLair query to correspond to multiple prepared queries (with different numbers of bulk inserts).

SimonRichardson: > How does this sounds as an interface?

This would be nice, I'm just slightly concerned about it not being obvious. Everything in sqlair feels obvious, this feels a bit magical, but don't hold that against this. 

Maybe `$Person.**` or something similar, i.e. expand everything. Then you can test if you get a `**` then it should be a slice.

Aflynn50: Good point, this feature should be easy to find

internal/expr/parser.go
```diff
@@ -386,16 +386,15 @@ func (p *Parser) parseColumn() (fullName, bool, error) {
 	return fullName{}, false, nil
 }
 
-func (p *Parser) parseTarget() (fullName, bool, error) {
+func (p *Parser) parseOutputType() (fullName, bool, error) {
```
niemeyer: Does it make sense to rename the function above to the new convention as well?

internal/expr/parser.go
```diff
@@ -417,14 +416,23 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 
 // parseList takes a parsing function that returns a fullName and parses a
 // bracketed, comma seperated, list.
-func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+func (p *Parser) parseList(singletons bool, parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
```
niemeyer: Bools are typically not a great API for these cases. Just imagine reading some random code with
```
parseList(true, ...)
parseList(false, ...)
```
It's unclear what either mean.

But there's also a second issue here which is the meaning of the parameter itself. From our conversation, it looks like the intention here is to say "Hey, I want to parse a list, but actually don't parse a list." which is not nice. If it's not a list, the code shoud just not call a list parsing function.

There's also a sideways point coming out of this, which is about how we parse one item or many items in certain places. I suggest we enforce consistency for now, so either `bar AS &Foo.bar`, or `(bar) AS (&Foo.bar)`, but not mixed cases. I'd make that true even for `(bar) AS (&Foo.*)` for now at least.

niemeyer (changes requested):
Thanks, Alastair. Here is an early pass.

internal/expr/parser.go
Aflynn50: Changed to need either brackets on both sides or no brackets at all. When this PR goes out we will have to notify all current users of this breaking change as it may be unexpected.

internal/expr/parser.go
```diff
@@ -449,72 +447,96 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
...
+	if cols, colsBracketed, ok := p.parseColumns(); ok {
 		p.skipBlanks()
 		if p.skipString("AS") {
 			p.skipBlanks()
-			if targetTypes, ok, err := p.parseTargets(); err != nil {
+			if targetTypes, typesBracketed, ok, err := p.parseTargetTypes(); err != nil {
 				return nil, false, err
 			} else if ok {
+				if colsBracketed && !typesBracketed {
+					return nil, false, fmt.Errorf(`column %d: missing brackets around types after "AS"`, p.pos)
```
TheSignPainter98: Somehow the text after the `AS` doesn't feel like it's specifying types to me, would `...missing brackets around field specifier(s) after "AS"` work?

internal/expr/expr_test.go
```diff
@@ -259,11 +283,35 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+	`INSERT INTO person (name, id) VALUES (@sqlair_0, @sqlair_1)`,
+}, {
+	"multiple structs insert values",
+	"INSERT INTO person (name, postalcode) VALUES ($Person.name, $Address.id)",
+	"[Bypass[INSERT INTO person ] Input[[name postalcode] [Person.name Address.id]]]",
+	[]any{Person{}, Address{}},
+	`INSERT INTO person (name, postalcode) VALUES (@sqlair_0, @sqlair_1)`,
+}, {
+	"double insert values",
+	"INSERT INTO person (*) VALUES ($Person.*)ON CONFLICT DO NOTHING",
```
TheSignPainter98: Minor format note, should there be a space before uses of `ON` in this struct?

internal/expr/parser.go
```diff
@@ -449,72 +447,96 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
 
 		nextItem = p.skipByte(',')
 	}
-	return nil, false, fmt.Errorf("column %d: missing closing parentheses", parenPos)
+	return nil, false, fmt.Errorf("column %d: missing closing brackets", parenPos)
```
TheSignPainter98: This was actually already correct :) Apparently 'round brackets' would also work, but by default 'brackets' is synonymous with 'square brackets'

internal/expr/parser.go
```diff
@@ -449,72 +447,96 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
...
+	}
+
+	return nil, false, false, nil
 }
 
 // parseOutputExpression requires that the ampersand before the identifiers must
 // be followed by a name byte.
 func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
 	start := p.pos
-
```
TheSignPainter98: I think perhaps this whitespace should remain to match the whitespace above the `Case 2` comment. These both feel like headers to me

internal/expr/parser.go
```diff
@@ -523,26 +545,58 @@ func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
 			}
 		}
 	}
-
 	cp.restore()
 	return nil, false, nil
 }
 
+func (p *Parser) parseSourceType() (fullName, bool, error) {
```
TheSignPainter98: Does this mix terminology slightly? `parseSourceType` is used in `parseInputExpression`, would it be cleaner to call this `parseInputType` or something similar?

internal/expr/prepare.go
```diff
@@ -56,28 +56,139 @@ func starCount(fns []fullName) int {
...
-	info, ok := ti[p.sourceType.prefix]
-	if !ok {
-		ts := getKeys(ti)
-		if len(ts) == 0 {
-			return nil, fmt.Errorf(`type %q not passed as a parameter`, p.sourceType.prefix)
-		} else {
-			return nil, fmt.Errorf(`type %q not passed as a parameter, have: %s`, p.sourceType.prefix, strings.Join(ts, ", "))
+func prepareInput(ti typeNameToInfo, p *inputPart) ([]fullName, []typeMember, error) {
+	var inCols = make([]fullName, 0)
+	var typeMembers = make([]typeMember, 0)
```
TheSignPainter98: Just for consistency, could these use `:=` like the block below? Also, with these values declared before the helper closures, it looks like they should be used in them, but this doesn't appear to be the case. Could their declaration be moved later?

internal/expr/prepare.go
```diff
@@ -56,28 +56,139 @@ func starCount(fns []fullName) int {
...
+
+	numTypes := len(p.sourceTypes)
+	numColumns := len(p.targetColumns)
+	starTypes := starCount(p.sourceTypes)
+	starColumns := starCount(p.targetColumns)
```
TheSignPainter98: Would it be better to declare these closer to where they're first used?

internal/expr/prepare.go
```diff
@@ -56,28 +56,139 @@ func starCount(fns []fullName) int {
...
+
+	// Check target struct type and its tags are valid.
+	var info typeInfo
+	var ok bool
+	var err error
```
TheSignPainter98: I might be missing something, but I can't seem to see where these are used? All the places I see have them shadowed, so would it be better to remove these entirely? Otherwise, could they be declared closer to where they're used?

internal/expr/prepare.go
```diff
@@ -291,3 +418,32 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 
 	return &PreparedExpr{inputs: inputs, outputs: outputs, sql: sql.String()}, nil
 }
+
+// printCols prints a bracketed, comma seperated list of fullNames.
+func printCols(cs []fullName) string {
```
TheSignPainter98: To me, `print` in the name of this function implies that `fmt.Print` or similar is called. Perhaps this could be renamed to something like `formatColNames`?

internal/expr/prepare.go
```diff
@@ -291,3 +418,32 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+		if i < len(cs)-1 {
+			s.WriteString(", ")
+		}
+	}
+	s.WriteString(")")
+	return s.String()
+}
+
+// namedParams returns n incrementing parameters begining at start.
+func namedParams(start int, n int) string {
```
TheSignPainter98: This function feels very similar to `printCols`, should the name be consistent with the above?

package_test.go
```diff
@@ -123,7 +143,7 @@ func (s *PackageSuite) TestValidIterGet(c *C) {
 		types:    []any{Person{}},
 		inputs:   []any{},
 		outputs:  [][]any{{&Person{}}, {&Person{}}, {&Person{}}, {&Person{}}},
-		expected: [][]any{{&Person{30, "Fred", 1000}}, {&Person{20, "Mark", 1500}}, {&Person{40, "Mary", 3500}}, {&Person{35, "James", 4500}}},
+		expected: [][]any{{&Person{30, "Fred", 1000}}, {&Person{20, "Mark", 1500}}, {&Person{40, "Mary", 3500}}, {&Person{35, "Dave", 4500}}},
```
TheSignPainter98: I'm guessing the change from James to Dave is to make everyone have four-letter names :D

package_test.go
```diff
@@ -216,18 +236,16 @@ func (s *PackageSuite) TestValidIterGet(c *C) {
...
 	}}
 
 	tests = append(tests, testsWithShadowPerson...)
 
-	dropTables, sqldb, err := personAndAddressDB()
+	dropTables, db, err := personAndAddressDB()
```
TheSignPainter98: This is nicer, but do these naming changes belong in this PR?

TheSignPainter98 (commented):
Logic looks `ok`, just a few cosmetic things

marco6:
Few thoughts we already discussed on syntax 1, just to keep track. In case the struct has a database-generated ID, this API creates a problem.

Consider the struct used as an example:

```SQL
CREATE TABLE person(
	id INTEGER NOT NULL AUTO_INCREMENT,
	name TEXT NOT NULL,
	team TEXT NOT NULL
);
```

And the go representation:
```go
type Person struct {
	ID	int	`db:"id"`
	Name	string	`db:"name"`
	Team	string  `db:"team"`
}
```

Normally, you would insert into that table with something like:

```go
func newPerson(person Person) (id int) {
	res, _ := db.Exec("INSERT INTO person(name, team) VALUES (?, ?)", person.Name, person.Team);
	return res.LastInsertedId(); // Doesn't work for postgres BTW.
}
```

Now, replacing with the first syntax would look like:

```go
func newPerson(person Person) (id int) {
	res, _ := db.Exec("INSERT INTO person(*) VALUES (&Person.*)", person);
	return res.LastInsertedId();
}
```

Now this query seems right and innocent (and it is most likely consistend with other usages of the API). However, the query gets expanded, including the ID which is most likely `0` (or whatever client-generated - which can easily conflict in high-concurrency scenarios).

It seems this problem doesn't affect syntax 2 & 3, even if in those case you loose partially the niceness of not having to list all fields manually...

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: Hmm thats a fair point. Would target types or even just targets sound better to you? Thats a word we use elsewhere

internal/expr/expr_test.go
Aflynn50: This one is actually on purpose, just to check that it works without a space, I'll put it in the test name though.

internal/expr/parser.go
Aflynn50: Ah dammit, I'll change it back, the problem is I want to use the word "bracketed" as a var name, and thought that I should make it consistent across the codebase. I guess an error message can be different to the code though.

internal/expr/parser.go
Aflynn50: Thats what it was called before, but Gustavo suggested it should be brought in line with the input expression type, which has fields `sourceTypes` and `targetColumns`.

internal/expr/prepare.go
Aflynn50: I never know when to use `:=` vs `var`, it really seems to just be a matter of personal preference.

I've moved the `numTypes, ...` declarations down to just before where they are used. The reason those two are at the top is because they are the return values, so I'll just move them to be named return values.

internal/expr/prepare.go
Aflynn50: Moved.

internal/expr/prepare.go
Aflynn50: I've moved them to be closer to where they are used. Having them here makes the code shorter, but redeclaring them where they are used makes it more explicit.

internal/expr/prepare.go
Aflynn50: Good point, changed.

internal/expr/prepare.go
Aflynn50: I've changed the name to `generateNamedParams` and added an example in the comment to make it clear what it actually does.

package_test.go
Aflynn50: You would be correct.

package_test.go
Aflynn50: This was discussed with Gustavo in the last meeting, though they could well be put in another PR.

Aflynn50:
> Few thoughts we already discussed on syntax 1, just to keep track. In case the struct has a database-generated ID, this API creates a problem.
> 
> Consider the struct used as an example:
> 
> ```sql
> CREATE TABLE person(
> 	id INTEGER NOT NULL AUTO_INCREMENT,
> 	name TEXT NOT NULL,
> 	team TEXT NOT NULL
> );
> ```
> 
> And the go representation:
> 
> ```go
> type Person struct {
> 	ID	int	`db:"id"`
> 	Name	string	`db:"name"`
> 	Team	string  `db:"team"`
> }
> ```
> 
> Normally, you would insert into that table with something like:
> 
> ```go
> func newPerson(person Person) (id int) {
> 	res, _ := db.Exec("INSERT INTO person(name, team) VALUES (?, ?)", person.Name, person.Team);
> 	return res.LastInsertedId(); // Doesn't work for postgres BTW.
> }
> ```
> 
> Now, replacing with the first syntax would look like:
> 
> ```go
> func newPerson(person Person) (id int) {
> 	res, _ := db.Exec("INSERT INTO person(*) VALUES (&Person.*)", person);
> 	return res.LastInsertedId();
> }
> ```
> 
> Now this query seems right and innocent (and it is most likely consistend with other usages of the API). However, the query gets expanded, including the ID which is most likely `0` (or whatever client-generated - which can easily conflict in high-concurrency scenarios).
> 
> It seems this problem doesn't affect syntax 2 & 3, even if in those case you loose partially the niceness of not having to list all fields manually...

I'll work on adding the `omitempty` tag to this PR, I knew this was an issue but on reflection it is quite a big issue that probably warrants being address straight away rather than in a subsequent PR.

A field with the `omitempty` tag will be omitted from asterisk input expressions meaning queries like the one above will be safe.

internal/expr/parser.go
Aflynn50: True, may have even been removed by accident.

internal/expr/prepare.go
```diff
@@ -60,149 +60,185 @@ func starCount(fns []fullName) int {
...
+	}()
+
+	numTypes := len(p.sourceTypes)
+	numColumns := len(p.targetColumns)
+	starTypes := starCount(p.sourceTypes)
+
+	// Check for standalone input expression and prepare if found.
+	// For example:
+	//  "$P.name"
+	if numColumns == 0 {
```
letFunny: Can we add a helper to tell whether a given `inputPart` is part of an INSERT or not? That way we can instantly read the code without remembering that `numColumns == 0` implies INSERT

internal/expr/prepare.go
```diff
@@ -254,30 +290,35 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
-			inLoc, err := prepareInput(ti, p)
+			inCols, typeMembers, err := prepareInput(ti, p)
 			if err != nil {
 				return nil, err
 			}
-			sql.WriteString("@sqlair_" + strconv.Itoa(inCount))
-			inCount++
-			inputs = append(inputs, inLoc)
+
+			if len(p.targetColumns) == 0 {
```
letFunny: Ditto as per the comment above about creating a helper.

internal/expr/typeinfo.go
```diff
@@ -53,8 +53,15 @@ func (f structField) memberName() string {
...
 
+// typeInfo exposes useful information about types used in SQLair queries.
 type typeInfo interface {
 	typ() reflect.Type
+
+	// typeMember returns the type member associated with a given column name.
+	typeMember(string) (typeMember, error)
+
+	// getAllMembers returns all members a type associated with column names.
+	getAllMembers() ([]typeMember, error)
```
letFunny: As discussed offline, this can be left as is or refactored as a switch statement inside the `prepareColumnsAndTypes` function, both approaches look good.

letFunny (commented):
Really nice addition to the library. The code looks very good, just added a very few minor comments.

internal/expr/prepare.go
Aflynn50: I've added helper functions here for all the checks relating to standalone input expressions.

internal/expr/prepare.go
Aflynn50: Added.

internal/expr/prepare.go
```diff
@@ -60,149 +60,202 @@ func starCount(fns []fullName) int {
...
+	}
+	return outCols, typeMembers, err
 }
 
-type typeNameToInfo map[string]typeInfo
+// isStandaloneInput returns true if the input expression occurs on its own,
+// not inside an INSERT statement.
+// For example:
+//   "... WHERE x = $P.name"
+func isStandaloneInput(p *inputPart) bool {
```
letFunny: I think it will be clearer what the code does if we instead had a function called `isInsert` which does not require further explanation. Basically this function is defined as not an insert, I think it makes more sense the other way around.

internal/expr/typeinfo.go
Aflynn50: I think we'll keep it as it is here, I believe PR #103 which contains this should be in very soon

internal/expr/expr_test.go
```diff
@@ -506,6 +554,38 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 		query:       "SELECT * AS &.* FROM t",
 		prepareArgs: []any{struct{ f int }{f: 1}},
 		err:         `cannot prepare statement: cannot use anonymous struct`,
+	}, {
+		query:       "INSERT INTO person (*, postalcode) VALUES ($Person.name, $Address.id)",
+		prepareArgs: []any{Address{}, Person{}},
+		err:         "cannot prepare statement: input expression: invalid asterisk in columns: (*, postalcode) VALUES ($Person.name, $Address.id)",
```
letFunny, line 579: Should this be a parser error if it does not depend on the types but only on the syntax?

internal/expr/parser.go
```diff
@@ -203,6 +203,12 @@ loop:
 		}
 
 		p.pos++
+
+		// An opening parentheses might be the start of an expression.
+		if p.peekByte('(') {
+			break loop
+		}
```
letFunny, line 329: Why is the `peekByte` needed? naively it seems that the parsing of `(...) VALUES (...)` should be the same as for `(...) AS (...)` which does not require this change.

internal/expr/prepare.go
```diff
@@ -145,54 +123,127 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []fullName, typeMe
...
+			err = fmt.Errorf("input expression: %s: %s", err, p.raw)
+		}
+	}()
+
+	// Check for standalone input expression and prepare if found.
+	// For example:
+	//  "$P.name"
+	if !isInsert(p) {
+		if len(p.sourceTypes) > 1 {
+			return nil, nil, fmt.Errorf("internal error: cannot group standalone input expressions")
```
letFunny, line 210: I checked the coverage and this is never executed, either we are missing a test or this statement is unreachable.

internal/expr/prepare.go
```diff
@@ -145,54 +123,127 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []fullName, typeMe
...
+	// Case 3: Explicit columns and types.
+	// For example the expressions:
+	//  "(col1, col2) VALUES ($P.name, $A.id)"
+	//  "(col1, col2) AS (&P.name, &P.id)"
 	if numColumns == numTypes {
-		for i, c := range p.sourceColumns {
-			t := p.targetTypes[i]
-			if info, err = fetchInfo(t.prefix); err != nil {
+		for i, c := range columns {
+			t := types[i]
```
letFunny, line 168: [Nit] Can we rename this to `typeNames`, I was not expecting `types[i]` to resolve to a `fullName`.

internal/expr/prepare.go
```diff
@@ -145,54 +123,127 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []fullName, typeMe
...
+		err = fmt.Errorf("output expression: %s: %s", err, p.raw)
+	}
+	return outCols, typeMembers, err
 }
 
-type typeNameToInfo map[string]typeInfo
+// isInsert returns true if the input expression inside an INSERT statement.
+func isInsert(p *inputPart) bool {
+	return len(p.targetColumns) > 0
+}
```
letFunny: Have you thought about putting this as a method of `inputPart` ? To me, it is more clear where the information about the struct is all together in its methods.

letFunny (commented):
I did another full pass on the code and left some comments

internal/expr/expr_test.go
Aflynn50: This is a good point, this error could be moved there. The columns of an input or output expression can be either a list of types or an asterisk.

Now that I think about it it would be very nice to have some type level representation of the difference between the two rather than just counting the number of stars.

internal/expr/parser.go
Aflynn50: Its because its very common to do `INSERT INTO table (...) VALUES(...)` [e.g. in this sqlite tutorial](https://www.sqlitetutorial.net/sqlite-insert/). Advance looks for interesting chars that may *preceed* the expression but "(" is *part of* the expression so it needs to be handled separately.

It is also allowed to omit the space after the `table` so simply adding `"S"` to the list of interesting chars would not work.

internal/expr/prepare.go
Aflynn50: Ah, I haven't been in the habit of testing internal errors since they would require white box testing which I try to keep to a minimum. In my opinion it is fine to just verify these by inspection, since they are really just a kinder way to crash.

internal/expr/prepare.go
Aflynn50: Yea, will change that.

internal/expr/prepare.go
Aflynn50: Yea, I suppose it would go best there.

internal/expr/prepare.go
letFunny: Ah so it is just a safeguard for a statement we think is unreachable, then I agree

internal/expr/parser.go
letFunny: Makes sense, thanks

internal/expr/expr_test.go
Aflynn50: Upon further consideration this would require changes outside the scope of the PR. We can look into it.

letFunny (changes requested):
Let's put this PR on hold until we get the pipeline redesign merged.

PR #70, opened 2023-05-17
Add examples

Aflynn50:
Add an example program that creates a database then uses SQLair to do some simple queries on it.

It also contains two examples of using the `Outcome` struct.


example/example.go
```diff
@@ -0,0 +1,156 @@
...
+	err = db.Query(nil, create).Run()
+	if err != nil {
+		panic(err)
+	}
+
+	insertPerson := sqlair.MustPrepare("INSERT INTO person (name, id, team) VALUES ($Person.name, $Person.id, $Person.team);", Person{})
+
+	var al = Person{"Alasatir", 1, "engineering"}
+	var ed = Person{"Ed", 2, "engineering"}
+	var marco = Person{"Marco", 3, "consulting"}
```
marco6: I *really* think this should be fixed as Marco doesn't seem a name right for consulting lol

example/example.go
```diff
@@ -0,0 +1,156 @@
...
+		team text
+	)`)
+	err = db.Query(nil, create).Run()
+	if err != nil {
+		panic(err)
+	}
+
+	insertPerson := sqlair.MustPrepare("INSERT INTO person (name, id, team) VALUES ($Person.name, $Person.id, $Person.team);", Person{})
+
+	var al = Person{"Alasatir", 1, "engineering"}
```
marco6: Is your name right?

Aflynn50: lol good catch

example/example.go
Aflynn50: hmmm, maybe `mrManino` then?

Aflynn50: Fine, I'll change it to engineering

example_test.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	fmt.Printf("are in %s.\n", l1.Name)
+
+	// Example 3
+	// Cycle through employees until we find one in the Penthouse.
+
+	// A map with a key type of string is used to
+	// pass arguments that are not fields of structs.
+	// sqlair.M is of type map[string]any but if
+	// the map has a key type of string it can be used.
```
letFunny: I think what we want to say in this comment is that you need to create a named type like sqlair.M and that the only requisite is that the keys are strings. Is that the case?

example_test.go
```diff
@@ -0,0 +1,338 @@
...
+	)
+	err = db.Query(nil, drop).Run()
+	if err != nil {
+		panic(err)
+	}
+
+	// Output:
+	//Alastair is on the Engineering team.
+	//Alastair (Engineering), Ed (Engineering), Marco (Engineering), Pedro (Management), Serdar (Presentation Engineering), Lina (Presentation Engineering), are in The Basement.
+	//Gustavo from team Leadership is in The Penthouse.
```
letFunny: [Nitpick] I think that whitespace is trimmed from output comments, so we could make it more readable with some spacing in the beginning.

example_test.go
```diff
@@ -0,0 +1,338 @@
...
+		VALUES ($M.name, $M.id)
+	`, sqlair.M{})
+
+	// This will insert Fred with id 30 into the database.
+	err = db.Query(nil, stmt, m).Run()
+	if err != nil {
+		panic(err)
+	}
+
+	// Any named map with a key type of string can be used with SQLair.
```
letFunny: Ditto as above

example_test.go
```diff
@@ -0,0 +1,338 @@
...
+		VALUES ($Location.room_name, $Location.room_id)`,
+		Location{},
+	)
+
+	var l1 = Location{ID: 1, Name: "The Basement"}
+	var l2 = Location{ID: 2, Name: "Court"}
+	var l3 = Location{ID: 3, Name: "The Market"}
+	var l4 = Location{ID: 4, Name: "The Bar"}
+	var l5 = Location{ID: 5, Name: "The Penthouse"}
+	var locations = []Location{l1, l2, l3, l4, l5}
```
letFunny: [Nitpick] Prefer ` := ` initialization instead of `var`.

letFunny (commented):
Really like the examples here, it would have been very helpful to see them when starting to work on SQLair. I left some minor nitpicky comments about spacing and declarations, feel free to leave those out of this PR.

example_test.go
Aflynn50: Changed.

example_test.go
Aflynn50: Yea that is the case, changed the comment to clarify.

example_test.go
Aflynn50: Changed

example_test.go
Aflynn50: Clarified.

letFunny (approved):

niemeyer (changes requested):
Having examples is a great idea, but this needs a pass to make it a bit more friendly/idiomatic. There are a few issues such as lines being over-broken (one line split into too many almost empty lines), errors being ignored, over repetition of logic, functions/logic which are too long, etc.

I would suggest having less examples, simple and sweet, at least to start with.

example_test.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	// Select an employee from the database.
+	selectEmployee, err := sqlair.Prepare(`SELECT &Employee.* FROM employees`, Employee{})
+	if err != nil {
+		panic(err)
+	}
+	// Get fetches a single result from the database and scans it into the
+	// arguement. To fetch all results from the database use GetAll, and to
```
letFunny:
```suggestion
	// argument. To fetch all results from the database use GetAll, and to
```

PR #72, opened 2023-05-17
Rename Source and Target

Aflynn50:
This PR changes the name of the `source` and `target` fields of the `outputExpression` to `sourceColumns` and `targetTypes`.

This PR also changes the `source` in the `inputExpression` to `sourceType`.

PR #73, opened 2023-05-18
Mixed types output exprs

Aflynn50:
In output expressions, we currently only allow a single type after the `AS`. The type must be followed either by an asterisk _or_ by column names. A mixture of both is not allowed. For example, we allow: `* AS &P.*` or `* AS &P.name, &P.id`.

We could allow statements that do have multiple types and mix asterisks where it is unambiguous. 

This PR expands the syntax to allow multiple types after the `AS` as well as mixing of asterisk and non-asterisk fields. For example it would now allow `* AS (&Person.id, &Address.*, &Manager.name)`.

If a field of a struct is filled twice (i.e. in `* AS (&P.*, &P.name)`) an error will be returned.

This PR includes #72.


internal/expr/prepare.go
```diff
@@ -108,103 +109,95 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
...
 		}
+		return outCols, typeMembers, nil
 	}
 
-	// Case 2: None star target cases e.g. "...(&P.name, &P.id)".
-
-	// Case 2.1: Explicit columns e.g. "name_1 AS P.name".
-	if len(p.sourceColumns) > 0 {
+	// Case 2: Explicit columns, single asterisk type e.g. "(col1, t.col2) AS &P.*".
+	if t := p.targetTypes[0]; t.name == "*" {
```
niemeyer: This PR looks nice in general. There's only an issue in the way these branches are being organized which would be good to improve. The branches don't actually do the verification to ensure that what they are handling is indeed what they believe to be, because they sort of trust the error handling doing far from here to be correct. Except, well, it might be checking different things, now or tomorrow, and this woud make the logic wrong and bogus in awkward ways. The mistake is so easy to make that it looks like it's been made already, as the above line is claiming to be handling a case that it's not (it only looks for the asterisk on the first member).

As a suggestion, move the case verifiaction logic into the actual branches they're supposed to protect, and inline the error checking logic as well so that it's clear what's supported and what's not, next to the actual logic supporting it.

niemeyer (changes requested):

internal/expr/prepare.go
Aflynn50: I've moved the logic of `CheckValidOutputExpr` into `PrepareOutput` and fixed the bug with multiple asterisk types. It looks a lot cleaner now and reads easier

niemeyer (approved):
Great, thank you!

PR #75, opened 2023-05-30
Change mapKey and structField to be passed by reference

Aflynn50:
This PR changes the implementation of `typeMember` for `mapKey` and `structField`. It is now the pointer that implements the type rather than the struct itself. The structs are also passed by reference around the code rather than being copied.


niemeyer (approved):
Thanks!

PR #76, opened 2023-06-01
Add support for input slices

Aflynn50:
This PR adds support for slices in SQLair inputs.

For example, for a named slice type S:

`SELECT &P.* FROM p WHERE id IN ($S[:])`

Slices of any length can be passed to the query as an argument. All the slice will be included in the query. There are plans to extend this syntax to allow taking slices of slices from within the SQL query itself.

The length of the slice arguments that the user passes are not know when the SQLair statement is prepared. The implementation therefore moves the generation of the SQL from the `Prepare` phase to the `Query` phase.

Normally, statements prepared on the database are cached and reused, however, because the SQLair generated SQL of the slice statements depend on the arguments to the query, statements containing slices are not currently cached. They are generated per query and not saved. There are plans cache statements containing slices however it is still a work in progress.


internal/expr/parser.go
```diff
@@ -417,14 +424,23 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
...
+	if singletons {
+		// Case 1: Unbracketed singleton
+		if obj, ok, err := parseFn(p); err != nil {
+			return nil, false, err
+		} else if ok {
+			return []fullName{obj}, true, nil
+		}
+	}
+
+	//Case 2: Bracketed list
```
TheSignPainter98:
```suggestion
	// Case 2: Bracketed list
```
:)

internal/expr/parser.go
```diff
@@ -417,14 +424,23 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 
 // parseList takes a parsing function that returns a fullName and parses a
 // bracketed, comma seperated, list.
-func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+func (p *Parser) parseList(singletons bool, parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+	if singletons {
+		// Case 1: Unbracketed singleton
```
TheSignPainter98: Should this `Case 1` comment be on the same level as the `Case 2` one?

internal/expr/prepare.go
```diff
@@ -199,18 +234,96 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
...
 	for _, t := range p.target {
 		outCols = append(outCols, fullName{name: t.name})
 	}
-	return outCols, typeMembers, nil
+	return &outputChunk{outCols}, typeMembers, nil
 }
 
 type typeNameToInfo map[string]typeInfo
 
+type sqlChunk interface {
```
TheSignPainter98: I don't think the intuition with this interface is obvious, is it worth adding a few doc comments? Also, all of the implementors I see here do nothing, what is this used for?

Should the signature for `chunk` instead be this?
```go
func chunk(*info) (string)
```
where `info` is some struct which contains the relevant counters so the chunk implementation can be in charge of updating the necessary numbers.

Also, should `generateInputSQL`, `generateInSQL` and `generateOutputSQL` be refactored into these methods?

internal/expr/parts.go
```diff
@@ -42,6 +42,17 @@ func (p *inputPart) String() string {
...
+type inPart struct {
+	types []fullName
+	raw   string
+}
+
+func (p *inPart) String() string {
+	return fmt.Sprintf("In[%+v]", p.types)
+}
+
+func (p *inPart) part() {}
```
TheSignPainter98: This empty implementation looks suspicious, is this to make `inPart` implement some interface?

internal/expr/prepare.go
```diff
@@ -199,18 +234,96 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
...
+		case *mapKey:
+			length := 1
+			if tm.slice != nil {
+				log.Printf("length is %d", tm.slice.length)
+				length = tm.slice.length
+			}
+			for j := 0; j < length; j++ {
+				sql.WriteString("@sqlair_")
+				sql.WriteString(strconv.Itoa(inCount))
+				if i < len(typeMembers)-1 || j < length-1 {
```
TheSignPainter98: This guard seems to have two responsibilities. Perhaps it would be cleaner if there were `if` here which adds a comma between map elements (`j < length - 1`), then a separate `if` at the end of the loop which adds a comma between type members (`i < len(typeMembers)-1`) at the end of the loop. This would also remove the need to add a comma in the `*structField` case

internal/expr/prepare.go
```diff
@@ -253,47 +366,65 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+		case *outputChunk:
+			outCount, s = generateOutputSQL(outCount, chunk.outCols)
+		case *inputChunk:
+			inCount, s = generateInputSQL(inCount)
+		case *inChunk:
+			inCount, s = generateInSQL(inCount, chunk.typeMembers)
+		case *bypassChunk:
+			s = chunk.str
+		default:
+			panic("not valid type")
```
TheSignPainter98: Really pedantic thing, would `"invalid type"` read more naturally?

internal/expr/prepare.go
```diff
@@ -253,47 +366,65 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+	return &PreparedExpr{inputs: inputs, outputs: outputs, sqlChunks: sqlChunks}, nil
+}
+
+func (pe *PreparedExpr) sql() string {
+	var inCount int
+	var outCount int
+	var s string
+	var sql bytes.Buffer
+
+	for _, chunk := range pe.sqlChunks {
```
TheSignPainter98: As all these implement `sqlChunk`, it feels like this loop could be made more concise:
```go
for _, chunk := range pe.sqlChunks {
    sql.WriteString(chunk.chunk()) // parameters and return values would need to be reworked :)
}
```

internal/expr/query.go
```diff
@@ -3,12 +3,14 @@ package expr
...
 	"database/sql"
 	"fmt"
+	"log"
 	"reflect"
 	"strconv"
 	"strings"
 )
 
 func (qe *QueryExpr) QuerySQL() string {
+	log.Println(qe.sql)
```
TheSignPainter98: Is this a leftover from debugging?

internal/expr/query.go
```diff
@@ -87,17 +90,38 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 			val = v.FieldByIndex(tm.index)
-		case mapKey:
+			qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(argCount), val.Interface()))
+			argCount++
+		case *mapKey:
 			val = v.MapIndex(reflect.ValueOf(tm.name))
 			if val.Kind() == reflect.Invalid {
 				return nil, fmt.Errorf(`map %q does not contain key %q`, outerType.Name(), tm.name)
 			}
+			if val.Kind() == reflect.Interface {
```
TheSignPainter98: As these branches are mutually exclusive, would it be better to use a `switch` here instead?

internal/expr/query.go
```diff
@@ -87,17 +90,38 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+				val = val.Elem()
+			}
+			if val.Kind() == reflect.Slice {
```
TheSignPainter98: Could this be generalised to handle both `reflect.Slice` and `reflect.Array`?

internal/expr/typeinfo.go
```diff
@@ -13,12 +13,18 @@ type typeMember interface {
...
 
+type sliceInfo struct {
+	length int
+}
+
 type mapKey struct {
 	name    string
 	mapType reflect.Type
+	// sliceInfo is nil if the map value is not allowed to be a slice.
+	slice *sliceInfo
```
TheSignPainter98: The `nil` semantics not being reflected in the field name feels messy. Perhaps renaming this field to something like `allowedSliceInfo` would help?

internal/expr/prepare.go
```diff
@@ -77,33 +78,67 @@ func starCheckOutput(p *outputPart) error {
...
-		return f, nil
+		return &inputChunk{}, []typeMember{f}, nil
 	default:
-		return nil, fmt.Errorf(`internal error: unknown info type: %T`, info)
+		return nil, nil, fmt.Errorf(`internal error: unknown info type: %T`, info)
 	}
 }
 
+// prepareIn generates the values to fetch for sqlair input expressions following an IN statement.
+func prepareIn(ti typeNameToInfo, p *inPart) (*inChunk, []typeMember, error) {
```
TheSignPainter98: Now that there are `prepareIn`, `prepareInput` and `prepareOutput`, perhaps it would be cleaner if these were refactored into methods? Perhaps a common interface could then be used to simplify where these methods would be called

internal/expr/typeinfo.go
Aflynn50: Fair, changed

internal/expr/prepare.go
Aflynn50: True, it would be a good thing to do, though I might leave it for a future PR, theres probably going to be quite a few others that get in before this one that touch these methods so having them in a different place will make merging much more of a pain than it needs to be. But certainly something to do

internal/expr/query.go
Aflynn50: I was considering that, I'll have a go at changing it and see what happens

internal/expr/query.go
Aflynn50: They're not, the first one mutates `val`.

internal/expr/query.go
Aflynn50: Yep :)

internal/expr/prepare.go

internal/expr/prepare.go

internal/expr/parts.go
Aflynn50: Yea, it is. The part method exists just as a flag for the interface.

internal/expr/parser.go

internal/expr/parser.go

internal/expr/prepare.go
Aflynn50: Changed :)

internal/expr/prepare.go
Aflynn50: Changed, a lot shorter now.

letFunny (commented):
Looks good

internal/expr/expr_test.go
```diff
@@ -484,24 +503,24 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
+		err:         `cannot prepare statement: need struct, map or slice, got pointer to struct`,
 	}, {
 		query:       "SELECT * AS &Person.* FROM t",
 		prepareArgs: []any{map[string]any{}},
 		err:         `cannot prepare statement: cannot use anonymous map`,
 	}, {
 		query:       "SELECT * AS &Person.* FROM t",
 		prepareArgs: []any{nil},
-		err:         `cannot prepare statement: need struct or map, got nil`,
+		err:         `cannot prepare statement: need struct, map or slice, got nil`,
```
niemeyer: We're getting close to it being "need [every possible type] but got nil". :)  How about "need valid value, got nil" for these cases instead?

internal/expr/expr_test.go
```diff
@@ -510,6 +529,26 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 		query:       "SELECT &NoTags.* FROM t",
 		prepareArgs: []any{NoTags{}},
 		err:         `cannot prepare statement: output expression: no "db" tags found in struct "NoTags": &NoTags.*`,
+	}, {
+		query:       "SELECT street FROM t WHERE x = $Address.*",
+		prepareArgs: []any{Person{}, Manager{}, Address{}},
+		err:         `cannot prepare statement: input expression: star type not allowed in standalone input: $Address.*`,
```
niemeyer: "star type",  "standalone input", ... it feels like some of these concepts are not really part of the set of ideas we need to convey to get the system to work properly, and they are naturally somewhat alien to developers that are used to either Go or SQL, so the less we need to explain, the better.

How about replacing most (all?) of these with the more boring:

```
cannot prepare statement: "$Address.*" used in invalid context
```
?

internal/expr/expr_test.go
```diff
@@ -43,6 +43,10 @@ type IntMap map[string]int
 
 type StringMap map[string]string
 
+type S []any
+type T []int
+type U []string
```
niemeyer: These type names would benefit from some love. :)

internal/expr/parser.go
```diff
@@ -532,18 +539,43 @@ func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
...
+
+	cp.restore()
+	return nil, false, nil
+}
+
+func (p *Parser) parseSourceType() (fullName, bool, error) {
 	if p.skipByte('$') {
 		if fn, ok, err := p.parseGoFullName(); ok {
-			if fn.name == "*" {
-				return nil, false, fmt.Errorf(`asterisk not allowed in input expression "$%s"`, fn)
```
niemeyer: It's not clear that this is worth losing.  It doesn't seem to make much sense to have people saying "foo IN $Type.*", does it?  Try to picture what kind of type a developer would create that would hold homogeneous fields that were supposed to  indiscriminately be injected into SQL expressions as more fields are added?

internal/expr/parser.go
```diff
@@ -532,18 +539,43 @@ func (p *Parser) parseOutputExpression() (*outputPart, bool, error) {
...
+
+	if p.skipString("IN") {
+		p.skipBlanks()
+		if types, ok, err := p.parseList((*Parser).parseSourceType); err != nil {
 			return nil, false, err
+		} else if ok {
+			return &inPart{
+				types: types,
+				raw:   p.input[cp.pos:p.pos],
+			}, true, nil
```
niemeyer: Per previous reviews, let's please not have this pattern as a rule. This line is just:
```
return &inPart{types: types, raw: p.input[cp.pos:p.pos]}, true, nil
```
If that's not readable enough, then the better alternative is:
```
part := &inPart{types: types, raw: p.input[cp.pos:p.pos]}
return part, true, nil

Comment (niemeyer) [inline internal/expr/parser.go]: Per previous reviews, let's please not have this pattern as a rule. These four lines are just:
```
return &inPart{types: types, raw: p.input[cp.pos:p.pos]}, true, nil
```
If that's not readable enough, then the better alternative is:
```
part := &inPart{types: types, raw: p.input[cp.pos:p.pos]}
return part, true, nil

internal/expr/parts.go
```diff
@@ -42,6 +42,17 @@ func (p *inputPart) String() string {
 
 func (p *inputPart) part() {}
 
+type inPart struct {
+	types []fullName
```
niemeyer: How do we handle `foo IN ("one", $Foo.two)`, and why is `IN` special in that regard, when in general we're happy to inject the values blindly? Wouldn't it be simpler if we just let the value through?

I'm missing the design conversation for this feature. Do we have a spec for it?

internal/expr/prepare.go
```diff
@@ -77,14 +80,63 @@ func prepareInput(ti typeNameToInfo, p *inputPart) (tm typeMember, err error) {
 			return nil, fmt.Errorf(`type %q not passed as a parameter (have "%s")`, p.sourceType.prefix, strings.Join(ts, `", "`))
 		}
 	}
-
+	if _, ok = info.(*sliceInfo); ok {
+		return nil, fmt.Errorf(`cannot use slice type %q outside of IN clause`, info.typ().Name())
+	}
+	if p.sourceType.name == "*" {
+		return nil, fmt.Errorf(`star type not allowed in standalone input`)
```
niemeyer: Per earlier comment, what is a "standalone input"?

internal/expr/prepare.go
```diff
@@ -77,14 +80,63 @@ func prepareInput(ti typeNameToInfo, p *inputPart) (tm typeMember, err error) {
 			return nil, fmt.Errorf(`type %q not passed as a parameter (have "%s")`, p.sourceType.prefix, strings.Join(ts, `", "`))
 		}
 	}
-
+	if _, ok = info.(*sliceInfo); ok {
+		return nil, fmt.Errorf(`cannot use slice type %q outside of IN clause`, info.typ().Name())
```
niemeyer: This seems rather limiting. Per above note, I think we should first talk about what is the overall representation and model of slices, and then see if that model works well with IN and other cases where there are lists involved on the SQL side.

internal/expr/prepare.go
```diff
@@ -77,14 +80,63 @@ func prepareInput(ti typeNameToInfo, p *inputPart) (tm typeMember, err error) {
...
+				// "%s" is used instead of %q to correctly print double quotes within the joined string.
+				return nil, fmt.Errorf(`type %q not passed as a parameter (have "%s")`, t.prefix, strings.Join(ts, `", "`))
+			}
+		}
+		var tm typeMember
+		if t.name == "*" {
+			switch info := info.(type) {
+			case *structInfo, *mapInfo:
+				return nil, fmt.Errorf(`cannot use %s %q with asterisk in input expression`, info.typ().Kind(), t.prefix)
+			case *sliceInfo:
```
niemeyer: So we index the slice with `$Slice.*`?  What about specific positions?  `$Slice.3`, `$Slice.[3]`?  Maybe not for now, but do we have a clear path for when that's requested?

niemeyer (changes requested):
Thanks for the PR, Alastair. Clearly some interesting ideas, and below follow some preliminar comments, but I think the thing I'm missing the most is the design conversation for:

1) How are slices represented in SQLair - that's an important conversation to avoid bad design

2) How are IN expressions processed - similarly so

internal/expr/parser.go
letFunny: $Type.* is still disallowed (see [test](https://github.com/canonical/sqlair/pull/76/files#diff-060008e27564bf048cc01912a18a63f283931138073eb81ef7ad5f9c2141df6cR545-R547)), the difference now is that the logic has been moved from the parser itself to the prepare. Before we knew that a star was directly an error but now we need the type to see if it is a slice (valid) or a map/struct (not valid).

internal/expr/expr_test.go
Aflynn50: Changed the `nil` case to "need valid value, got nil". 

I've also changed the other errors where the user provides the wrong type (e.g. int) from:
`"need struct, map or slice, got int"`
to:
`"unsupported type: int"`

internal/expr/expr_test.go
Aflynn50: I agree that "standalone input" is not needed here. I would like to keep the "star" or "asterisk" though because here it is the offending part of the statement and I think its worth bringing the users attention to that. I've changed it to:
```
cannot prepare statement: input expression: asterisk used in invalid context: $Address.*
```
I've used "asterisk" instead of "star" to be more precise.

I would argue that the concept of the asterisk should be something that the developer is familiar with if they are choosing to use SQLair. It is one of the primary features of the library.

internal/expr/expr_test.go
Aflynn50: Changed the names to reflect the content of the slices.

internal/expr/parser.go

internal/expr/prepare.go
Aflynn50: Changed, see comment above.

internal/expr/parser.go

internal/expr/expr_test.go
```diff
@@ -622,46 +626,53 @@ func (s *ExprSuite) TestValidQuery(c *C) {
...
+	}, {
+		query:       "SELECT foo FROM t WHERE x = $StringMap.street, y = $Person.id",
+		prepareArgs: []any{Person{}, StringMap{}},
+		queryArgs:   []any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
+		queryValues: []any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
+	}, {
+		query:       "SELECT name FROM person WHERE id IN ($S.*)",
+		prepareArgs: []any{sqlair.S{}},
+		queryArgs:   []any{sqlair.S{1, 2, 3, 4, 5, 6}},
+		queryValues: []any{sql.Named("sqlair_0", 1), sql.Named("sqlair_1", 2), sql.Named("sqlair_2", 3), sql.Named("sqlair_3", 4), sql.Named("sqlair_4", 5), sql.Named("sqlair_5", 6), sql.Named("sqlair_6", nil), sql.Named("sqlair_7", nil)},
```
letFunny: Are the nulls here still valid? From what we discussed yesterday I thought the plan was to support exact lengths only and re-preparing to the new exact length if the supplied slice is longer than expected.

sqlair.go
```diff
@@ -279,6 +287,9 @@ func (q *Query) Iter() *Iterator {
 		sqlstmt = q.tx.sqltx.Stmt(q.sqlstmt)
 		close = sqlstmt.Close
 	}
+	if q.isTemp {
+		close = sqlstmt.Close
+	}
```
letFunny, line 287: Instead of doing this, what do you think of adding a finalizer to the temporary `sqlstmt` in Query(...) so that we do not need to pass the `isTemp` flag which only carries that bit of state.

I am mostly thinking about how to make the implementation as compatible with future changes to the cache as possible.

internal/expr/query.go
```diff
@@ -93,36 +94,40 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 				sv := v.Index(i)
 				qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(argCount), sv.Interface()))
 				argCount++
 			}
 			// Add NULL for extra slice arguments.
-			for i = i; i < maxSliceLen; i++ {
+			for i = i; i < defaultSliceLen; i++ {
 				qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(argCount), nil))
 				argCount++
 			}
```
letFunny: As stated above, I think we planned to remove this.

sqlair.go
```diff
@@ -165,17 +166,24 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
 		ctx = context.Background()
 	}
 
-	sqlstmt, err := db.prepareStmt(ctx, db.sqldb, s)
+	qe, tempStmt, err := s.pe.Query(inputArgs...)
```
letFunny: I am brainstorming about how to remove the extra return here and how to simply it a bit, especially thinking about later revamping the cache. What do you think of storing the array lengths in `Statement`? For now it will be the default length but in the future we can store the latest one in order to update the cache. That way below we can just check if the `inputArgs` lengths do not match and re-prepare and re-generate the SQL. The advantages are:
* Arguably it makes sense to store the length of the prepared statement in the struct that contains the rest of the information.
* We do not have to carry a temporal value in the loop inside `Query` or here.

internal/expr/parser.go
```diff
@@ -6,6 +6,79 @@ import (
...
+func (p *inputPart) String() string {
+	return fmt.Sprintf("Input[%+v]", p.sourceType)
+}
+
+func (p *inputPart) part() {}
+
+// outputPart represents a named target output variable in the SQL expression,
+// as well as the source table and column where it will be read from.
+type outputPart struct {
+	preparedColumns []columnName
```
letFunny: This might not be important but I would like to store this in a different place, rather than at the Parser output. Before thing change, the parts contained all the information about the parsed state and only that, not carrying extra state. Now, we need to fill the `preparedColumns` attribute before creating the SQL with no indication of it in the type system.

internal/expr/prepare.go
```diff
@@ -354,23 +294,58 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+					length = sliceLens[sliceCount]
+				}
+				for i := 0; i < length; i++ {
+					sql.WriteString("@sqlair_")
+					sql.WriteString(strconv.Itoa(inCount))
+					if i < length-1 {
+						sql.WriteString(", ")
+					}
+					inCount++
+					sliceCount++
```
letFunny: Should this be outside this inner for loop?

letFunny (commented):
I did a pass. There are a few important comments and the rest are mainly concerned with how to remove state and/or refactor the code in a way that makes it extensible in the future when we change the cache.

internal/expr/prepare.go
Aflynn50: Yes it should, I've moved it and added a test for it.

sqlair.go
Aflynn50: I think a finalizer would be a bit too heavyweight here. We know exactly when we can close the `sql.Stmt` (on `Iter.Close`) so we should just close it there.
If we have a LRU cache or some other way of keeping these `sql.Stmt` values then we should close them on eviction from the cache. If they are in the cache then the finalizer on them will never run. We can still have a finalizer on the `sqlair.Statement` because that is not in the cache itself and we can close *all* the `sql.Stmt` that correspond to the different versions of the statement.

internal/expr/expr_test.go
Aflynn50: Have removed them, the number of inputs for a slice in the query is now exactly the length of the argument slice.

internal/expr/query.go
Aflynn50: It is now removed :)

sqlair.go
Aflynn50: I've changed this to return the `sliceLens` directly here instead of the `tempStmt`. I've actually put them inside an object called `StmtCritieron` that could be a general purpose thing.

If we try and store the "latest" slice lengths in the query we can get into problems with concurrency. Its very hard to define the most recent one if two queries on a single statement and happening at the same time.

I've created a spec [here](https://docs.google.com/document/d/1cMaVo3NhKsVvsGj0IxpYaSsx-4pYtEY9rOAZK6UOuuo/edit) for how we could go about handling having multiple queries.

internal/expr/parser.go
Aflynn50: Thats very fair, I've changed it to be stored in the PreparedExpr itself.

Aflynn50:
#### Notes from in person discussion with @letFunny and @niemeyer 
Before the meeting on 03/11/23 this PR only allowed slices in `IN` clauses in the SQL statement. A SQLair `IN` clause could contain slices, structs (though not with an asterisk) and maps. The syntax for slice inputs was `$S.*` where `S` is a slice type. The slices had a maximum length (set at 8). If the user passed a slice with a shorter length then the inputs to the `IN` clause would be padded out to 8 with `NULL` values. 

This approach had a number of drawbacks:
- There was no way to mix literals and SQLair inputs in the `IN` clause. (e.g. `IN ($S.*, "mystring")`) 
- There is no obvious way to extend the syntax to access slices of the slice (e.g. `S[:10]` in regular Go)
- Slices could not be used elsewhere in the query due to the padding with `NULL` values giving unpredictable results
- The slices had a fixed length

A new approach was proposed in the meeting:
- Slices are written as `S[:]` in the query. This syntax will put the whole slice in the query.
- The SQL generated by SQLair has the exact number of elements in the slice (e.g. if the length of slice is 3 we generate 3 input placeholders in the query for it)

Aflynn50:
#### Update on PR state and notes from in person discussion with @letFunny and @niemeyer 
Before the meeting on 03/11/23 this PR only allowed slices in `IN` clauses in the SQL statement. A SQLair `IN` clause could contain slices, structs (though not with an asterisk) and maps. The syntax for slice inputs was `$S.*` where `S` is a slice type. The slices had a maximum length (set at 8). If the user passed a slice with a shorter length then the inputs to the `IN` clause would be padded out to 8 with `NULL` values. 

This approach had a number of drawbacks:
- There was no way to mix literals and SQLair inputs in the `IN` clause. (e.g. `IN ($S.*, "mystring")`) 
- There is no obvious way to extend the syntax to access slices of the slice (e.g. `S[:10]` in regular Go)
- Slices could not be used elsewhere in the query due to the padding with `NULL` values giving unpredictable results
- The slices had a fixed length

A new approach was proposed in the meeting:
- Slices are written as `S[:]` in the query. This syntax will put the whole slice in the query. To reference part of the slice the user can write `S[:10]`, `S[10:]` or `S[10:20]`
- The SQL generated by SQLair has the exact number of elements in the slice (e.g. if the length of slice is 3 we generate 3 input placeholders in the query for it). This has the disadvantage that we cannot cache the prepared query and reuse it with the current setup since the generate SQL depends on the argument slice length passed to `Query`. A PR will be brought in the future to address this.
- Slices are not limited to only `IN` clauses. They can be placed anywhere in the query.

As the PR stands at time of writing, the slices are no longer limited to being in the `IN` clause, the slices can be any length (though the statement prepared on the database is not cached) and the slice is no longer padded out to a fixed length with `NULL` values.

The new syntax `S[:]` is implemented (in a naive way) for inserting a full slice into a query.

Supporting taking slices of the slice within the query is a work in progress and may appear either in this PR or in a follow up PR.

internal/expr/prepare.go
Aflynn50: See update comment below.

internal/expr/prepare.go
Aflynn50: See update comment below

internal/expr/parts.go
Aflynn50: See update comment below

internal/expr/expr_test.go
```diff
@@ -604,6 +624,59 @@ func (s *ExprSuite) TestPrepareMapError(c *C) {
...
+		[]any{sql.Named("sqlair_0", 1), sql.Named("sqlair_1", 2), sql.Named("sqlair_2", 3)},
+	}, {
+		"many slices",
+		"SELECT * AS &Person.* FROM person WHERE id IN ($Person.id, $S[:], $Manager.id, $IntSlice[:], $StringSlice[:])",
+		"[Bypass[SELECT ] Output[[*] [Person.*]] Bypass[ FROM person WHERE id IN (] Input[Person.id] Bypass[, ] Input[S[:]] Bypass[, ] Input[Manager.id] Bypass[, ] Input[IntSlice[:]] Bypass[, ] Input[StringSlice[:]] Bypass[)]]",
+		[]any{sqlair.S{}, Person{}, Manager{}, IntSlice{}, StringSlice{}},
+		[]any{sqlair.S{1, 2, 3}, Person{ID: 1}, Manager{ID: 1}, IntSlice{4, 5, 6}, StringSlice{"7", "8", "9"}},
+		"SELECT address_id AS _sqlair_0, id AS _sqlair_1, name AS _sqlair_2 FROM person WHERE id IN (@sqlair_0, @sqlair_1, @sqlair_2, @sqlair_3, @sqlair_4, @sqlair_5, @sqlair_6, @sqlair_7, @sqlair_8, @sqlair_9, @sqlair_10)",
+		[]any{sql.Named("sqlair_0", 1), sql.Named("sqlair_1", 1), sql.Named("sqlair_2", 2), sql.Named("sqlair_3", 3), sql.Named("sqlair_4", 1), sql.Named("sqlair_5", 4), sql.Named("sqlair_6", 5), sql.Named("sqlair_7", 6), sql.Named("sqlair_8", 7), sql.Named("sqlair_9", 8), sql.Named("sqlair_10", 9)},
```
letFunny: Should the last three arguments, the ones coming from `StringSlice`, be strings? e.g. `sql.Named("sqlair_8", "7")`

internal/expr/expr_test.go
```diff
@@ -682,22 +755,22 @@ func (s *ExprSuite) TestQueryError(c *C) {
...
 		prepareArgs: []any{Address{}, Person{}},
 		queryArgs:   []any{nil, Person{Fullname: "Monty Bingles"}},
-		err:         "invalid input parameter: need struct or map, got nil",
+		err:         "invalid input parameter: need valid value, got nil",
 	}, {
 		query:       "SELECT street FROM t WHERE x = $Address.street, y = $Person.name",
 		prepareArgs: []any{Address{}, Person{}},
 		queryArgs:   []any{(*Person)(nil)},
-		err:         "invalid input parameter: need struct or map, got nil",
+		err:         "invalid input parameter: need valid value, got nil",
```
letFunny, line 769: [Nit] Can we make the format of both error messages similar, this one is:
```
need something, got something else
```
and the one below is:
```
error: got something else
```

internal/expr/expr_test.go
```diff
@@ -411,6 +414,15 @@ comment */
 string
 of three lines' AND id = $Person.*`,
 		err: `cannot parse expression: line 3, column 26: asterisk not allowed in input expression "$Person.*"`,
+	}, {
+		query: "SELECT &S[:] FROM t",
+		err:   `cannot parse expression: cannot use slice type "S" in output expression: &S[:]`,
```
letFunny: This is an idea, but do you think we could do without mentioning the type `S` in the message, I find the phrase "slice type" a little bit weird, what do you think of the following?
```
cannot parse expression: cannot use slice in output expression: &S[:]
```

internal/expr/prepare.go
```diff
@@ -211,11 +217,11 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 	// Generate and save reflection info.
 	for _, arg := range args {
 		if arg == nil {
-			return nil, fmt.Errorf("need struct or map, got nil")
+			return nil, fmt.Errorf("need valid value, got nil")
 		}
 		t := reflect.TypeOf(arg)
 		switch t.Kind() {
-		case reflect.Struct, reflect.Map:
+		case reflect.Struct, reflect.Map, reflect.Slice, reflect.Array:
```
letFunny: If we are planning to remove array in the future for the range syntax, should we do it know to later preserve backwards compatibility?

internal/expr/query.go
```diff
@@ -12,12 +12,21 @@ func (qe *QueryExpr) QueryArgs() []any {
 	return qe.args
 }
 
+func (qe *QueryExpr) SQL() string {
+	return qe.pe.sql(qe.sc)
+}
+
+func (qe *QueryExpr) IsTemp() bool {
+	return len(qe.sc.sliceLens) > 0
+}
```
letFunny: Can we add a comment here saying that this function basically checks if the statement can be re-prepared?

internal/expr/query.go
```diff
@@ -46,12 +55,14 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 		if v.Kind() == reflect.Invalid || (v.Kind() == reflect.Pointer && v.IsNil()) {
-			return nil, fmt.Errorf("need struct or map, got nil")
+			return nil, fmt.Errorf("need valid value, got nil")
 		}
 		v = reflect.Indirect(v)
 		t := v.Type()
-		if v.Kind() != reflect.Struct && v.Kind() != reflect.Map {
-			return nil, fmt.Errorf("need struct or map, got %s", t.Kind())
+		switch v.Kind() {
+		case reflect.Struct, reflect.Map, reflect.Slice, reflect.Array:
```
letFunny: Same comment for array as above.

internal/expr/prepare.go
```diff
@@ -270,23 +269,66 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
+type stmtCriterion struct {
+	sliceLens []int
+}
+
+func (pe *PreparedExpr) sql(sc *stmtCriterion) string {
+	var sql bytes.Buffer
+	var inCount int
+	var outCount int
+	var outputPartCount int
+	var sliceCount int
```
letFunny: I like this implementation but I can be hard to keep track of all the indexes, maybe it makes sense to couple this tighter, for example we can make `outputCols` a slice indexed by a pointer to a queryPart. If we used maps for all of them we can just check based on the *part, which is the thing we are iterating over.

internal/expr/query.go
```diff
@@ -81,15 +94,34 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+			}
+			sliceLens = append(sliceLens, sliceLen)
+			i := 0
+			for i = 0; i < v.Len(); i++ {
+				sv := v.Index(i)
+				qargs = append(qargs, sql.Named("sqlair_"+strconv.Itoa(argCount), sv.Interface()))
+				argCount++
+			}
+		default:
+			return nil, fmt.Errorf(`internal error: unknown type: %T`, tm)
```
letFunny, line 121: We are using `panic` for the internal error above, should we do the same here?

internal/expr/typeinfo.go
```diff
@@ -193,6 +227,8 @@ func generateTypeInfo(t reflect.Type) (typeInfo, error) {
 		info.tags = tags
 
 		return &info, nil
+	case reflect.Slice, reflect.Array:
```
letFunny: Same comment for array as above.

internal/expr/typeinfo.go
```diff
@@ -49,10 +49,22 @@ func (f *structField) outerType() reflect.Type {
...
+type sliceType struct {
+	sliceType reflect.Type
+}
+
+func (st *sliceType) outerType() reflect.Type {
+	return st.sliceType
+}
+
+func (st *sliceType) memberName() string {
+	return sliceExtention
```
letFunny, line 65: Should we panic here and in the one for map? Should this method be moved to be implemented only for structs and that way we can check in the code that we are in fact dealing with the correct thing?

package_test.go
```diff
@@ -671,6 +695,13 @@ func (s *PackageSuite) TestValidGetAll(c *C) {
 		inputs:   []any{},
 		slices:   []any{&[]sqlair.M{}, &[]CustomMap{}},
 		expected: []any{&[]sqlair.M{{"name": "Mark"}}, &[]CustomMap{{"id": int64(20)}}},
+	}, {
+		summary:  "simple in",
+		query:    "SELECT * AS &Person.* FROM person WHERE id IN ($S[:])",
+		types:    []any{Person{}, sqlair.S{}},
+		inputs:   []any{sqlair.S{20, 35, 36, 37, 38, 39, 40}},
+		slices:   []any{&[]*Person{}},
+		expected: []any{&[]*Person{&Person{20, "Mark", 1500}, &Person{40, "Mary", 3500}, &Person{35, "James", 4500}}},
```
letFunny: Is this test essentially duplicated?

sqlair.go
letFunny: Yeah you are right, we know exactly when to close the statements both with the current implementation and with the LRU cache so finalizers are not necessary.

But what happens if the user creates a `Query` object `q`, never calls `Get` and then `q` goes out of scope. The temporary statement would never be closed, right?

internal/expr/export_test.go
```diff
@@ -1,5 +1,6 @@
 package expr
 
+// Generate the SQL for a query with no slices in.
```
letFunny:
```suggestion
// Generate the SQL for a query with no slices.
```

letFunny (commented):
Code looking good, I especially like the changes to the error messages which make them a lot more clear. I also find the structure now to be a lot more readable and the state more clearly defined.

sqlair.go
letFunny: You are right about the concurrency, it makes more sense to later follow an approach like the one you described in the document.

internal/expr/expr_test.go
letFunny: Just checked this in the call, turns out you can use numbers to insert into string columns in SQLite (see https://www.db-fiddle.com/f/77WHfffDAwFnvxEGvM6GYm/0).

Aflynn50: Good catch! turns out I was missing the check that actually tested the `qe.QueryArgs()` against this expected value.

sqlair.go
Aflynn50: This is true. I would class this as misuse of the API, but as we have discussed in the meeting today, there are solutions that we can bring in the future as part of other changes that might fix this such as putting a finalizer on the `sql.Stmt` itself that closes it.

internal/expr/expr_test.go
Aflynn50: Sounds good, changed.

internal/expr/prepare.go
Aflynn50: Removed array support. If it is not going to work with the new slice of slice syntax we are going to introduce then its not worth having here. Also if you do have an array, it could be passed to the query by doing `s := append(sqlair.S{}, arr...)` and using `S` in the query.

internal/expr/query.go
Aflynn50: Yea, I'll add a comment explaining this.

internal/expr/query.go
Aflynn50: Removed.

internal/expr/query.go
Aflynn50: We only use panic in the `sql` method because it doesn't make sense for it to return an error (i.e. the only errors it could return are internal errors). Here we are already returning an error so we might as well make use of it.

internal/expr/typeinfo.go
Aflynn50: Removed.

internal/expr/typeinfo.go
Aflynn50: The `memberName` for maps is used, its just for slices that it is not.

Aflynn50: I'm reluctant to have a panic here. There is a place that `memberName()` is used in an error message, and even though I believe it is currently unreachable with slices, I would rather have a weird error message than a panic.

package_test.go
Aflynn50: Yea, I can removed it. I had added it because this is `GetAll` rather than `Get` and there have been some uncaught bugs in the past because I didn't test on `GetAll`. But those were with output types and since we are only dealing here with input types that should be ok.

internal/expr/export_test.go
Aflynn50: Changed.

internal/expr/prepare.go
Aflynn50: I've changed this though I'm in two minds about it. On one hand the map feels a bit like overkill for this, we can just count the position, its very much something we control, so why not just use it? On the other hand its a very easy place to introduce errors, I don't want to optimize something for the sake of it when it solves far more problems to just use the safe, more expensive option.

internal/expr/expr_test.go
Aflynn50: Changed.

package_test.go
letFunny: Keep it then

internal/expr/parser.go
```diff
@@ -24,7 +24,18 @@ type valueAccessor struct {
 	typeName, memberName string
 }
 
+// sliceExtention is the string following a type to indicate that it is a
+// slice. It indicates that all elements of the input slice are passed to the
+// query.
+// A more comprehensive slice syntax/semantics is planned. Matching against
```
jameinel, line 30: I think this doc string is incomplete now that you have `[:10]` etc syntax supported.

internal/expr/expr_test.go
```diff
@@ -604,6 +628,61 @@ func (s *ExprSuite) TestPrepareMapError(c *C) {
...
+		"single slice",
+		"SELECT name FROM person WHERE id IN ($S[:])",
+		"[Bypass[SELECT name FROM person WHERE id IN (] Input[S[:]] Bypass[)]]",
+		[]any{sqlair.S{}},
+		[]any{sqlair.S{1, 2, 3}},
+		"SELECT name FROM person WHERE id IN (@sqlair_0, @sqlair_1, @sqlair_2)",
+		[]any{sql.Named("sqlair_0", 1), sql.Named("sqlair_1", 2), sql.Named("sqlair_2", 3)},
+	}, {
+		"many slices",
+		"SELECT * AS &Person.* FROM person WHERE id IN ($Person.id, $S[:], $Manager.id, $IntSlice[:], $StringSlice[:])",
```
jameinel, line 650: since sqlair.S is a []any, it can't be passed an []int, or a []string, but it can be passed sqlair.S{1, "two", 3.0}, etc.
I didn't see tests for:
1) What error you get if you use $S[:] syntax, but then pass in an []int or []string. Ideally we would confirm that the error messages you get point users into understanding why []int isn't an []any
2) That you can represent mixed types (or it fails clearly) when passing a mixed type sqlair.S{1, "two"}

jameinel (commented):
a couple of small notes

Aflynn50:
This PR is being put on hold whilst we redesign the pipeline in the `expr` package. The current abstractions are not suited to the changes needed to add support for slices. This feature is still the priority, but may come in the form of multiple smaller PRs rather than this large one.

The problem with the current pipeline is that the SQL is geenrated at the Prepare stage meaning that it cannot depend on the input arguments. When a slice input is used in a query, the length of the input slice determines the number of argument placeholders in the generated SQL. The pipeline needs to be redesigned to allow for the SQL to depend on the input arguments.

PR #78, opened 2023-06-02
[JUJU-3907] Rename sqlair.TX and associations to use "Tx"

manadart:
Here we rename the sqlair type`TX` to `Tx`, along with associated methods/types. 

This is for:
- Idiomatic Go (abbreviation, not acronym).
- Congruence with `sql.Tx` from the standard library.

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -319,38 +319,38 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 	return nil
 }
 
-type TX struct {
+type Tx struct {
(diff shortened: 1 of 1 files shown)


niemeyer:
We've discussed this before, and the rationale behind the change is that both DB and TX have almost exactly the same API, so writing code for both made it obvious that there's an inconsistency, and we've decided to fix it on the SQLair layer.  Happy to have Db and Tx instead if that sounds better? ;-)

PR #79, opened 2023-06-05
Do a db/tx prepare on the creation of a query

Aflynn50:
This PR prepares the query on the database/transaction when the `Query` is created.


niemeyer (changes requested):
This is preparing the statement once per query, which isn't very useful. We want to cache the database-prepared statement with a key based on the sqlair statement and probably the database this is being sent to (not sure if the sql package offers any better key for this), so that this is reused across Query objects every time we see that sqlair.Statement with in an environment that would allow reuse.

PR #80, opened 2023-06-07
Set to zero on `NULL`

Aflynn50:
With `sql.Rows.Scan`, when you are trying to read a `NULL` from the query results into a Go type, you will get an error (unless the type has `nil`). This PR changes that behaviour for SQLair. Instead of an error, the value will be set to a zero on finding a Null in the database.

If the type is already a pointer or implements `sql.Scanner` it is passed straight to `Rows.Scan`. However, if it not, a pointer to it is passed to `Scan` and after scanning, if the pointer is `nil` then the value is zeroed, otherwise it is set.


package_test.go
```diff
@@ -375,6 +375,106 @@ func (s *PackageSuite) TestIterGetErrors(c *C) {
...
+
+	var tests = []struct {
+		summary  string
+		query    string
+		types    []any
+		inputs   []any
+		outputs  []any
+		expected []any
+	}{{
+		summary:  "nulls with omityempty",
```
cole-miller: typo here (extra `y` in `omitempty`), repeated a few times further down

Aflynn50: Fixed :))

internal/expr/query.go
```diff
@@ -100,6 +102,20 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+	Value reflect.Value
+	Valid bool
+}
+
+func (nt *NullT) Scan(src any) error {
+	if src == nil {
+		nt.Valid = false
+		return nil
+	}
+	nt.Valid = true
```
MathieuBordere: is it an issue that `nt.Valid` can be `true` when `ConvertAssign` returns an error?

Aflynn50: Fixed

internal/expr/query.go
```diff
@@ -100,6 +102,23 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+func (nt *NullT) Scan(src any) error {
+	if src == nil {
+		nt.Valid = false
+		return nil
+	}
+	err := convert.ConvertAssign(nt.Value.Addr().Interface(), src)
+	if err == nil {
+		nt.Valid = true
+	}
+	return err
```
MathieuBordere: you never set `nt.Valid` to `false` in case of `err`. I would probably write it like this

```
nt.Valid = false
if src == nil {
    return nil
}
err := convert.ConvertAssign(nt.Value.Addr().Interface(), src)
if err == nil {
    nt.Valid = true
}
return err
```

Aflynn50: Good point, I was relying on the fact that its initialised to false and will only be used once, but theres no reason not to do it that way. Changed.

MathieuBordere: yeah, but then setting it to `false` when `src == nil` was inconsistent. Anyway, just nitpicking :-)

niemeyer (changes requested):
As discussed, we probably don't want it done like this. There are three separate problems here:

1) When _sending_ a primary key to the database, we may want it to be omitted if unset so that the database can initialize it. That would work similarly to how `omitempty` generally works in encoders and decoders in Go, so it sounds okay.

2) When _reading_, which is what this PR seems to be talking about, this isn't `omitempty` and it's not clear that is needed. It sounds like the intent here is to allow people to unmarshal NULL values into non-pointers, and that's generally supported by encoders and decoders transparently. In other words, if an int is set to NULL, it's zeroed out. We shouldn't have to force people to spell it out so this works.

3) There's a third thing mentioned which is the existence of types such as `sql.NullString`. This is the opposite of the former case in a sense, and is used when the high-level user code _wants_ to know whether this was set to NULL or not. People shouold be able to use this when they want, and I believe we don't need to do anything extraordinary in SQLair for this to be supported since we're already leveraging database/sql infrastructure.

internal/expr/query.go
```diff
@@ -174,7 +181,18 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 			if !val.CanSet() {
 				return nil, nil, fmt.Errorf("internal error: cannot set field %s of struct %s", tm.name, tm.structType.Name())
 			}
-			ptrs = append(ptrs, val.Addr().Interface())
+			pt := reflect.PointerTo(val.Type())
+			if val.Type().Kind() != reflect.Pointer && !pt.Implements(scannerInterface) {
+				// Rows.Scan will return an error if it tries to scan NULL into a type that does not have nil.
+				// Pointers are the only nil type that work with Rows.Scan so for none pointer types we generate
+				// a pointer to them and pass this to Rows.Scan. After scanning the variable is set, if the
```
niemeyer: This sentence doesn't seem to be reading properly. Needs some rewording.

internal/expr/query.go
```diff
@@ -119,6 +121,11 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 		mapVal  reflect.Value
 	}
 	var mapSetInfos = []mapSetInfo{}
+	type scanInfo struct {
+		pointerScanVal reflect.Value
+		fieldVal       reflect.Value
+	}
+	var scanInfos = []scanInfo{}
```
niemeyer: These should be initialized as nil:
```
var mapSetInfos []mapSetInfo
var scanInfos []scanInfo
```
Also, the nomenclature here is a bit uneven. Both of these are doing exactly the same thing: having a proxy value in the scan phase, so it can be set into the final value on success. So maybe we could try to unify:
```
type scanProxy struct {
        scan reflect.Value
        real reflect.Value
        key  reflect.Value
}
var scanProxies []scanProxy
```
Then we can probably unify the logic for both. If not a map, we just don't set the key.

internal/expr/query.go
```diff
@@ -186,6 +204,13 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 		for _, mi := range mapSetInfos {
 			mi.mapVal.SetMapIndex(mi.keyVal, mi.scanVal)
 		}
+		for _, si := range scanInfos {
+			if !si.pointerScanVal.IsNil() {
+				si.fieldVal.Set(si.pointerScanVal.Elem())
+			} else {
+				si.fieldVal.Set(reflect.Zero(si.fieldVal.Type()))
```
niemeyer: This will change given the above, but FWIW the common logic may be pulled out of the branch and done after it (si.fieldVal.Set).

internal/expr/typeinfo.go
```diff
@@ -39,10 +39,6 @@ type structField struct {
 
 	// The tag assosiated with this field
 	tag string
-
-	// OmitEmpty is true when "omitempty" is
-	// a property of the field's "db" tag.
-	omitEmpty bool
```
niemeyer: Let's please keep this. See the note regarding the three cases mentioned in the prior review. This is one of them. It's just not what is being handled right now.

niemeyer (changes requested):
That's close, thanks!

internal/expr/query.go
Aflynn50: Logic unified.

internal/expr/query.go
Aflynn50: Changed as discussed in meeting.

internal/expr/query.go
Aflynn50: Done.

internal/expr/typeinfo.go
Aflynn50: Added back.

niemeyer (approved):
Thanks!  This feature is great to have.

PR #81, opened 2023-06-08
Prepare statements on DB when Query created

Aflynn50:
Prepare SQLair statements on the database when a new Query is created with DB.Query or TX.Query. DB prepared statements are cached for reuse. 

Each statement is assigned a unique ID which is used as the key to the cache of `*sql.Stmt`s. When the statement is added to the cache a [finalizer](https://pkg.go.dev/runtime#SetFinalizer) is set on the `sqlair.Statement` object which will remove it from the cache and and close the prepared statement when it is garbage collected. This does mean that any errors thrown when the prepared statement is closed are ignored.

A single `sqlair.Statement` can be prepared on multiple databases. So the cache is double indexed, first by the statement ID, then by the database ID (it is of type `map[stmtID]map[dbID]*sql.Stmt`). There is also a separate store indexed first by the database ID and then by the statement ID (`map[dbID]map[stmtID]bool`) that is used to record which statements are prepared on which DB so that they can be efficiently removed from the cache when the finalizer on the database is triggered.

Whilst testing for this PR a bug was discovered with the setup of the database test suite. The sqlite database was opened with `sql.Open("sqlite3", ":memory:")`, this in fact means that every new connection to the database creates a new, empty, DB. And since the Go database driver handles connections behind the scenes this lead to some unexpected bugs.

To fix this problem is now fixed but it turned out that the `DROP TABLE` statements had never been working as intended (this went unnoticed as a new connection/DB was created for each test before).

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -1,20 +1,20 @@
 package sqlair
 
-type TestIndex struct {
-	TX *TX
-	S  *Statement
+import (
+	"database/sql"
+	"sync"
+
+	"github.com/google/uuid"
+)
+
+func (s *Statement) ID() uuid.UUID {
+	return s.id
 }
 
-// CheckCacheEQ checks the order of the contents of the
-// private LRU cache that stores prepared statements.
-func (db *DB) CheckCacheEQ(tis []TestIndex) bool {
-	e := db.preparedCache.ll.Front()
-	for i := 0; i < db.preparedCache.ll.Len(); i++ {
-		k := e.Value.(*entry).key
-		if i >= len(tis) || tis[i].TX != k.tx || tis[i].S != k.s {
-			return false
-		}
-		e = e.Next()
-	}
-	return true
+func (db *DB) Cache() (map[uuid.UUID]*sql.Stmt, *sync.RWMutex) {
+	return db.stmtCache.c, &db.stmtCache.m
+}
+
+func (tx *TX) Cache() (map[uuid.UUID]*sql.Stmt, *sync.RWMutex) {
+	return tx.stmtCache.c, &tx.stmtCache.m
 }
--- a/export_test.go
+++ b/export_test.go
@@ -14,5 +14,5 @@ func (db *DB) CacheID() int64 {
 }
 
 func Cache() (map[int64]map[int64]*sql.Stmt, map[int64]map[int64]bool, *sync.RWMutex) {
-	return stmtDBCache, dbStmtCache, &cacheMutex
+	return stmtDBCache, dbStmts, &cacheMutex
 }
--- a/export_test.go
+++ b/export_test.go
@@ -1,0 +1,18 @@
+package sqlair
+
+type TestIndex struct {
+	TX *TX
+	S  *Statement
+}
+
+func (db *DB) CheckCacheEQ(tis []TestIndex) bool {
+	e := db.preparedCache.ll.Front()
+	for i := 0; i < db.preparedCache.ll.Len(); i++ {
+		k := e.Value.(*entry).key
+		if i >= len(tis) || tis[i].TX != k.tx || tis[i].S != k.s {
+			return false
+		}
+		e = e.Next()
+	}
+	return true
+}
--- a/export_test.go
+++ b/export_test.go
@@ -5,6 +5,8 @@ type TestIndex struct {
 	S  *Statement
 }
 
+// CheckCacheEQ checks the order of the contents of the
+// private LRU cache that stores prepared statements.
 func (db *DB) CheckCacheEQ(tis []TestIndex) bool {
 	e := db.preparedCache.ll.Front()
 	for i := 0; i < db.preparedCache.ll.Len(); i++ {
--- a/export_test.go
+++ b/export_test.go
@@ -9,14 +9,10 @@ func (s *Statement) CacheID() int64 {
 	return s.cacheID
 }
 
-func (tx *TX) CacheID() int64 {
-	return tx.cacheID
-}
-
 func (db *DB) CacheID() int64 {
 	return db.cacheID
 }
 
-func Cache() (map[int64]map[int64]*sql.Stmt, *sync.RWMutex) {
-	return stmtCache, &cacheMutex
+func Cache() (map[int64]map[int64]*sql.Stmt, map[int64]map[int64]bool, *sync.RWMutex) {
+	return stmtDBCache, dbStmtCache, &cacheMutex
 }
--- a/export_test.go
+++ b/export_test.go
@@ -3,18 +3,20 @@ package sqlair
 import (
 	"database/sql"
 	"sync"
-
-	"github.com/google/uuid"
 )
 
-func (s *Statement) ID() uuid.UUID {
-	return s.id
+func (s *Statement) CacheID() int64 {
+	return s.cacheID
 }
 
-func (db *DB) Cache() (map[uuid.UUID]*sql.Stmt, *sync.RWMutex) {
-	return db.stmtCache.c, &db.stmtCache.m
+func (tx *TX) CacheID() int64 {
+	return tx.cacheID
 }
 
-func (tx *TX) Cache() (map[uuid.UUID]*sql.Stmt, *sync.RWMutex) {
-	return tx.stmtCache.c, &tx.stmtCache.m
+func (db *DB) CacheID() int64 {
+	return db.cacheID
+}
+
+func Cache() (map[int64]map[int64]*sql.Stmt, *sync.RWMutex) {
+	return stmtCache, &cacheMutex
 }
--- a/export_test.go
+++ b/export_test.go
@@ -14,5 +14,5 @@ func (db *DB) CacheID() int64 {
 }
 
 func Cache() (map[int64]map[int64]*sql.Stmt, map[int64]map[int64]bool, *sync.RWMutex) {
-	return stmtDBCache, dbStmts, &cacheMutex
+	return stmtDBCache, dbStmtCache, &cacheMutex
 }
--- a/package_test.go
+++ b/package_test.go
@@ -912,7 +912,7 @@ func (s *PackageSuite) TestPreparedCaching(c *C) {
 	c.Assert(err, IsNil)
 	selectChurchRoad, err := sqlair.Prepare(`
 		SELECT &Address.*
-		FROM address 
+		FROM address
 		WHERE street = "Church Road"
 	`, Address{})
 	c.Assert(err, IsNil)
@@ -969,6 +969,8 @@ func (s *PackageSuite) TestPreparedCaching(c *C) {
 	err = tx.Query(nil, selectMark).Get(&p)
 	c.Assert(err, IsNil)
 	c.Assert(p, Equals, mark)
+
+	tx.Commit()
 
 	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
 	c.Assert(err, IsNil)
--- a/package_test.go
+++ b/package_test.go
@@ -910,7 +910,7 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 	db := sqlair.NewDB(sqldb)
 
 	mark := Person{20, "Mark", 1500}
-	churchRoad := Address{1500, "Sad World", "Church Road"}
+	mainStreet := Address{1000, "Happy Land", "Main Street"}
 	var p = Person{}
 	var a = Address{}
 
@@ -924,7 +924,6 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 	}
 
 	var markStmtID uuid.UUID
-	var churchRoadStmtID uuid.UUID
 	{
 		selectMark, err := sqlair.Prepare(`
 			SELECT &Person.*
@@ -939,30 +938,30 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 		c.Assert(err, IsNil)
 		c.Assert(p, Equals, mark)
 
+		// Check the prepared selectMark is in the cache.
 		c.Assert(lookup(db.Cache())(markStmtID), Equals, true)
-		// Check the prepared selectMark in the cache.
 
 		err = db.Query(nil, selectMark).Get(&p)
 		c.Assert(err, IsNil)
 		c.Assert(p, Equals, mark)
 
+		// Check selectMark is still in the cache.
 		c.Assert(lookup(db.Cache())(markStmtID), Equals, true)
-		// Check its still in the cache.
 
 		err = db.Query(nil, selectMark).Get(&p)
 		c.Assert(err, IsNil)
 		c.Assert(p, Equals, mark)
 	}
-	runtime.GC()
-	time.Sleep(1 * time.Millisecond)
 	// The finalizer for selectMark will run after the GC.
-	// 1 millisecond should be long enough for the finalizer to finish.
-
+	runtime.GC()
+	// Wait 1 millisecond for the finalizer to finish.
+	time.Sleep(1 * time.Millisecond)
 	// Check selectMark has been removed from the cache.
 	c.Assert(lookup(db.Cache())(markStmtID), Equals, false)
 
 	// Test with transactions.
 	var tx *sqlair.TX
+	var mainStreetStmtID uuid.UUID
 	{
 		selectMark, err := sqlair.Prepare(`
 			SELECT &Person.*
@@ -982,7 +981,7 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 
 		tx, err = db.Begin(nil, nil)
 		c.Assert(err, IsNil)
-		// Run selectMark on a tx. This should repreare the same
+		// Run selectMark on a Tx. This should reprepare the same
 		// *sql.Stmt on the Tx since it is already prepared on the DB.
 		err = tx.Query(nil, selectMark).Get(&p)
 		c.Assert(err, IsNil)
@@ -990,28 +989,28 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 
 		c.Assert(lookup(tx.Cache())(markStmtID), Equals, true)
 
-		// Check it is still in the cache.
+		// Check selectMark is still in the cache.
 		err = tx.Query(nil, selectMark).Get(&p)
 		c.Assert(err, IsNil)
 		c.Assert(p, Equals, mark)
 
 		c.Assert(lookup(tx.Cache())(markStmtID), Equals, true)
 
-		// Run a different query on the tx that has not
-		// already been prepared on the database.
-		selectChurchRoad, err := sqlair.Prepare(`
+		// Run a different query on the Tx that has not already been
+		// prepared on the DB.
+		selectMainStreet, err := sqlair.Prepare(`
 			SELECT &Address.*
 			FROM address
-			WHERE street = "Church Road"
+			WHERE street = "Main Street"
 		`, Address{})
 		c.Assert(err, IsNil)
 
-		churchRoadStmtID = selectChurchRoad.ID()
+		mainStreetStmtID = selectMainStreet.ID()
 
-		err = tx.Query(nil, selectChurchRoad).Get(&a)
+		err = tx.Query(nil, selectMainStreet).Get(&a)
 		c.Assert(err, IsNil)
-		c.Assert(a, Equals, churchRoad)
-		c.Assert(lookup(tx.Cache())(churchRoadStmtID), Equals, true)
+		c.Assert(a, Equals, mainStreet)
+		c.Assert(lookup(tx.Cache())(mainStreetStmtID), Equals, true)
 
 		tx.Commit()
 	}
@@ -1019,10 +1018,10 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 	runtime.GC()
 	time.Sleep(1 * time.Millisecond)
 
-	c.Assert(lookup(tx.Cache())(churchRoadStmtID), Equals, false)
+	c.Assert(lookup(tx.Cache())(mainStreetStmtID), Equals, false)
 	c.Assert(lookup(db.Cache())(markStmtID), Equals, false)
 
-	// Check cache is actually empty at the end
+	// Check the cache is empty at the end.
 	cache, _ := db.Cache()
 	c.Assert(cache, HasLen, 0)
 	cache, _ = tx.Cache()
--- a/package_test.go
+++ b/package_test.go
@@ -959,7 +959,6 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 	c.Assert(lookupStmt(dbID, markStmtID), Equals, false)
 
 	// Test with transactions.
-	var tx *sqlair.TX
 	var mainStreetStmtID int64
 	var txID int64
 	{
@@ -979,10 +978,10 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 
 		c.Assert(lookupStmt(dbID, markStmtID), Equals, true)
 
-		tx, err = db.Begin(nil, nil)
+		tx, err := db.Begin(nil, nil)
 		c.Assert(err, IsNil)
-
 		txID = tx.CacheID()
+
 		// Run selectMark on a Tx. This should reprepare the same
 		// *sql.Stmt on the Tx since it is already prepared on the DB.
 		err = tx.Query(nil, selectMark).Get(&p)
--- a/package_test.go
+++ b/package_test.go
@@ -1015,6 +1015,11 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 
 		tx.Commit()
 	}
+	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
+	c.Assert(err, IsNil)
+
+	err = db.Close()
+	c.Assert(err, IsNil)
 
 	runtime.GC()
 	time.Sleep(1 * time.Millisecond)
@@ -1029,8 +1034,32 @@ func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
 	// Check the cache is empty at the end.
 	c.Assert(cache[dbID], HasLen, 0)
 	c.Assert(cache[txID], HasLen, 0)
+}
+
+func (s *PackageSuite) TestDBClose(c *C) {
+	dropTables, sqldb, err := personAndAddressDB()
+	c.Assert(err, IsNil)
+	db := sqlair.NewDB(sqldb)
 
 	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
+	c.Assert(err, IsNil)
+
+	err = db.Close()
+	c.Assert(err, IsNil)
+
+	selectMark, err := sqlair.Prepare(`
+		SELECT &Person.*
+		FROM person
+		WHERE name = "Mark"
+	`, Person{})
+	c.Assert(err, IsNil)
+
+	var p = Person{}
+	err = db.Query(nil, selectMark).Get(&p)
+	c.Assert(err, ErrorMatches, "sql: database is closed")
+
+	//Check Close is idemepotent
+	err = db.Close()
 	c.Assert(err, IsNil)
 }
 
--- a/package_test.go
+++ b/package_test.go
@@ -468,6 +468,14 @@ func (s *PackageSuite) TestNulls(c *C) {
 	}
 
 	db := sqlair.NewDB(sqldb)
+	defer func() {
+		for _, dropTable := range dropTables {
+			err = db.Query(nil, sqlair.MustPrepare(dropTable)).Run()
+			if err != nil {
+				c.Fatal(err)
+			}
+		}
+	}()
 
 	insertNullPerson, err := sqlair.Prepare("INSERT INTO person VALUES ('Nully', NULL, NULL, NULL);")
 	c.Assert(err, IsNil)
@@ -490,11 +498,6 @@ func (s *PackageSuite) TestNulls(c *C) {
 			c.Assert(t.outputs[i], DeepEquals, s,
 				Commentf("\ntest %q failed:\ninput: %s", t.summary, t.query))
 		}
-	}
-
-	err = db.Query(nil, sqlair.MustPrepare(dropTables)).Run()
-	if err != nil {
-		c.Fatal(err)
 	}
 }
 
--- a/package_test.go
+++ b/package_test.go
@@ -949,10 +949,17 @@ func (s *PackageSuite) TestQueryMultipleRuns(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(iterOutputs, DeepEquals, iterExpected)
 
-	// Run them all again for good measure.
+	// Repeat in a transaction.
 	allOutput = &[]*Person{}
 	iterOutputs = []any{&Person{}, &Person{}, &Person{}, &Person{}}
 	oneOutput = &Person{}
+
+	tx, err := db.Begin(nil, nil)
+	c.Assert(err, IsNil)
+	q = tx.Query(nil, stmt)
+	err = q.Get(oneOutput)
+	c.Assert(err, IsNil)
+	c.Assert(oneExpected, DeepEquals, oneOutput)
 
 	err = q.GetAll(allOutput)
 	c.Assert(err, IsNil)
@@ -974,10 +981,12 @@ func (s *PackageSuite) TestQueryMultipleRuns(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(iterOutputs, DeepEquals, iterExpected)
 
-	q = db.Query(nil, stmt)
 	err = q.Get(oneOutput)
 	c.Assert(err, IsNil)
 	c.Assert(oneExpected, DeepEquals, oneOutput)
+
+	err = tx.Commit()
+	c.Assert(err, IsNil)
 }
 
 func (s *PackageSuite) TestTransactions(c *C) {
--- a/package_test.go
+++ b/package_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
 	"runtime"
 	"testing"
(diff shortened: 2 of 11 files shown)


package_test.go
```diff
@@ -894,12 +894,88 @@ func (s *PackageSuite) TestTransactionErrors(c *C) {
...
 	c.Assert(err, IsNil)
 	err = q.Run()
-	c.Assert(err, ErrorMatches, "sql: transaction has already been committed or rolled back")
+	c.Assert(err, ErrorMatches, "sql: statement is closed")
 
 	err = db.Query(ctx, sqlair.MustPrepare(dropTables)).Run()
 	c.Assert(err, IsNil)
 }
 
+func (s *PackageSuite) TestPreparedCaching(c *C) {
```
TheSignPainter98: Should this be `TestPrepareCaching` (no 'd')?

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
 	started bool
 }
 
+// A Least Recently Used Cache using a doubly linked list.
+type lruCache struct {
+	ll   *list.List
+	c    map[index]*list.Element
```
TheSignPainter98: I'm finding it quite difficult to follow the uses of these fields, could these names be a little longer?

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
+	key   index
+	value *sql.Stmt
+}
+
+func newLRUCache(size int) *lruCache {
+	c := &lruCache{}
+	c.ll = list.New()
+	c.c = make(map[index]*list.Element)
+	c.size = size
+	return c
```
TheSignPainter98: I think this function would feel more natural if it filled in the fields of the `lruCache` instance when it's created. I'm not sure how to name the syntax exactly, but I think perhaps something like this would look cleaner?
```go
func newLRUCache(size int) *lruCache {
    return &lruCache{
        ll: list.New(),
        c: make(map[index]*list.Element),
        size: size,
    }
}

File: sqlair.go [hunk 4]
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
+// A Least Recently Used Cache using a doubly linked list.
+type lruCache struct {
+	ll   *list.List
+	c    map[index]*list.Element
+	size int
+}
+
+// The cache key.
+type index struct {
+	tx *TX
```
TheSignPainter98: Storing `tx` directly like this in the cache would keep it (and hence the underlying database connection) alive. Although correct use of the API would require the user to remove `tx` from the cache, it may be too easy to forget to do this, causing their database connection to effectively leak.

Is there some tx ID or optional tag which could be used instead to break this dependence?

sqlair.go
```diff
@@ -55,31 +57,34 @@ func MustPrepare(query string, typeSamples ...any) *Statement {
 }
 
 type DB struct {
-	db *sql.DB
+	prepedCache      *lruCache
```
TheSignPainter98: s/prepedCache/preppedCache/g :P

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
 	started bool
 }
 
+// A Least Recently Used Cache using a doubly linked list.
+type lruCache struct {
```
TheSignPainter98: Unless there are going to be other types of cache you need to choose between later, I think it might be cleaner just to call this `cache`

sqlair.go
```diff
@@ -365,5 +456,28 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+
+	var ok bool
+	var ps *sql.Stmt
+	ps, ok = tx.db.prepedCache.lookup(s, tx)
+	if !ok {
+		ps, ok = tx.db.prepedCache.lookup(s, nil)
+		if !ok {
+			ps, err = tx.tx.PrepareContext(ctx, qe.QuerySQL())
+		} else {
+			// If the statement is alredy prepared on the db, prepare it on the tx.
```
TheSignPainter98: s/alredy/already/ :)

Also is there another way of saying 'prepare it on the tx' which doesn't use 'tx'? Otherwise it feels a little cryptic

sqlair.go
```diff
@@ -365,5 +456,28 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+
+	tx.db.prepedCacheMutex.Lock()
+	defer tx.db.prepedCacheMutex.Unlock()
+
+	var ok bool
+	var ps *sql.Stmt
+	ps, ok = tx.db.prepedCache.lookup(s, tx)
+	if !ok {
+		ps, ok = tx.db.prepedCache.lookup(s, nil)
+		if !ok {
```
TheSignPainter98: Would it be cleaner to swap the branches of this `if` to keep the code handling lookup and and then the result of that lookup close together? Also, would the assignment and `if` benefit from being merged? 

I notice that regardless of whether the cache lookup was successful, the new value was added to the cache. Is this correct?

Essentially, would something like this work?
```go
if ps, ok = tx.db.prepedCache.lookup(...); ok {
    ps = tx.tx.Stmt(ps) // ...
} else {
    ps, err = tx.tx.PrepareContext(...)
    if err != nil {
        err = tx.db.prepedCache.add(...)
    }
}
```

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
+	c.ll = list.New()
+	c.c = make(map[index]*list.Element)
+	c.size = size
+	return c
+}
+
+func (c *lruCache) lookup(s *Statement, tx *TX) (*sql.Stmt, bool) {
+	e, ok := c.c[index{tx: tx, s: s}]
+	if !ok {
+		return nil, ok
```
TheSignPainter98: I think this return would be slightly clearer as `return nil, false` as there's only one value `ok` could be here

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
+	e := c.ll.PushFront(&entry{key: k, value: ps})
+	c.c[k] = e
+	if c.ll.Len() > c.size {
+		b := c.ll.Back()
+		delete(c.c, b.Value.(*entry).key)
+		c.ll.Remove(b)
+		// Close the prepared statement on removal from the cache.
+		err = b.Value.(*entry).value.Close()
+	}
+	return err
```
TheSignPainter98: I think it would be clearer to remove the `var err error` and just have two return statements (also it's a line shorter :P)

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
+	defer db.prepedCacheMutex.Unlock()
+	ps, ok := db.prepedCache.lookup(s, nil)
+	if !ok {
+		ps, err = db.db.PrepareContext(ctx, qe.QuerySQL())
+		if err == nil {
+			err = db.prepedCache.add(s, nil, ps)
+		}
+	}
+
+	return &Query{stmt: ps, qe: qe, ctx: ctx, err: err}
```
TheSignPainter98: This one return statement feels like it's tying together two paths, making it unclear at first glance which fields are and aren't `nil`. Perhaps it would be better to have a separate return in case the `db.prepedCache.add(...)` errors? This would also remove the need for a shared `var err error`.

sqlair.go
```diff
@@ -92,15 +97,98 @@ type Iterator struct {
...
 
+// A Least Recently Used Cache using a doubly linked list.
+type lruCache struct {
+	ll   *list.List
+	c    map[index]*list.Element
+	size int
+}
+
+// The cache key.
+type index struct {
```
TheSignPainter98: This comment feels so short it could be encoded in the type name. Perhaps this type could be renamed to `cacheKey`?

package_test.go
Aflynn50: I'll change it to `TestPreparedStmtCaching` to disambiguate it from `sqlair.Prepare`

sqlair.go
Aflynn50: I've changed `c` to `cache` and put some explanation comments above `ll` saying that it is a doubly linked list and that the last element is the LRU. It may still be a bit confusing but idk if the other two names can really be improved much to add clarity (i.e. `dlList` is still pretty opaque and `doublyLinkedList` is a bit long imo).

sqlair.go
Aflynn50: Very true, changed.

sqlair.go
Aflynn50: The `TX` will only be in the cache if it hasn't been committed or rolled back. When they do do either of those operations it is removed. I don't think that it would keep the db connection alive even if the `TX` was not removed as long as it had been committed or rolled back the connection is closed. If there is a situation that arises where the cache api is misused like this then the tests would fail.

I'm not too worried about this in general since its not a public api and its not going to be widely used within the program.

sqlair.go
Aflynn50: Ah, nice, in that case I think I'll just change it to `prepared` since its only one letter different

sqlair.go
Aflynn50: True, will change

sqlair.go
Aflynn50: Changed it to: "turn a Stmt prepared on the db into one tied to a specific transaction.". Hopefully thats a bit clearer

sqlair.go
Aflynn50: True. I've changed the order, though it does need to be added to the cache for both the branches so I left the `add` outside the branches.

sqlair.go
Aflynn50: Good point.

sqlair.go

sqlair.go
TheSignPainter98: I think it could be better to name the field after what it does rather than how it's made, perhaps `recentlyUsed` could work?

sqlair.go
TheSignPainter98: This makes sense about it not being a public API, it'll be easier to police correct usage of this as it's internal.

I see in the definition of `TX` that there's (indirectly) a `*sql.DB` which would be kept alive which made me suspicious

sqlair.go
TheSignPainter98: Ah right, the add also handles the position in the least recently used list. To me, `add` sounds like the element isn't already present in the cache, but it currently has two responsibilities:

- Adding new elements
- Bumping already-included elements to the start of the LRU list

Perhaps it would be cleaner to separate these out into two functions:
- `add` only adds new elements
- a new function, perhaps called `bump`, which moves the given entry to the front of the LRU list and is only used for existing elements

sqlair.go
TheSignPainter98: I'd say 'convert' rather than 'turn,' but other than that, this sounds good

sqlair.go
Aflynn50: I would argue that the current implementation makes for a pretty clean and simple interface for the cache, i.e. if you add an element to the cache you can expect it to be at the front of the queue, irrespective of whether it was in the cache before. By adding bump we would end up with conditionals and a few more lines of code every time we wanted to add something to the cache, because when add something we also always want to bump it.

sqlair.go
Aflynn50: Changed to have a separate return for the `nil` case.

sqlair.go
Aflynn50: True, that makes it clearer.

export_test.go
```diff
@@ -0,0 +1,20 @@
+package sqlair
+
+import (
+	"database/sql"
+	"sync"
+
+	"github.com/google/uuid"
```
niemeyer: We don't need that kind of UUID for this purpose. Let's just have a monotonic integer atomically incremeneted.

sqlair.go
```diff
@@ -25,7 +28,14 @@ var ErrNoRows = sql.ErrNoRows
 // Statement represents a SQL statement with valid SQLair expressions.
 // It is ready to be run on a SQLair DB.
 type Statement struct {
+	id uuid.UUID
 	pe *expr.PreparedExpr
+
+	// A list of *sql.Stmt caches that this Statement appears in. This is
+	// needed to remove and close these the db-prepared statements when the
+	// finalizer for this Statement is run.
+	caches      []*cache
```
niemeyer: That creates the same problem that was solved by the ID system above: caches will pile up here until the system runs out of memory. To allow for a single finalizer to exist, we need a double-sided cache so that we can clean both sides when either of them go out of memory. Vaguely:
```
var cacheMutex sync.Mutex
var cacheStmtDB = map[stmtID] = map[dbID] = dbSTMT
var cacheDBStmt = map[dbID] = map[stmtID] = dbSTMT
```

sqlair.go
```diff
@@ -25,7 +28,14 @@ var ErrNoRows = sql.ErrNoRows
 // Statement represents a SQL statement with valid SQLair expressions.
 // It is ready to be run on a SQLair DB.
 type Statement struct {
+	id uuid.UUID
```
niemeyer: Can we please call this as `cacheID` so it's more obvious why we have it?

sqlair.go
```diff
@@ -92,15 +100,66 @@ type Iterator struct {
 	started bool
 }
 
+// cache is a cache for statements prepared on a database/transaction.
+type cache struct {
```
niemeyer: Given the earlier points we might not need this anymore, but if we keep it for some reason let's please give it a more distinctive name.

niemeyer (changes requested):
Thanks, that captures some of the ideas we discussed, but it's opening the same issue from the opposite side. More details below.

export_test.go
Aflynn50: Changed.

sqlair.go

sqlair.go

sqlair.go
Aflynn50: Done.

sqlair.go
```diff
@@ -99,8 +156,39 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+	ps, ok := stmtDBCache[s.cacheID][db.cacheID]
+	cacheMutex.RUnlock()
+	if !ok {
+		ps, err = db.db.PrepareContext(ctx, s.pe.SQL())
+		if err != nil {
+			return nil, err
+		}
+		cacheMutex.Lock()
+		stmtDBCache[s.cacheID][db.cacheID] = ps
+		dbStmts[db.cacheID][s.cacheID] = true
```
niemeyer: Race condition here: two different statements may be inserted because the lock wasn't held. One of those isn't closed. It needs to check again with the write lock acquired.

package_test.go
```diff
@@ -227,6 +226,14 @@ func (s *PackageSuite) TestValidIterGet(c *C) {
...
 
 	db := sqlair.NewDB(sqldb)
+	defer func() {
+		for _, dropTable := range dropTables {
+			err = db.Query(nil, sqlair.MustPrepare(dropTable)).Run()
+			if err != nil {
+				c.Fatal(err)
+			}
+		}
+	}()
```
niemeyer: Lots of copy & paste. Let' please have a function that takes just the table names.

Also, in the particular for this specific case, we have a list of tests below that have idea about this cleanup logic here. Next week one of us might add an entry to the table, and nothing will pick up that the cleanup logic is left behind. Ideally these two worlds would be connected somehow.

package_test.go
```diff
@@ -1017,41 +1057,212 @@ func (s *PackageSuite) TestTransactions(c *C) {
...
+		c.Errorf("expected %q, got %q", sqlair.ErrTXDone, err)
+	}
 	err = q.Run()
-	c.Assert(err, ErrorMatches, "sql: transaction has already been committed or rolled back")
+	if !errors.Is(err, sql.ErrTxDone) {
+		c.Errorf("expected %q, got %q", sql.ErrTxDone, err)
+	}
+}
+
+func (s *PackageSuite) TestPreparedStmtCaching(c *C) {
```
niemeyer, line 1053: This function looks like it would benefit from some nice reorganization. As it is it's a very long list of statements that seem to depend on each other, with multiple levels that just run variables out of scope on purpose but without clarity for the reader. This will probably become unmaintainable quickly, if it isn't yet.  Breaking down into nice functions that take care of individual tasks, with proper scoping, might help.

sqlair.go
```diff
@@ -21,11 +24,42 @@ import (
...
+// When a SQLair Statement is run on a DB/TX with the Query method it is first
+// prepared on the database. The prepared statement is then stored in the
+// stmtDBCache.
+// When the statement is placed in the cache a finalizer function is set on it.
+// On garbage collection it cycles through the dbIDs and closes each sql.Stmt.
+// The finalizer then removes the stmtID from stmtDBCache and dbStmts.
+// Similarly a finalizer is set on the SQLair DB which closes the sql.DB and
+// removes its dbID from dbStmts and stmtDBCache.
+var stmtDBCache = make(map[stmtID]map[dbID]*sql.Stmt)
+var dbStmts = make(map[dbID]map[stmtID]bool)
```
niemeyer: The common suffix seems beneficial here: dbStmtCache

sqlair.go
```diff
@@ -42,7 +76,14 @@ func Prepare(query string, typeSamples ...any) (*Statement, error) {
...
 	}
-	return &Statement{pe: preparedExpr}, nil
+
+	cacheID := atomic.AddInt64(&stmtIDCount, 1)
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	stmtDBCache[cacheID] = make(map[dbID]*sql.Stmt)
+	var s = Statement{pe: preparedExpr, cacheID: cacheID}
+	runtime.SetFinalizer(&s, stmtFinalizer)
+	return &s, nil
```
niemeyer: For a minor clarity improvement, can we please move the & to the var definition.

sqlair.go
```diff
@@ -55,31 +96,46 @@ func MustPrepare(query string, typeSamples ...any) *Statement {
...
-	return &DB{db: db}
+func dbFinalizer(db *DB) {
+	// There is no need to close the sql.Stmts here, the resources in the
+	// database are freed on db.Close.
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	db.db.Close()
+	stmtCache := dbStmts[db.cacheID]
+	for stmtCacheID, _ := range stmtCache {
+		delete(stmtDBCache[stmtCacheID], db.cacheID)
```
niemeyer: Although it's commented above, I think we should close these statements here. From our conversation, it's not even clear that statements are indeed being cleaned up, and it won't hurt to have the logic doing the right thing regardless. The db.Close should be done afterwards, again just for correctness.

sqlair.go
```diff
@@ -55,31 +96,46 @@ func MustPrepare(query string, typeSamples ...any) *Statement {
...
-	qe  *expr.QueryExpr
-	qs  querySubstrate
-	ctx context.Context
-	err error
+	qe   *expr.QueryExpr
+	stmt *sql.Stmt
+	ctx  context.Context
+	err  error
+	// If this query is not in a transaction tx is nil.
+	tx *TX
```
niemeyer: The comment might be inlined and moved to the positive for maginal clarity:
```
tx *TX  // Only set for queries in transactions.
```

sqlair.go
```diff
@@ -152,19 +240,31 @@ func (q *Query) Iter() *Iterator {
...
+		result, err = q.stmt.ExecContext(q.ctx, q.qe.QueryArgs()...)
+	}
+	if err != nil {
+		return &Iterator{qe: q.qe, err: err}
 	}
-	return &Iterator{qe: q.qe, rows: rows, cols: cols, err: err, result: result}
+	var onClose func() error
+	if q.tx != nil {
+		onClose = func() error {
+			return q.stmt.Close()
```
niemeyer: Seems a bit unusual to have this inside a closure when the logic is all local and clearly defined. We might just have the stmt and close it at the proper place.

sqlair.go
```diff
@@ -327,6 +434,15 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
...
 	tx *sql.Tx
+	db *DB
+	// done transitions from 0 to 1 exactly once, on Commit or Rollback.
+	// Once done, all operations fail with ErrTxDone. Use atomic operations
+	// only when checking value.
+	done int32
+}
+
+func (tx *TX) isDone() bool {
+	return atomic.LoadInt32(&tx.done) == 1
```
niemeyer, line 483: Instead of having the comment above, we might just have a second function here:
```
func (tx *TX) setDoneOnce() error { ... }
```
isDone and setDonceOnce together would be about the size of the comment, and actually implement what it says clearly, and more correctly (it's currently ambiguous: "use only atomic" vs. "use only when checking").

sqlair.go
```diff
@@ -369,7 +491,31 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+		return &Query{ctx: ctx, err: ErrTXDone}
+	}
 
 	qe, err := s.pe.Query(inputArgs...)
-	return &Query{qs: tx.tx, qe: qe, ctx: ctx, err: err}
+	if err != nil {
+		return &Query{ctx: ctx, err: err}
+	}
+
+	var ps *sql.Stmt
```
niemeyer: A couple of nitpicks: a bit curious to have that long comment sandwished between the two variables that are used for the purpose of the code after the comment. Also, per the above this would be `sqlstmt`.

sqlair.go
```diff
@@ -55,31 +96,46 @@ func MustPrepare(query string, typeSamples ...any) *Statement {
...
-}
-
 // Query holds the results of a database query.
 type Query struct {
-	qe  *expr.QueryExpr
-	qs  querySubstrate
-	ctx context.Context
-	err error
+	qe   *expr.QueryExpr
+	stmt *sql.Stmt
```
niemeyer: We need a way to differentiate between our stmt and the sql stmt.  How about using a convention that is actually a bit unusual in terms of casing: our variables are named `tx`, `db`, and `stmt`, while the ones in the sql package are named `sqltx`, `sqldb`, and `sqlstmt`. How does that sound?

If accepted, let's please have a pass over the diff in this PR to clean up the use cases so we flag each of these appropriately.

niemeyer (changes requested):
Looks pretty good!  First pass, and maybe final:

sqlair.go
Aflynn50: Ah yea, good point, I was aware of this and thought it wouldn't be a problem but the overwritten statement would never get closed. Changed.

sqlair.go

package_test.go
Aflynn50: Changed to be a function.

package_test.go
Aflynn50: Changed to use functions and added comments to explain.

sqlair.go

sqlair.go

sqlair.go
Aflynn50: Added `stmt` straight into the `Iterator`.

sqlair.go
Aflynn50: Changed.

sqlair.go
Aflynn50: I'll make this change

sqlair.go
Aflynn50: Moved.

package_test.go
```diff
@@ -334,12 +344,15 @@ func (s *PackageSuite) TestIterGetErrors(c *C) {
...
-	dropTables, sqldb, err := personAndAddressDB()
+	tables, sqldb, err := personAndAddressDB()
 	if err != nil {
 		c.Fatal(err)
 	}
 
 	db := sqlair.NewDB(sqldb)
+	defer func() {
+		c.Assert(dropTables(db, tables...), IsNil)
+	}()
```
niemeyer: In the future, it would be slightly nicer to have dropTables taking c and doing the work itself, so the closure can be dropped in all those defers.

package_test.go
```diff
@@ -1017,41 +1021,204 @@ func (s *PackageSuite) TestTransactions(c *C) {
...
+	q1 := `
+		SELECT &Person.*
+		FROM person
+		WHERE name = "Fred"
+	`
+	q2 := `
+		SELECT &Person.*
+		FROM person
+		WHERE name = "Mark"
+	`
```
niemeyer: Nitpick: I've noticed a tendency in recent PRs to overbreak SQL statements. These are easy one-liners.

package_test.go
```diff
@@ -1017,41 +1021,204 @@ func (s *PackageSuite) TestTransactions(c *C) {
...
+		stmt, err := sqlair.Prepare(q2, Person{})
+		c.Assert(err, IsNil)
 
-	err = db.Query(ctx, sqlair.MustPrepare(dropTables)).Run()
+		db1ID := createDBAndTestStmt(stmt)
+		db2ID := createDBAndTestStmt(stmt)
+
+		// Run the garbage collector and wait one millisecond for the finalizer to finish.
+		runtime.GC()
+		time.Sleep(1 * time.Millisecond)
```
niemeyer, line 1155: Looks like a clear source of future races. Since apparently this comes from conventions in the runtime, I'm hoping that it's not the time and in fact it just deschedules the goroutine, which would be okay. If not, we'll be seeing this code again soon. :)

sqlair.go
```diff
@@ -55,31 +96,44 @@ func MustPrepare(query string, typeSamples ...any) *Statement {
...
 }
 
-func NewDB(db *sql.DB) *DB {
-	return &DB{db: db}
+func dbFinalizer(db *DB) {
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	stmtCache := dbStmtCache[db.cacheID]
+	for stmtCacheID, _ := range stmtCache {
+		stmtDBCache[stmtCacheID][db.cacheID].Close()
```
niemeyer: Not sure if Go optimizes it yet, but instead of accessing the map multiple times for the same key, it's conventional to just take the value out and use it.

sqlair.go
```diff
@@ -214,6 +319,10 @@ func (iter *Iterator) Get(outputArgs ...any) (err error) {
 
 // Close finishes the iteration and returns any errors encountered.
 func (iter *Iterator) Close() error {
+	var cerr error
+	if iter.sqlstmt != nil {
+		cerr = iter.sqlstmt.Close()
```
niemeyer: Best to set it to nil here as well. Makes a difference on repeated closing, which often happens as a side effect of deferrals.

sqlair.go
```diff
@@ -369,7 +500,31 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+	// A transaction is tied to a specific database connection. When a
+	// Statement is prepared on a database it requires a free connection that
+	// does not have a transaction in progress. If all connections are busy
+	// then it is possible to enter a deadlock.
+	// When all connections are busy, prepare on the transaction rather than
+	// the DB.
+	var sqlstmt *sql.Stmt
+	dbStats := tx.db.sqldb.Stats()
+	if dbStats.MaxOpenConnections > 0 && dbStats.MaxOpenConnections == dbStats.InUse {
+		sqlstmt, err = tx.sqltx.PrepareContext(ctx, s.pe.SQL())
```
niemeyer: There's clearly a race here. Let's chat about this, please.

niemeyer (changes requested):
Thanks for the changes. Only one sticking point remaining.

sqlair.go
Aflynn50: Fixed by adding an explicit connection underneath the transaction.

sqlair.go

sqlair.go

package_test.go
Aflynn50: I'll do a thorough investigation, this was used in the Go runtime package test, but in the latest versions they have actually updated their tests to use channels to indicate when the finalizer is done. We can't do that here because it would involve using a different finalizer to the one in the code so I have stuck with the old approach for now, though I would like to do a more thorough investigation at some point.

package_test.go

package_test.go
Aflynn50: I'll add this to #95 once this is in.

sqlair.go
```diff
@@ -464,24 +468,34 @@ func (db *DB) Begin(ctx context.Context, opts *TXOptions) (*TX, error) {
 	if ctx == nil {
 		ctx = context.Background()
 	}
-	sqltx, err := db.sqldb.BeginTx(ctx, opts.plainTXOptions())
-	return &TX{sqltx: sqltx, db: db}, err
+	sqlconn, err := db.sqldb.Conn(ctx)
+	if err != nil {
+		return nil, err
+	}
+	sqltx, err := sqlconn.BeginTx(ctx, opts.plainTXOptions())
```
niemeyer, line 502: The error needs to be checked and nil returned on issues.

sqlair.go
```diff
@@ -464,24 +468,34 @@ func (db *DB) Begin(ctx context.Context, opts *TXOptions) (*TX, error) {
...
 }
 
 // Commit commits the transaction.
 func (tx *TX) Commit() error {
 	if err := tx.setDone(); err != nil {
 		return err
 	}
-	return tx.sqltx.Commit()
+	if err := tx.sqltx.Commit(); err != nil {
+		return err
```
niemeyer: The connection needs to be closed, whether there are errors or not.

sqlair.go
```diff
@@ -464,24 +468,34 @@ func (db *DB) Begin(ctx context.Context, opts *TXOptions) (*TX, error) {
...
 }
 
 // Rollback aborts the transaction.
 func (tx *TX) Rollback() error {
 	if err := tx.setDone(); err != nil {
 		return err
 	}
-	return tx.sqltx.Rollback()
+	if err := tx.sqltx.Rollback(); err != nil {
+		return err
```
niemeyer: Same.

sqlair.go
```diff
@@ -509,57 +523,16 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
-			dbStmtCache[tx.db.cacheID][s.cacheID] = true
-		}
-		cacheMutex.Unlock()
-	}
-	sqlstmt = tx.sqltx.Stmt(sqlstmt)
-	if conn != nil {
-		err = conn.Close()
-		if err != nil {
-			return &Query{ctx: ctx, err: err}
-		}
```
niemeyer: It's great to see that gone.

sqlair.go
```diff
@@ -152,19 +256,29 @@ func (q *Query) Iter() *Iterator {
...
-		result, err = q.qs.ExecContext(q.ctx, q.qe.QuerySQL(), q.qe.QueryArgs()...)
+		result, err = q.sqlstmt.ExecContext(q.ctx, q.qe.QueryArgs()...)
+	}
+	if err != nil {
+		return &Iterator{qe: q.qe, err: err}
 	}
-	return &Iterator{qe: q.qe, rows: rows, cols: cols, err: err, result: result}
+	var sqlstmt *sql.Stmt
+	if q.tx != nil {
+		sqlstmt = q.sqlstmt
```
niemeyer: Sorry for missing this earlier, but there's something here that doesn't feel quite right.  This is handing to Iter a sqlstmt for being closed, but that sqlstmt lives in the Query. That can't be right. There's nothing in the API of the Query saying that it is a requirement for Iter.Close to be called directly or via one of the other methods, for example,  and that this is a requirement for resources not to leak. Indeed it would be unfriendly to ask for that.  Then, what if Iter is called a second time in the Query?  This woud find a closed sqlstmt, and break down?

Luckily, there's a trivial solution which will actually simplify other things too: create the transaction-bound stmt inside this function instead of elsewhere, and close it immediately after this function is done with it. Then let's clean up the rest of the code since we don't need to track it around anymore.

sqlair.go
niemeyer: Please see the other comment above. There is a more relevant issue here unfortunately.

sqlair.go
```diff
@@ -372,7 +524,20 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+		return &Query{ctx: ctx, err: err}
+	}
 
 	qe, err := s.pe.Query(inputArgs...)
-	return &Query{qs: tx.tx, qe: qe, ctx: ctx, err: err}
+	if err != nil {
+		return &Query{ctx: ctx, err: err}
+	}
+
+	sqlstmt = tx.sqltx.Stmt(sqlstmt)
```
niemeyer: Instead of doing this here, let's do it at execution time so that we can close the stmt one immediately.

niemeyer (changes requested):
Thanks. A few more details to address still:

sqlair.go
Aflynn50: If the connection is closed before the query results have been read the first call to `Rows.Next` returns false so sadly the quick fix doesn't work.

You are right that it would be hostile to require the user to call `Close` though. The `database/sql` rows object will automatically close when `Rows.Next` returns false. I've added this behavior to the `Iter` so that the connection is closed when `Iter.Next` returns false.

sqlair.go
Aflynn50: See comment above.

sqlair.go
Aflynn50: Changed.

sqlair.go
Aflynn50: Closing of connection deferred at the start.

sqlair.go
Aflynn50: Closing of connection deferred at the start.

sqlair.go
Aflynn50: I've left this as is, see comment above about the stmt.

sqlair.go
niemeyer: Why would the connection be closed?  The connection is associated with the TX. If people close the transaction (Commit, Rollback) before reading results are finished, they will get an error, as they should?

> I've added this behavior to the Iter so that the connection is closed when Iter.Next returns false.

How would it make sense to close the connection because the iterator is finished?

Aflynn50: Ah sorry, I meant to write prepared statement, not connection.

Aflynn50: The code I've implemented closes the stmt, not the connection.

niemeyer (approved):
Thanks for all the tuning.

PR #82, opened 2023-06-12
Allow SQL functions as column names

Aflynn50:
Relax the restriction that columns in output expressions can only be of the form `prefix.name`. SQLair will now accept `func(...)` as a column name in output expressions of the form `func(...) AS &Type.tag`.

The functions are parsed as a string made up of alphanumeric chars + `_` following by an opening bracket containing anything at all (including SQLair io expressions).

An error will be thrown if the user tries to put a function into a map or struct with an asterisk, i.e. `func(...) AS &MyStruct.*`. 

Functions in output expressions can now also contain io expressions e.g. `func($P.id) AS &M.id`. This has been achieved by refactoring `Prepare` to generate the SQL inside `prepareOutput` and `prepareInput` and build up the `ParsedExpr` progressively rather than doing a double pass (`prepareParts` can be called recursively from inside `prepareOutput`, adding the prepared information to the `ParsedExpr` on the fly)


internal/expr/parser.go
```diff
@@ -386,6 +392,44 @@ func (p *Parser) parseColumn() (fullName, bool, error) {
 	return fullName{}, false, nil
 }
 
+func (p *Parser) parseFuncName() (string, bool, error) {
+	cp := p.save()
+	if p.skipName() {
+		if p.skipByte('(') {
+			bracketLevel := 1
+			for {
```
TheSignPainter98: Just to make the breaking conditions a little clearer, could a condition be used instead? Like
```go
for p.pos != len(p.input) && bracketLevel != 0 {
```

Aflynn50: I've actually just changed it to:
```go
if p.pos == len(p.input) {
	return "", false, nil
}
```
In this case we probably want to actually return false (an error is probably a bad idea since the thing we are parsing could well not be a function).

internal/expr/expr_test.go
```diff
@@ -467,6 +473,10 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 		query:       "SELECT * AS &.* FROM t",
 		prepareArgs: []any{struct{ f int }{f: 1}},
 		err:         `cannot prepare expression: cannot use anonymous struct`,
+	}, {
+		query:       "SELECT avg(*) AS &Person.* FROM t",
+		prepareArgs: []any{Person{}},
+		err:         `cannot prepare expression: invalid tag/column name "avg(*)" in "avg(*) AS &Person.*"`,
```
niemeyer: Per discussion, let's please constraint it so that both for maps and structs, the field must be named explicitly whenever a function is used, star or not. We can support something later if it becomes clearly necessary, but forbidding funky syntax is the better and safer way to start.

internal/expr/parser.go
```diff
@@ -386,6 +392,42 @@ func (p *Parser) parseColumn() (fullName, bool, error) {
...
+				continue
+			}
+			if p.skipByte(')') {
+				bracketLevel--
+				if bracketLevel == 0 {
+					break
+				}
+				continue
+			}
+			p.pos++
```
niemeyer: It doesn't feel great that we're picking up what specific kind of syntax is supported here. This is very brittle and will fail often. We need to design this so that we don't need to be valiidating the details of this expression, just like we don't validate everywhere else except for the details that we do care about. So I'd try to consider that whole left side of the expression as either a column name (table, asterisk, etc), in which case we know exactly what's going on, or as a function holding arbitrary expressions as parameters, in which case we need to recurse into the parser. A good litmus test that this is well designed is supporting input parts inside the function parameters.

niemeyer (changes requested):
Supporting this seems worthwhile, but the function parsing seems a bit simplistic at the moment. Some suggestions:

internal/expr/expr_test.go

internal/expr/parser.go
Aflynn50: I've changed `parseFunc` to recurse into the parser if we find a SQL function, it now works with nested input expressions.

The function `parseColumns` has the behaviour you describe above, it first checks for a function as a column, then for a regular column name

internal/expr/parser.go
```diff
@@ -89,14 +97,19 @@ func (pe *ParsedExpr) String() string {
 	return out.String()
 }
 
+// bypass adds the string between the previous I/O part and the current part.
+func (p *Parser) addBypass() {
```
TheSignPainter98: The function name is out-dated in the comment just here :)

internal/expr/parser.go
```diff
@@ -206,17 +219,38 @@ loop:
...
-
+	p.partStart = p.pos
 	return nil
 
 }
 
+// startSubpart is called to allow parsing of nested parts.
+func (p *Parser) startSubpart() *checkpoint {
+	scp := p.save()
+	p.init(p.input[p.pos:])
```
TheSignPainter98: As `init` feels like a method which is only called at the beginning, perhaps a comment would be good here? Perhaps something like
```go
// Clear parser state.
```
Otherwise, I think a name-change is necessary here, maybe `setInput`?

internal/expr/parser.go
```diff
@@ -206,17 +219,38 @@ loop:
...
+	return scp
+}
+
+// collectSubpart must be called to collect all parts parsed since startSubpart and move the parser up a nesting level.
+func (p *Parser) collectSubpart(scp *checkpoint) *ParsedExpr {
+	pos := p.pos
+	pe := &ParsedExpr{p.parts}
+	scp.restore()
+	p.pos = p.pos + pos
```
TheSignPainter98: I think this can be `+=` :P

internal/expr/expr_test.go
```diff
@@ -50,6 +50,12 @@ var tests = []struct {
 	prepareArgs      []any
 	expectedPrepared string
 }{{
+	"function with nested input",
+	`SELECT max($Person.id, 20) AS &Manager.id FROM person`,
+	`[Bypass[SELECT ] Output[[[Bypass[max(] Input[Person.id] Bypass[, 20)]]] [Manager.id]] Bypass[ FROM person]]`,
+	[]any{Person{}, Manager{}},
+	`SELECT max(@sqlair_0, 20) AS _sqlair_0 FROM person`,
+}, {
```
TheSignPainter98: Would it be good to add a test where the function calls are nested? E.g.
```sql
SELECT max(min(100, $Person.id), 20) AS &...
```

internal/expr/parser.go
```diff
@@ -206,17 +219,38 @@ loop:
...
 
-	p.skipBlanks()
-
+	p.partStart = p.pos
 	return nil
 
 }
 
+// startSubpart is called to allow parsing of nested parts.
+func (p *Parser) startSubpart() *checkpoint {
```
TheSignPainter98: This comment feels slightly off to me, it seems like a comment I'd expect on a call itself rather than the method definition. I think it's worth mentioning the checkpoint, perhaps something like this?
```go
// startSubpart creates a checkpoint and prepares to parse a nested part.
```
Also, given that the sub-part checkpoints are handled slightly differently (when restoring, the new position is computed rather than being a simple replacement), is it worth adding a new `subpartCheckpoint` type here? This would also match the use of `scp` rather than `cp` so it could feel natural

internal/expr/parser.go
```diff
@@ -206,17 +219,38 @@ loop:
...
+	scp := p.save()
+	p.init(p.input[p.pos:])
+	return scp
+}
+
+// collectSubpart must be called to collect all parts parsed since startSubpart and move the parser up a nesting level.
+func (p *Parser) collectSubpart(scp *checkpoint) *ParsedExpr {
```
TheSignPainter98: As with `startSubpart`, I think the language can be cleaned slightly here, perhaps something like this?
```go
// collectSubpart returns the nested expression parsed since the
// corresponding sub-part checkpoint and restores the parser to
// its new position.
```

internal/expr/parser.go
```diff
@@ -392,47 +426,36 @@ func (p *Parser) parseColumn() (columnExpr, bool, error) {
...
+			if p.bracketLevel == 0 {
+				break
 			}
 
-			if p.skipByte('(') {
-				// Keep track of bracket nesting level in function to identify
-				// closing bracket.
-				bracketLevel++
-				continue
+			if p.pos == len(p.input) {
```
TheSignPainter98: Should these two break conditions be merged?

internal/expr/parser.go
```diff
@@ -89,14 +97,19 @@ func (pe *ParsedExpr) String() string {
...
 // part.
 func (p *Parser) add(part queryPart) {
-	// Add the string between the previous I/O part and the current part.
 	if p.prevPart != p.partStart {
-		p.parts = append(p.parts,
-			&bypassPart{p.input[p.prevPart:p.partStart]})
+		p.parts = append(p.parts, &bypassPart{p.input[p.prevPart:p.partStart]})
 	}
 
 	if part != nil {
```
TheSignPainter98: Just from our discussion, could this be removed due to the new `addBypass` method?

internal/expr/parser.go
```diff
@@ -392,47 +426,36 @@ func (p *Parser) parseColumn() (columnExpr, bool, error) {
...
-		rawFunc := p.input[cp.pos:p.pos]
-		// Parse any expressions inside the function.
-		funcParser := NewParser()
-		pe, err := funcParser.Parse(rawFunc)
-		if err != nil {
-			return funcExpr{}, false, err
 		}
-		return funcExpr{raw: rawFunc, pe: pe}, true, nil
+		// Add any remaining unparsed string input to the parser.
+		p.add(nil)
```
TheSignPainter98: Just from our discussion earlier, instead of `p.add(nil)`, would the new `p.addBypass()` fit here? Otherwise, `addBypass` isn't used in this PR.

Also, just thinking about the naming, as `add` implicitly adds the bypass at the moment, it looks odd that (at a glance) the bypass is only explicitly added once. Perhaps this method should be renamed to something like `addRemainderAsBypass` or just `addRemainder`?

internal/expr/parser.go
```diff
@@ -392,47 +426,36 @@ func (p *Parser) parseColumn() (columnExpr, bool, error) {
 	return fullName{}, false, nil
 }
 
-func (p *Parser) parseFuncName() (columnExpr, bool, error) {
+func (p *Parser) parseFunc() (columnExpr, bool, error) {
 	cp := p.save()
-	if p.skipName() && p.skipByte('(') {
-		bracketLevel := 1
+	scp := p.startSubpart()
+	if p.skipName() && p.peekByte('(') {
```
TheSignPainter98: The change to `peekByte` here feels slightly off, is this because `advance` is needed to handle the change in bracket level? Perhaps `skipByte` should handle this too

TheSignPainter98 (commented):
Looks good, just a little food for thought

internal/expr/parser.go
Aflynn50: I've actually removed this function and left `add` as it is, but I have added `addRemainder`.

internal/expr/parser.go
Aflynn50: Added comment.

internal/expr/parser.go
Aflynn50: Changed

internal/expr/expr_test.go
Aflynn50: I do have a test for this, the only reason this test was separate (at the top) was so I could debug it more easily. I have added a second test for having multiple input statements in different levels of the the nested functions though.

internal/expr/parser.go
Aflynn50: Changed the comment. I did have a separate type for it at some point yesterday but with the current implementation you can actually use the checkpoint it returns as a regular checkpoint. I've changed the use of it in `parseFunc` to reflect this as well.

internal/expr/parser.go
Aflynn50: Cleaned up.

internal/expr/parser.go
Aflynn50: Merged.

internal/expr/parser.go
Aflynn50: See comment above.

internal/expr/parser.go
Aflynn50: I've added an `addRemainder` method to use for this case, its a good name for it.

internal/expr/parser.go
Aflynn50: So theres actually another reason this is here. If an input expression starts right after the bracket then advance won't notice its there. It checks if the preceding character is of interest so if it starts on a '$' then it won't stop since it would be expecting to parse an input expression on the next character, i.e. the one after the '$'.

I agree its not amazing to have `peekByte` there, but the other solutions that don't involve re-implementing the main parser are all more ugly.

internal/expr/parser.go
```diff
@@ -99,16 +107,23 @@ func (p *Parser) add(part queryPart) {
 			&bypassPart{p.input[p.prevPart:p.partStart]})
```
letFunny: Can we change this to use the newly created function `p.addRemainder()`?

internal/expr/prepare.go
```diff
@@ -114,7 +119,18 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
 		return info, nil
 	}
 
-	addColumns := func(info typeInfo, tag string, column fullName) error {
+	addSQL := func(column string, tm typeMember) error {
```
letFunny: I think the naming here is a bit confusing as `addSQL` is really adding a column to the SQL output and adding it to the PreparedExpr object. Maybe it would be clearer by moving
```
pe.outputs = append(pe.outputs, tm)
```
elsewhere, and changing the name from `addSQL` to `addColumnSQL` or something similar.

internal/expr/prepare.go
```diff
@@ -50,47 +49,54 @@ func getKeys[T any](m map[string]T) []string {
...
-func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember, error) {
-	var outCols = make([]fullName, 0)
-	var typeMembers = make([]typeMember, 0)
-
+func (pe *PreparedExpr) prepareOutput(ti typeNameToInfo, p *outputPart) (err error) {
+	defer func() {
+		if err == nil && len(pe.sql) > 1 {
+			pe.sql = pe.sql[:len(pe.sql)-2]
+		}
+	}()
```
letFunny: Can we write a comment in this function saying that this is removing the trailing `, ` that we are adding in the output (which seems to be what is doing)?

Also, this seems to be rather contrived, is the alternative of tracking the state and not outputting the trailing comma much more complex?

internal/expr/prepare.go
```diff
@@ -125,81 +141,99 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
...
 			}
 
-			if err = addColumns(info, t.name, c); err != nil {
-				return nil, nil, err
+			switch c := c.(type) {
+			case funcExpr:
+				if err = pe.prepareParts(ti, c.pe.queryParts); err != nil {
+					return err
+				}
+				err = addColumn(info, t.name, fullName{})
```
letFunny: Why are we passing here an empty column name? Is the intention that the `addColumn` function prints only the "AS ..." part in the line below?
```
pe.sql += column + " AS " + markerName(len(pe.outputs)) + ", "
```

letFunny (commented):
Looking good, let's try to make it as simple as possible.

internal/expr/parser.go
Aflynn50: I've changed it. Might as well reuse the code. I've left the commend though to make it clear what is happening.

Aflynn50:
#### Notes from an in person discussion with @niemeyer and @letFunny.
An aim of this PR is to allow users to write input statements inside functions that are part of output expressions. For example: `SELECT max($Person.id, 10) AS &Person.id FROM ...`. Output expressions are currently parsed into an `outputPart`. This includes the list of columns captured in the output expression. The columns must now contain a representation of the function, including any `inputParts` it contains.

At the time of writing, this PR saves functions as a `ParsedExpr` which is stored within the `outputPart`. When the SQL is generated the `ParsedExpr` must be recursed over by `Prepare` to generate the SQL for the function (including any `inputParts` within it). This requied some changes to the way preparing worked such as building up the SQL at the same time as it is prepared rather than doing it afterwards.

An alternative method of parsing the functions would be to not save them in the output expression. For example: `Bypass[SELECT max(] Input[Person.id] Bypass[, 10)] Output[Person.id]. The limitation with this approach is that it is not easily possible to parse functions in the current syntax. For example:
```
   SELECT (col1, max($Person.id, 10), col3) AS (&Person.id1, &Person.id2, &Person.id3) FROM ...
-> Bypass[SELECT (col1, max(] Input[Person.id] Bypass[, 10), col3)] Output[Person.id1 Person.id2 Person.id3]
```
Parsing like this makes it a lot harder to generate the SQL. The parser would need to remember the position of each of the columns somewhere to generate the SQL needed:
```
SELECT col1 AS _sqlair_0, max($Person.id, 10) AS _sqlair_1, col3 AS _sqlair_2 FROM ...
```
This appraoch is implemented [here](/Aflynn50/sqlair/tree/expressive-columns-flat-prepare), but only for functions associated with a single type rather than in a list of columns.

internal/expr/prepare.go
Aflynn50: addColumn has disappeared with the latest updates from main :)

internal/expr/prepare.go
Aflynn50: I've added the comment.

I would much rather track the state but it would end up adding in so many lines of extra code that I just can't bring myself to do it. We currently write a comma in four different places, each of which would need a few lines to handle this.

internal/expr/prepare.go
Aflynn50: Yes, its now more clear with the new logic from merging with main, I've added a comment too.

PR #83, opened 2023-06-26
Add support for embedded/nested structs

Aflynn50:
Add support for embedded/nested structs to SQLair input and output expressions. Embedded structs are structs included without a field name and nested structs are those included with a field name.
e.g. embedded:
```go
type S struct {
    E
}
```
and nested:
```go
type S struct {
    e E
}
```
A SQLair output expression can now be something like `&MyStruct.myfieldtag` where the tagged field was nested/embedded multiple layers deep. The tags of the inner structs are referenced as if they were tags of the top level struct. An asterisk expression (e.g. `&MyStruct.*`) will now include all embedded/nested tagged fields in a struct.

Pointers to nested structs are also handled as above.

If the field containing a nested struct has a tag on it then it will be ignored (unless the struct in the field implements `sql.Scanner`).

If there are clashing tags between nested structs an error will be returned from `Prepare`.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -51,6 +51,22 @@ type IntSlice []int
 
 type StringSlice []string
 
+type StructWithEmbedded struct {
+	E1 *E1
+	E3
+}
+
+type E1 struct {
+	F1 string `db:"col1"`
+}
+type E2 struct {
+	F2 string `db:"col2"`
+}
+type E3 struct {
+	F3 string `db:"col3"`
+	E2 E2
+}
+
 var tests = []struct {
 	summary        string
 	query          string
@@ -66,8 +82,8 @@ var tests = []struct {
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
-	summary:        "spaces and tabs",
-	query:          "SELECT p.* 	AS 		   &Person.*",
+	summary: "spaces and tabs",
+	query: "SELECT p.* 	AS 		   &Person.*",
 	expectedParsed: "[Bypass[SELECT ] Output[[p.*] [Person.*]]]",
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
@@ -403,6 +419,26 @@ AND z = @sqlair_0 -- The line with $Person.id on it
 	expectedParams: []any{},
 	// This is valid in SQLite (though not in MySQL).
 	expectedSQL: "SELECT name FROM person WHERE id IN ()",
+}, {
+	summary:        "embedded struct asterisk output",
+	query:          "SELECT &StructWithEmbedded.* FROM address WHERE id = 1000",
+	expectedParsed: "[Bypass[SELECT ] Output[[] [StructWithEmbedded.*]] Bypass[ FROM address WHERE id = 1000]]",
+	typeSamples:    []any{StructWithEmbedded{}},
+	expectedSQL:    "SELECT col1 AS _sqlair_0, col2 AS _sqlair_1, col3 AS _sqlair_2 FROM address WHERE id = 1000",
+}, {
+	summary:        "embedded struct columns output",
+	query:          "SELECT &StructWithEmbedded.col1, &StructWithEmbedded.col2, &StructWithEmbedded.col3 FROM address WHERE id = 1000",
+	expectedParsed: "[Bypass[SELECT ] Output[[] [StructWithEmbedded.col1]] Bypass[, ] Output[[] [StructWithEmbedded.col2]] Bypass[, ] Output[[] [StructWithEmbedded.col3]] Bypass[ FROM address WHERE id = 1000]]",
+	typeSamples:    []any{StructWithEmbedded{}},
+	expectedSQL:    "SELECT col1 AS _sqlair_0, col2 AS _sqlair_1, col3 AS _sqlair_2 FROM address WHERE id = 1000",
+}, {
+	summary:        "embedded struct input",
+	query:          "SELECT &Address.* FROM address WHERE id = $StructWithEmbedded.col1 AND district = $StructWithEmbedded.col2 AND street = $StructWithEmbedded.col3",
+	expectedParsed: "[Bypass[SELECT ] Output[[] [Address.*]] Bypass[ FROM address WHERE id = ] Input[StructWithEmbedded.col1] Bypass[ AND district = ] Input[StructWithEmbedded.col2] Bypass[ AND street = ] Input[StructWithEmbedded.col3]]",
+	typeSamples:    []any{Address{}, StructWithEmbedded{}},
+	inputArgs:      []any{&StructWithEmbedded{E1: &E1{F1: "v1"}, E3: E3{E2: E2{F2: "v2"}, F3: "v3"}}},
+	expectedParams: []any{"v1", "v2", "v3"},
+	expectedSQL:    "SELECT district AS _sqlair_0, id AS _sqlair_1, street AS _sqlair_2 FROM address WHERE id = @sqlair_0 AND district = @sqlair_1 AND street = @sqlair_2",
 }}
 
 func (s *ExprSuite) TestExprPkg(c *C) {
@@ -674,6 +710,20 @@ func (s *ExprSuite) TestBindTypesErrors(c *C) {
 		S string
 	}
 	type myArray [10]any
+	type AmbiguousTags struct {
+		ID1 int `db:"id"`
+		ID2 int `db:"id"`
+	}
+	type IDS1 struct {
+		ID1 int `db:"id"`
+	}
+	type IDS2 struct {
+		ID2 int `db:"id"`
+	}
+	type AmbiguousNestedTags struct {
+		IDS1
+		IDS2
+	}
 	tests := []struct {
 		query       string
 		typeSamples []any
@@ -798,6 +848,14 @@ func (s *ExprSuite) TestBindTypesErrors(c *C) {
 		query:       "SELECT street FROM t WHERE x IN ($myArray[:])",
 		typeSamples: []any{myArray{}},
 		err:         `cannot prepare statement: need supported type, got array`,
+	}, {
+		query:       "SELECT &AmbiguousTags.* FROM t",
+		typeSamples: []any{AmbiguousTags{}},
+		err:         `cannot prepare statement: db tag "id" appears in both field "ID2" and field "ID1" of struct "AmbiguousTags"`,
+	}, {
+		query:       "SELECT &AmbiguousNestedTags.* FROM t",
+		typeSamples: []any{AmbiguousNestedTags{}},
+		err:         `cannot prepare statement: db tag "id" appears in both field "ID2" and field "ID1" of struct "AmbiguousNestedTags"`,
 	}}
 
 	for i, test := range tests {
--- a/package_test.go
+++ b/package_test.go
@@ -500,6 +500,32 @@ func (s *PackageSuite) TestNulls(c *C) {
 }
 
 func (s *PackageSuite) TestValidGet(c *C) {
+	// Embedded/nested structs.
+	type E1 struct {
+		Street string `db:"street"`
+	}
+	type E2 struct {
+		ID int `db:"id"`
+	}
+	type E3 struct {
+		District string `db:"district"`
+		E2       E2
+	}
+	type StructWithEmbedded struct {
+		E1 *E1
+		E3
+	}
+	// Struct with pointers.
+	type StructWithPointers struct {
+		Street   *string         `db:"street"`
+		District *sql.NullString `db:"district"`
+		ID       *int            `db:"id"`
+	}
+
+	// Pointer variables for pointer struct test.
+	mainStreet := "Main Street"
+	oneThousand := 1000
+
 	var tests = []struct {
 		summary  string
 		query    string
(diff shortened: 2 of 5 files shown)


internal/expr/expr_test.go
```diff
@@ -45,6 +45,21 @@ type IntMap map[string]int
...
+	ID int `db:"id"`
+}
+type E3 struct {
+	District string `db:"district"`
+	E2       E2
+}
+type EmbeddedStruct struct {
+	E1 E1
+	E3 *E3
+}
```
TheSignPainter98, line 68: I'm assuming that structs named `E*` are to be embedded, but I don't think any are here, is this intentional?

Also, I think the name of `EmbeddedStruct` is backwards as this struct isn't embedded into anything else. Would it be better to call this `StructWithEmbedded`?

package_test.go
```diff
@@ -496,6 +496,33 @@ func (s *PackageSuite) TestNulls(c *C) {
...
+	}
+	type E3 struct {
+		District string `db:"district"`
+		E2       E2
+	}
+	type EmbeddedStruct struct {
+		E1 E1
+		E3 *E3
+	}
+
```
TheSignPainter98:
```suggestion
```
It could be nice to group all the types together as in other tests :)

internal/expr/expr_test.go
```diff
@@ -401,6 +434,20 @@ func FuzzParser(f *testing.F) {
...
+		ID1 int `db:"id"`
+		ID2 int `db:"id"`
+	}
+	type IDS1 struct {
+		ID1 int `db:"id"`
+	}
+	type IDS2 struct {
+		ID2 int `db:"id"`
+	}
+	type DupNestedTags struct {
```
TheSignPainter98: These tags are duplicate, but I don't think that's necessarily the problem. Is it possible to have duplicate flags, but so long as these aren't accessed with the `.*` notation, this is okay? Just for clearer intention, would it be better to rename this to `AmbiguousNestedFlags`?

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
 	return typeInfo, nil
 }
 
+// getStructFields requires the caller to check that t is a struct.
```
TheSignPainter98: This feels a little off to me, when reading this comment, I expected to be told what the function does, but instead this just lists its preconditions. Given that this function name is very explicit, if the behaviour that `t` already has the correct kind is consistent across the rest of the API, could this comment be dropped altogether?

package_test.go
```diff
@@ -496,6 +496,33 @@ func (s *PackageSuite) TestNulls(c *C) {
...
+		ID int `db:"id"`
+	}
+	type E3 struct {
+		District string `db:"district"`
+		E2       E2
+	}
+	type EmbeddedStruct struct {
+		E1 E1
+		E3 *E3
+	}
```
TheSignPainter98, line 517: Same as before, should these be like the following?
```go
type EmbeddedStruct struct {
    E1
    *E3
}
```

package_test.go
```diff
@@ -496,6 +496,33 @@ func (s *PackageSuite) TestNulls(c *C) {
...
+
+	// Struct with pointers.
+	type PointerStruct struct {
```
TheSignPainter98: Would it be good to call this `StructWithPointers`?

package_test.go
```diff
@@ -496,6 +496,33 @@ func (s *PackageSuite) TestNulls(c *C) {
...
+		Street   *string         `db:"street"`
+		District *sql.NullString `db:"district"`
+		ID       *int            `db:"id"`
+	}
+
+	// Pointer variables for pointer struct test.
+	var mainStreet = "Main Street"
+	var oneThousand = 1000
```
TheSignPainter98: I notice this in existing code, so this might not be an issue for this PR, but would it be good to use `:=` here? The convention I like to use is
```go
var foo *Bar  // if the zero value of foo is needed
foo := &Bar{} // otherwise
```

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+			})
+		}
+	}
+	return fields, nil
+}
+
 // generate produces and returns reflection information for the input
 // reflect.Value that is specifically required for SQLair operation.
 func generateTypeInfo(t reflect.Type) (typeInfo, error) {
+	var tags []string
```
flotter: It looks to me like ```tags``` are only used for the ```reflect.Struct``` case, so I would not move the variable outside that scope?

internal/expr/typeinfo.go
```diff
@@ -123,31 +176,17 @@ func generateTypeInfo(t reflect.Type) (typeInfo, error) {
...
-				omitEmpty:  omitEmpty,
-				tag:        tag,
-				structType: t,
+		fields, err := getStructFields(t)
+		if err != nil {
+			return nil, err
+		}
+		for _, field := range fields {
+			tags = append(tags, field.tag)
+			if dup, ok := info.tagToField[field.tag]; ok {
```
flotter: Checking my understanding: The only thing we aim to check here is there is more than one ``db:`` tag with the same name?

The error message can be read (which is what I did first, not knowing the context):

```tag %q appears in field %q , and field %q in type %q"``` - with a break before the first "and".

The other issue is we are pretending the user shares the same flattened view we keep in the code, but attributing both fields to the top level struct type.

Is there perhaps a more direct way to point the user at the problem (no is also a valid answer)?

How about something like?

```struct %q definition includes a reuse of tag %q for field %q, previously used for field %q```

File: internal/expr/typeinfo.go [hunk 10]
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+// getStructFields requires the caller to check that t is a struct.
+func getStructFields(t reflect.Type) ([]*structField, error) {
+	var fields []*structField
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if !f.IsExported() {
+			return nil, fmt.Errorf("field %q of struct %s not exported", f.Name, t.Name())
+		}
+		ft := f.Type
+		k := ft.Kind()
```
flotter: I find it slightly difficult to infer what your single letter vars refer to lower down in the code without scrolling back up.

https://google.github.io/styleguide/go/decisions#variable-names

It is not only the fact that ```k``` is for ```kind```, but at which level is it referring to what. ```ft``` here could give a clue, but it does not. I look one line up and see its a type, but of what? Ok fine I can trace ```f``` further up so infer its a field, but a field at which level?

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+		// Check if the field type is an embedded/nested struct or pointer to
+		// one. If a pointer to the field type implementes Scanner then it is
+		// not a nested struct.
+		if (k == reflect.Struct && !reflect.PointerTo(ft).Implements(scannerInterface)) ||
+			(k == reflect.Pointer && ft.Elem().Kind() == reflect.Struct && !ft.Implements(scannerInterface)) {
```
flotter: How about having a helper function for this that takes a field?

```
// ...
fieldIsStruct(field reflect.StructField) bool {
    if (field.Type.Kind() == reflect.Struct && !reflect.PointerTo(field.Type).Implements(scannerInterface)) ||
       (field.Type.Kind() == reflect.Pointer && field.Type.Elem().Kind() == reflect.Struct && !ft.Implements(scannerInterface)) {
        return true
    }
    return false
}
```

This way ```ft``` and ```k``` only has to exist inside the if statement.

```
if fieldIsStruct(field) {
    // The context of this inner scope if now clearly limited to the field
    type := field.Type // This may still be a pointer to the field
    kind := type.Kind()
    
    // If this is a pointer to the field, get the type it points to
    :
} else {
:
}
```

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+			if k == reflect.Pointer {
+				ft = ft.Elem()
+			}
+			nestedFields, err := getStructFields(ft)
+			if err != nil {
+				return nil, err
+			}
+			for _, nestedField := range nestedFields {
```
flotter: Maybe something like this could help others ?

```
// Promote the nested struct fields into the current parent struct scope, but make
// sure we update the Index list to help us navigate back to the original nested
// location for access to the real location.
```

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+			if tag == "" {
+				continue
+			}
+			tag, omitEmpty, err := parseTag(tag)
+			if err != nil {
+				return nil, fmt.Errorf("cannot parse tag for field %s.%s: %s", t.Name(), f.Name, err)
+			}
+			fields = append(fields, &structField{
+				name:       f.Name,
+				index:      f.Index,
```
flotter: If I understand the code correctly, reflect.StructField.Index will only ever contain a single element, in other words, it contains only an index within its immediate parent struct (even though the parent struct may be embedded in another struct)? If this is true, it makes sense why we have to prepend the parent struct Index each time a recursive iteration completes.

internal/expr/typeinfo.go
```diff
@@ -109,9 +112,59 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+		f := t.Field(i)
+		if !f.IsExported() {
+			return nil, fmt.Errorf("field %q of struct %s not exported", f.Name, t.Name())
+		}
+		ft := f.Type
+		k := ft.Kind()
+		// Check if the field type is an embedded/nested struct or pointer to
+		// one. If a pointer to the field type implementes Scanner then it is
+		// not a nested struct.
+		if (k == reflect.Struct && !reflect.PointerTo(ft).Implements(scannerInterface)) ||
```
flotter: Is my understanding correct: Many Go builtin/base types have scanner interface implementations provided by database/sql, but in general, unless explicitly provided, custom structs do not have them implemented? But what happens if a struct type (or pointer to) does provide this interface?

flotter (commented):
Hope some of the comments help. Great work!!

internal/expr/typeinfo.go
Aflynn50: This is a good point, this error message is not super clear. I've changed it to `db tag %q appears in both field %q and field %q of struct %q`.

It would be good if it we said which nested struct they occurred in but its not really worth the extra complexity here. If you are using embedded structs then clashing field names will be flagged by the compiler, its not going to be very common that someone has no clashing field names but does have clashing tags, which is when this error is triggered.

internal/expr/typeinfo.go
Aflynn50: Good point! Changed.

internal/expr/typeinfo.go
Aflynn50: True, it is certainly confusing. I've changed `f` to `field` and `t` to `structType` but I've kept `k` and `ft` because they only have a scope of 6 lines and having those short names makes the huge if statement quite a bit more readable.

internal/expr/typeinfo.go
Aflynn50: That does make it a lot more readable, I've changed it to this. Thanks!

internal/expr/typeinfo.go
Aflynn50: Added comment.

internal/expr/typeinfo.go
Aflynn50: That's correct, the index is built up on the after the return of each recursion.

internal/expr/typeinfo.go
Aflynn50: In that case we don't have to do anything special. We use `Rows.Scan` from `database/sql` to fill the fields of the struct when the time comes. If a field implements scanner, this function will use the interface methods correctly to fill it.

Aflynn50: This is the case for the `sql.NullString`/`sql.NullInt64`/... types that are in some of the tests.

internal/expr/expr_test.go
Aflynn50: Thats a good point, I've changed the name of the main struct to `StructWithEmbedded`.

I think there is a mistake here actually yea. I had intended to have one nested, one nested with a pointer and one embedded but there are two nested. I've changed that.

I've left the `E*` names as they are because it should be fairly clear what is going on.

package_test.go
Aflynn50: Removed space.

internal/expr/expr_test.go
Aflynn50: By flags do you mean tags? Even without the `.*` notation if someone did `DupTags.id` that would still be ambiguous.

I've changed `Dup` to `Ambiguous`.

internal/expr/typeinfo.go
Aflynn50: I've added an explanation comment but I've left that one. I think its still useful to specify this precondition.

package_test.go
Aflynn50: Changed to reflect the types in the other file.

package_test.go

package_test.go
Aflynn50: True, I should stick to this convention. I just need to remember it when I'm actually writing the code haha.

flotter (approved):
Looks good to me. I was new to reflection, so I may have missed something. But from my perspective, and your tests, I think this is good.

internal/expr/typeinfo.go
```diff
@@ -109,6 +112,67 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+	ft := field.Type
+	k := ft.Kind()
+	// Check if it is a
+	return (k == reflect.Struct && !reflect.PointerTo(ft).Implements(scannerInterface)) ||
+		(k == reflect.Pointer && ft.Elem().Kind() == reflect.Struct && !ft.Implements(scannerInterface))
+}
+
+// getStructFields returns relevent reflection information about all struct
+// fields included nested/embedded fields.
+// The caller is required to check that t is a struct.
```
letFunny: * Can we say `structType` instead of `t`?

Additional [Nit]: Can we say that `structType` is required to be a struct instead of talking about the caller checking it?

internal/expr/typeinfo.go
```diff
@@ -109,6 +112,67 @@ func getTypeInfo(value any) (typeInfo, error) {
...
+// struct.
+func fieldIsStruct(field reflect.StructField) bool {
+	ft := field.Type
+	k := ft.Kind()
+	// Check if it is a
+	return (k == reflect.Struct && !reflect.PointerTo(ft).Implements(scannerInterface)) ||
+		(k == reflect.Pointer && ft.Elem().Kind() == reflect.Struct && !ft.Implements(scannerInterface))
+}
+
+// getStructFields returns relevent reflection information about all struct
```
letFunny:
```suggestion
// getStructFields returns relevant reflection information about all struct
```

letFunny (changes requested):
I think we should have a more comprehensive design before going forward with the implementation. As we discussed online, it is not obvious whether we want to promote the fields to the topmost struct in SQL or whether we want to support accessing nested structs in SQL (`Struct.FieldStruct.Field`).

Let's put this on hold.

Aflynn50:
Succeeded by #141

PR #84, opened 2023-06-29
Allow Unicode struct (field) names and column names

Aflynn50:
Change the parser to cycle through the input by rune rather than by byte. Go allows certain Unicode characters to be used in struct names and SQLite allows them to be used in column names, therefore we should allow them in these places too.

Also update the column and struct tag parser to accept column names in quotes, since this is also allowed in SQLite.

The logic of the parser has been kept unchanged when possible.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -668,11 +668,6 @@ func (s *ExprSuite) TestValidQuery(c *C) {
 		[]any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
 		[]any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
 	}, {
-		"SELECT foo FROM t WHERE x = $StringMap.street, y = $Person.id",
-		[]any{Person{}, StringMap{}},
-		[]any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
-		[]any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
-	}, {
 		"SELECT FROM person WHERE id = $Unicode我Struct.საფოსტო",
 		[]any{Unicode我Struct{}},
 		[]any{Unicode我Struct{X人: 30}},
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -39,11 +39,24 @@ type HardMaths struct {
 	Coef int `db:"coef"`
 }
 
+type QuotedColumnNames struct {
+	Ex int `db:"\"!!!\""`
+	Qu int `db:"'???'"`
+}
+
 type M map[string]any
+
+type lowerCaseMap map[string]any
 
 type IntMap map[string]int
 
 type StringMap map[string]string
+
+type Unicode我Struct struct {
+	X人 int    `db:"საფოსტო"`
+	X我 int    `db:"住所"`
+	X这 string `db:"鑑別"`
+}
 
 var tests = []struct {
 	summary          string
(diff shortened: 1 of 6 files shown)


benhoyt:
This looks very reasonable to me. At first I wondered why the lexer/parser itself needed to understand Unicode, but yeah, I don't see any way around it given that we're skipping/parsing Go struct names and the like.

Interestingly, ages ago I switched GoAWK's lexer the other way for performance reasons, from `rune` to `[]byte`, as the AWK grammar doesn't support Unicode identifiers (you can still use Unicode/UTF-8 fine in string literals and comments though, but there it's just skipping over the bytes till it reaches `"` or end of line.) See [here](https://github.com/benhoyt/goawk/commit/0fa32f929b27bc55bcb8d68507853f1083d8ae02) and [here](https://github.com/benhoyt/goawk/commit/43af0cbd2f7b19273b58a75bf0fab20f91a755bf).

Nit: "Compatablilty" is spelt wrong in the PR title. Might be better to have a more explicit title in any case, maybe "Allow Unicode struct (field) names and column names".

internal/expr/parser.go
```diff
@@ -244,32 +275,35 @@ func (p *Parser) skipStringLiteral() (bool, error) {
...
-// end of the string is reached and no matching byte was found, it returns
+// end of the string is reached and no matching char was found, it returns
 // false.
-func (p *Parser) skipByteFind(b byte) bool {
-	for i := p.pos; i < len(p.input); i++ {
-		if p.input[i] == b {
-			p.pos = i + 1
+func (p *Parser) skipCharFind(c rune) bool {
+	cp := p.save()
+	for p.pos < len(p.input) {
```
letFunny, line 298: An idea very open to discussion: what if we make nextChar return a bool and check that in the loops, that way we abstract the fact that we are actually reading from `p.input` at `p.pos`. 

I see that it is used in a lot of places in the code, so if it makes it more complex, I am not sure it would be worth it.

internal/expr/parser.go
```diff
@@ -297,41 +331,16 @@ func (p *Parser) skipBlanks() bool {
 func (p *Parser) skipString(s string) bool {
 	if p.pos+len(s) <= len(p.input) &&
 		strings.EqualFold(p.input[p.pos:p.pos+len(s)], s) {
+		// Manually advance the parser to the end of the string.
 		p.pos += len(s)
+		var size int
+		p.char, size = utf8.DecodeRuneInString(p.input[p.pos:])
+		p.nextPos = p.pos + size
```
letFunny, line 337: Can we instead call p.nextChar() or is there a reason not to?

internal/expr/parser.go
```diff
@@ -344,36 +353,72 @@ func (p *Parser) skipName() bool {
...
+// followed by letters, digits and underscores. This matches the allowed
+// characters in Go type names.
 func (p *Parser) parseIdentifier() (string, bool) {
 	mark := p.pos
-	if p.skipName() {
+
+	if unicode.IsLetter(p.char) || p.char == '_' {
+		p.nextChar()
+		for p.pos < len(p.input) && (unicode.IsLetter(p.char) ||
+			unicode.IsDigit(p.char) || p.char == '_') {
```
letFunny, line 372: [Nit] I think it is clearer if we keep all the statement in one line even if it is long because it improves readability.

internal/expr/parser.go
```diff
@@ -344,36 +353,72 @@ func (p *Parser) skipName() bool {
...
+	mark := p.pos
+
+	if ok, err := p.skipStringLiteral(); err != nil {
+		return "", false, err
+	} else if ok {
+		return p.input[mark:p.pos], true, nil
+	}
+
+	for p.pos < len(p.input) && (unicode.IsLetter(p.char) ||
+		unicode.IsDigit(p.char) || p.char == '_') {
```
letFunny, line 398: [Nit] Keep condition in one line for readability, same as above.

internal/expr/typeinfo.go
```diff
@@ -159,34 +158,33 @@ func generateTypeInfo(t reflect.Type) (typeInfo, error) {
...
 	}
 
-	name := options[0]
+	name := strings.TrimSpace(options[0])
 	if len(name) == 0 {
 		return "", false, fmt.Errorf("empty db tag")
 	}
 
-	if !validColNameRx.MatchString(name) {
+	var p = NewParser()
```
letFunny, line 221: [Nit] Change `var` to `:=p .

internal/expr/typeinfo.go
```diff
@@ -159,34 +158,33 @@ func generateTypeInfo(t reflect.Type) (typeInfo, error) {
...
+	p.init(name)
+	if col, ok, err := p.parseColumnName(); err != nil {
```
letFunny, line 223: Really good idea!

internal/expr/parser.go
letFunny: Additionally, does it make sense to also return the char itself? which would simplify some of the code

internal/expr/expr_test.go
```diff
@@ -618,6 +667,16 @@ func (s *ExprSuite) TestValidQuery(c *C) {
 		[]any{Person{}, StringMap{}},
 		[]any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
 		[]any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
+	}, {
+		"SELECT foo FROM t WHERE x = $StringMap.street, y = $Person.id",
+		[]any{Person{}, StringMap{}},
+		[]any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
+		[]any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
```
letFunny: What is this test testing?

letFunny (commented):
Looking good, left a few comments, most of them minor

internal/expr/expr_test.go
Aflynn50: Looks like a duplicate test added by mistake. Removed.

niemeyer (changes requested):
Sorry for being the source of bad news, but can we please not go there just yet?  We have lots of issues to address and logic to improve, but SQL+Go code using non-ASCII chars in their type names and field names is not one of them right now.

PR #85, opened 2023-07-07
Fix: GetAll not accepting slices of maps

Aflynn50:
This bugfix adds support to `GetAll` for queries involving maps. This was missed in the maps PR. Currently, `GetAll` gives an error saying it only accepts slices of structs.

This PR also adds a few more cases to the `Get` tests to verify support for maps and tides up/minorly corrects some other tests relating to maps.


niemeyer (commented):
Thanks for the fix.

niemeyer (approved):

PR #86, opened 2023-07-07
Error on unexpected output in `GetAll`

Aflynn50:
The PR adds a check in `GetAll` that triggers an error if the query has no output expressions but arguments were passed to `GetAll`. This PR brings the behaviour of `GetAll` in line with `Get` which already implements this check.

This error is useful if a user has not realised that all results in a query need to be inside output expressions and write something like this:
```SQL
SELECT name FROM person
```
In SQLair this will be executed wit `sql.Exec` and no results will be available since it contains no output expressions.


niemeyer (approved):
Thanks!

PR #87, opened 2023-07-14
Enforce matching bracketing in output expressions

Aflynn50:
This PR changes the syntax of output expressions. For output expressions with columns both sides now need to be unbracketed, or bracketed.

This PR is intended to ensure that the syntax stays consistent. This requirement may be relaxed in the future.

Now allowed:
- `(name, id) AS (&Person.*)`
- `(name) AS (&Person.*)`
- `name AS &Person.*`

Now not allowed:
- `(name, id) AS &Person.*`
- `(name) AS &Person.*`
- `name AS (&Person.*)`


niemeyer (approved):
Looks good, thanks!

PR #88, opened 2023-07-17
Error on struct with no tags

Aflynn50:
This PR adds in a check for output expressions that use an asterisk, e.g. `&MyStruct.*`. If the struct has no `db` tags then an error will be thrown. Currently, the statement `SELECT &MyStruct.* FROM t` would result in the SQL `SELECT  FROM t` which would trigger a parsing error from the database.


niemeyer (approved):
Nice and clean as well, thanks.

PR #89, opened 2023-07-17
Error on missing output columns

Aflynn50:
Add an error triggered when columns from output expressions do not appear in query results. This could happen when a output expressions is within a `WITH` clause of a query.

Also refactor `TestIterGetErrors` to be more concise and reliable. Before, if no error was thrown by `iter.Get` the test could still pass.

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -199,7 +199,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 
 	for i := 0; i < len(qe.outputs); i++ {
 		if !columnInResult[i] {
-			return nil, nil, fmt.Errorf("column %q for struct %q not found in results", qe.outputs[i].memberName(), qe.outputs[i].outerType().Name())
+			return nil, nil, fmt.Errorf(`query uses "&%s" outside of result context`, qe.outputs[i].outerType().Name())
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


package_test.go
```diff
@@ -274,64 +274,77 @@ func (s *PackageSuite) TestIterGetErrors(c *C) {
...
+		        (SELECT &Person.id
+		        FROM person)
+		        SELECT id
+		        FROM person, averageID
+		        WHERE id > averageID.avgid
+				LIMIT 1`,
+		types:   []any{Person{}},
+		inputs:  []any{},
+		outputs: []any{&Person{}},
+		err:     `cannot get result: column "id" for struct "Person" not found in results`,
```
niemeyer: As we discussed, the error here is unclear since the real issue is the &Person.id is being used in a wrong context. Maybe something along the lines of:
```
cannot get result: query uses &Person.id outside of result context
```
Also, FWIW the query right above seems broken a bit too much.

niemeyer (changes requested):

package_test.go

letFunny (approved):

internal/expr/query.go
```diff
@@ -151,6 +151,7 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 
 	// Generate the pointers.
 	var ptrs = []any{}
+	var columnInResult = make(map[int]bool)
```
niemeyer: Sorry for missing this earlier, but this can be just a simple slice:
```
var columnInResult = make([]bool, len(qe.outputs))
```

niemeyer (approved):
One detail, and LGTM

internal/expr/query.go
Aflynn50: Good point, changed from map to slice.

niemeyer (approved):
Thanks!

PR #90, opened 2023-07-25
Improve comments in `parser.go`

Aflynn50:
Correct and improve on the content of some of the comments in `parser.go`. Also limit the line width to 80 columns in the comments (assuming you have tabs set to 4 spaces).


internal/expr/parser.go
```diff
@@ -487,8 +487,8 @@ func (p *Parser) parseTargetTypes() (types []fullName, parentheses bool, ok bool
 	return nil, false, false, nil
 }
 
-// parseOutputExpression requires that the ampersand before the identifiers must
-// be followed by a name byte.
+// parseOutputExpression parsers a SQLair output expression. See the package
```
niemeyer: Typo ("parsers"). This comment can be dropped entirely, though, since it's literally reading the function name.

internal/expr/parser.go
```diff
@@ -541,7 +541,10 @@ func (p *Parser) parseInputExpression() (*inputPart, bool, error) {
 			if fn.name == "*" {
 				return nil, false, fmt.Errorf(`asterisk not allowed in input expression "$%s"`, fn)
 			}
-			return &inputPart{sourceType: fn, raw: p.input[cp.pos:p.pos]}, true, nil
+			return &inputPart{
+				sourceType: fn,
+				raw:        p.input[cp.pos:p.pos],
+			}, true, nil
```
niemeyer: FWIW, this seems slightly less readable for some reason, probably because it's adding too much spacing in what's a small amount of code, and at the same time moving away the other results being returned (true and nil).

Compare that with this one, for instance:
```
part := &inputPart{sourceType: rn, raw: p.input[...]}
return part, true, nil
```

internal/expr/parser.go
```diff
@@ -333,30 +337,27 @@ func (p *Parser) skipName() bool {
...
 		return "*", true
 	}
 	return p.parseIdentifier()
 }
 
-// parseIdentifier parses a name made up of only nameBytes. On success it
-// returns the parsed string and true. Otherwise, it returns the empty string
-// and false.
+// parseIdentifier parses a name made up of only name bytes.
 func (p *Parser) parseIdentifier() (string, bool) {
```
niemeyer, line 470: Something a bit curious here. What's the difference between a name and an identifier?  If there's none, we shoudn't have two names for them. If there is some, we should define what an identifier is in a way that is not just referring to it being a name.

internal/expr/parser.go
```diff
@@ -6,17 +6,20 @@ import (
...
 	// prevPart is the value of pos when we last finished parsing a part.
 	prevPart int
 	// partStart is the value of pos just before we started parsing the part
-	// under pos. We maintain partStart >= prevPart.
+	// under pos. Invariant: partStart >= prevPart.
 	partStart int
 	parts     []queryPart
 }
 
+// NewParser returns an uninitiated Parser.
```
niemeyer: What is an uninitialized Parser, in terms of the public API of this package?  Isn't it just NewParser => Parse?

internal/expr/parser.go
```diff
@@ -6,17 +6,20 @@ import (
 	"strings"
 )
 
+// Parser is an object for parsing SQL queries containing SQLair expressions.
+// It is used for generating a ParsedExpr with the Parse method.
```
niemeyer: In general we don't document the methods on the type, since they have their own documentation. There are exceptions for types which need some handholding due to their complexity, but that doesn't seem to be the case here.

niemeyer (changes requested):
Thanks, lots of improvements there.

Only a few details.. mostly low importance nitpicks. Blocking mainly due to the documentation of the package external interface being a bit misleading (see the "uninitialized" bit).

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: That's true, in terms of the API it doesn't mean much. I've removed this comment as there's really nothing to say in it that is not clear from the function name.

internal/expr/parser.go
Aflynn50: Removed comment.

internal/expr/parser.go
Aflynn50: There is none, I've added a separate PR to change it #109.

internal/expr/parser.go
Aflynn50: Changed it back to being a one liner.

PR #91, opened 2023-07-25
Improve comments in `prepare.go`

Aflynn50:
Correct and improve on the content of some of the comments in `prepare.go`. Also limit the line width to 80 columns in the comments (assuming you have tabs set to 4 spaces).


internal/expr/prepare.go
```diff
@@ -306,5 +317,9 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
 		}
 	}
 
-	return &PreparedExpr{inputs: inputs, outputs: outputs, sql: sql.String()}, nil
+	return &PreparedExpr{
+		inputs:  inputs,
+		outputs: outputs,
+		sql:     sql.String(),
+	}, nil
```
niemeyer: Same comment as the previous PR. Let's please not establish this as a pattern.

Do you find it inconvenient to read these short one-liners?

internal/expr/prepare.go
```diff
@@ -213,10 +223,9 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []fullName, typeMe
 
 type typeNameToInfo map[string]typeInfo
 
-// Prepare takes a parsed expression and struct instantiations of all the types
-// mentioned in it.
-// The IO parts of the statement are checked for validity against the types
-// and expanded if necessary.
+// Prepare takes a parsed expression and samples of all the types mentioned in
+// its SQLair expressions. The input and output parts of the statement are
+// validated against the sample types and columns are expanded where necessary.
```
niemeyer, line 207: Much better comment, thanks.

internal/expr/prepare.go
```diff
@@ -50,6 +51,9 @@ func getKeys[T any](m map[string]T) []string {
 	return keys
 }
 
+// starCount counts the number of fullNames in the list with asterisks in the
+// name field. It is used to check if there are any asterisk expressions in a
+// list of types or columns.
```
niemeyer:
```
// starCount counts the number of fullName values that have "*" as a name.
```
Note:
- fullName is a type name, so we can't make it plural.
- The "in" was misleading: it must _be_ an astarisk rather than _contain_ one.
- It's an anti-pattern to document who calls a function and why. We document what it does instead, and whoever needs it can call it..

internal/expr/prepare.go
```diff
@@ -9,7 +9,7 @@ import (
 	"strings"
 )
 
-// PreparedExpr contains an SQL expression that is ready for execution.
+// PreparedExpr represents a SQL statement for use in a query.
```
niemeyer: The original comment, while not a great comment indeed, was hinting at the expression pipeline composed by ParsedExpr, PreparedExpr, QueryExpr. The new comment loses that, and doesn't really improve the understanding much (all of those are SQL statements for use in a query, somehow).  Can we have a look at those three and try to improve things a bit further?

niemeyer (changes requested):

internal/expr/prepare.go
Aflynn50: I personally prefer the multi-line style, I find it easier to read with the indentation, but I will keep it as a one liner for consistency.

internal/expr/prepare.go
Aflynn50: Good points, changed.

internal/expr/prepare.go
Aflynn50: I've changed all three. This is how they currently look:
```
// ParsedExpr represents a parsed SQL string containing SQLair input and output
// expressions.
```
```
// PreparedExpr is a ParsedExpr that has been validated and processed into SQL
// ready for use in a query.
```
```
// QueryExpr represents a SQL statement and its arguments. It is is ready for
// execution on a database and then the scanning of the query results.
```
The other two comments are in PR #90 and PR #93 respectively.

PR #94, opened 2023-07-26
Improve comments in `sqlair.go`

Aflynn50:
Correct and improve on the content of some of the comments in `sqlair.go`. The line width is also limited to 80 columns in  comments (assuming you have tabs set to 4 spaces).

The example in the comment above the `M` type has been removed and an example of using `M` has been added to PR #70.

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -13,15 +13,19 @@ import (
 	"github.com/canonical/sqlair/internal/expr"
 )
 
-// M is a type that, as with other map types, can be used with SQLair for more dynamic behavior.
-// It can be used in querys to pass arbitrary values referenced by their key.
+// M is a convenience type that can be used in input and output expressions to
+// pass values referenced by their key. M is not a special type, any named map
+// type with string keys can be used with SQLair.
 //
-// For example:
+// Example 1:
+//     stmt := sqlair.MustPrepare("UPDATE people SET name=$M.name", sqlair.M{})
+//     err := db.Query(ctx, stmt, sqlair.M{"id": 10}).Run()
 //
-//	stmt := sqlair.MustPrepare("SELECT (name, postcode) AS &M.* FROM p WHERE id = $M.id", sqlair.M{})
-//	q := db.Query(ctx, stmt, sqlair.M{"id": 10})
-//	var resultMap = sqlair.M{}
-//	err := q.Get(resultMap) // => sqlair.M{"name": "Fred", "postcode": 10031}
+// Example 2:
+//     stmt := sqlair.MustPrepare("SELECT (name, postcode) AS &M.* FROM people id", sqlair.M{})
+//     m := sqlair.M{}
+//     err := db.Query(ctx, stmt).Get(m) // => sqlair.M{"name": "Fred", "postcode": 10031}
+
 type M map[string]any
 
 // S is a slice type that, as with other named slice types, can be used with
@@ -31,8 +35,8 @@ type S []any
 var ErrNoRows = sql.ErrNoRows
 var ErrTXDone = sql.ErrTxDone
 
-// Statement represents a SQL statement with valid SQLair expressions.
-// It is ready to be run on a SQLair DB.
+// Statement represents a parsed SQLair statement ready to be run on a database.
+// A statement can be used with any database.
 type Statement struct {
 	// te is the type bound SQLair query. It contains information used to
 	// generate query values from the input arguments when the Statement is run
@@ -40,10 +44,10 @@ type Statement struct {
 	te *expr.TypeBoundExpr
 }
 
-// Prepare expands the types mentioned in the SQLair expressions and checks
-// the SQLair parts of the query are well formed.
+// Prepare validates SQLair expressions in the query and generates a SQLair
+// statement Statement.
 // typeSamples must contain an instance of every type mentioned in the
-// SQLair expressions of the query. These are used only for type information.
+// SQLair expressions in the query. These are used only for type information.
 func Prepare(query string, typeSamples ...any) (*Statement, error) {
 	parser := expr.NewParser()
 	parsedExpr, err := parser.Parse(query)
@@ -58,7 +62,7 @@ func Prepare(query string, typeSamples ...any) (*Statement, error) {
 	return &Statement{te: typedExpr}, nil
 }
 
-// MustPrepare is the same as prepare except that it panics on error.
+// MustPrepare is the same as Prepare except that it panics on error.
 func MustPrepare(query string, typeSamples ...any) *Statement {
 	s, err := Prepare(query, typeSamples...)
 	if err != nil {
@@ -81,7 +85,7 @@ func (db *DB) PlainDB() *sql.DB {
 	return db.sqldb
 }
 
-// Query holds the results of a database query.
+// Query represents a query on a database. It is designed to be run once.
 type Query struct {
 	// run executes the Query against the db or the tx.
 	run func(context.Context) (*sql.Rows, sql.Result, error)
@@ -100,8 +104,9 @@ type Iterator struct {
 	started bool
 }
 
-// Query takes a context, prepared SQLair Statement and the structs mentioned in the query arguments.
-// It returns a Query object for iterating over the results.
+// Query builds a new query from a context, a SQLair Statement and the input
+// arguments. The query is run on the database when one of Iter, Run, Get or
+// GetAll is executed on the Query.
 func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query {
 	if ctx == nil {
 		ctx = context.Background()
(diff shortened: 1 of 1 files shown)


sqlair.go
```diff
@@ -13,33 +13,27 @@ import (
...
-// M is a type that, as with other map types, can be used with SQLair for more dynamic behavior.
-// It can be used in querys to pass arbitrary values referenced by their key.
-//
-// For example:
-//
-//	stmt := sqlair.MustPrepare("SELECT (name, postcode) AS &M.* FROM p WHERE id = $M.id", sqlair.M{})
-//	q := db.Query(ctx, stmt, sqlair.M{"id": 10})
-//	var resultMap = sqlair.M{}
-//	err := q.Get(resultMap) // => sqlair.M{"name": "Fred", "postcode": 10031}
+// M is a type that, as with other map types, can be used in SQLair input and
```
manadart: "M is a convenience type for values indexed by strings. It can be used in both input and output expressions." is probably all you need.

I would include the example, but unlike the one removed, I'd have 2 simple prepare examples, one with input and one without output.

sqlair.go
```diff
@@ -13,33 +13,27 @@ import (
...
+// output expressions. It can be used in queries to pass arbitrary values
+// referenced by their key.
 type M map[string]any
 
 var ErrNoRows = sql.ErrNoRows
 var ErrTXDone = sql.ErrTxDone
 
-// Statement represents a SQL statement with valid SQLair expressions.
-// It is ready to be run on a SQLair DB.
+// Statement represents a verified SQLair statement ready to be run on a DB.
```
manadart: "represents a _parsed_ SQLair statement".

I prefer to use plain English in comments, which would prefer "database" over "DB".

sqlair.go
```diff
@@ -86,7 +80,7 @@ type Query struct {
 	pq  *expr.PrimedQuery
 }
 
-// Iterator is used to iterate over the results of the query.
+// Iterator is used to iterate over the results of the query row by row.
```
manadart: "row by row" doesn't add anything here.

sqlair.go
```diff
@@ -296,6 +303,7 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 		sliceVals = append(sliceVals, sliceVal)
 	}
 
+	// Iterate through the query results.
```
manadart: "Iterate _over_".

sqlair.go
Aflynn50: Removed.

sqlair.go

sqlair.go
Aflynn50: I've changed the comment to be less verbose, though I want to still express that `M` is not a special type in any way i.e. that any map type can be used with SQLair.

I have been intending to do a separate PR with examples in it. I'll put some in the comment here and then remove them when I do that in another PR.

sqlair.go
```diff
@@ -81,7 +85,7 @@ func (db *DB) PlainDB() *sql.DB {
 	return db.sqldb
 }
 
-// Query holds the results of a database query.
+// Query represents a query on a database. It is designed to be run once.
```
letFunny, line 88: Can you only run it once though? I would assume you could pass in the values, store the query, and run it as much times as you want (think a select statement).

sqlair.go
```diff
@@ -100,8 +104,9 @@ type Iterator struct {
 	started bool
 }
 
-// Query takes a context, prepared SQLair Statement and the structs mentioned in the query arguments.
-// It returns a Query object for iterating over the results.
+// Query builds a new query from a context, a SQLair Statement and the input
+// arguments. The query is run on the database when one of Iter, Run, Get or
+// GetAll is executed on the Query.
```
letFunny: Question about go doc, is this going to show up as links in the documentation or do you have to use [sqlair.Iter], [sqlair.Run], etc.

sqlair.go
```diff
@@ -124,14 +129,18 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
 	return &Query{pq: pq, run: run, ctx: ctx, err: nil}
 }
 
-// Run is an alias for Get that takes no arguments.
+// Run is an alias for Get that takes no arguments. Run is used to run a query
```
letFunny: `Run is an alias for Get that takes no arguments.`  <- Do you think this is an implementation detail? I would just keep the second sentence.

sqlair.go
```diff
@@ -124,14 +129,18 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+// on a database and disregard any results.
 func (q *Query) Run() error {
 	return q.Get()
 }
 
-// Get runs the query and decodes the first result into the provided output arguments.
-// It returns ErrNoRows if output arguments were provided but no results were found.
-// An &Outcome{} variable may be provided as the first output variable.
+// Get runs the query and decodes the first result into the provided output
```
letFunny:
```suggestion
// Get runs the query and decodes the first row returned into the provided output
```

letFunny: I think result is normally used to get the statistics like when running `sql.Exec`.

sqlair.go
```diff
@@ -124,14 +129,18 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+// arguments. It returns ErrNoRows if output arguments were provided but no
+// results were found.
+//
+// An Outcome struct may be provided as the first output variable to fill it
```
letFunny: Should we say pointer to outcome struct? or is it enough to see the comment plus examples?

sqlair.go
```diff
@@ -169,6 +178,7 @@ func (q *Query) Get(outputArgs ...any) error {
 }
 
 // Iter returns an Iterator to iterate through the results row by row.
+// Close must be run once iteration is finished.
```
letFunny:
```suggestion
// Iterator.Close must be run once iteration is finished.
```
The same comment about godoc and picking up names above applies here. Whatever Intellij highlights should work.

sqlair.go
```diff
@@ -189,7 +199,7 @@ func (q *Query) Iter() *Iterator {
 }
 
 // Next prepares the next row for Get.
-// If an error occurs during iteration it will be returned with Iter.Close().
+// If an error occurs during iteration it will be returned with Iter.Close.
```
letFunny:
```suggestion
// If an error occurs during iteration it will be returned with Iterator.Close.
```

sqlair.go
```diff
@@ -198,8 +208,11 @@ func (iter *Iterator) Next() bool {
 	return iter.rows.Next()
 }
 
-// Get decodes the result from the previous Next call into the provided output arguments.
-// An &Outcome{} variable may be provided as the single output variable before the first call to Next.
+// Get decodes the result from the previous Next call into the provided output
+// arguments.
+//
+// Before the first call of Next an Outcome struct may be passed to Get as the
```
letFunny: Same comment about pointer to Outcome.

sqlair.go
```diff
@@ -250,18 +265,22 @@ func (iter *Iterator) Close() error {
...
+// Result returns a sql.Result containing information about the query
+// execution. If no result is set then Result returns nil.
 func (o *Outcome) Result() sql.Result {
 	return o.result
 }
 
 // GetAll iterates over the query and scans all rows into the provided slices.
 // sliceArgs must contain pointers to slices of each of the output types.
-// An &Outcome{} variable may be provided as the first output variable.
+// An &Outcome{} struct may be provided as the first output variable to get
```
letFunny: I think we should be consistent with "Outcome struct" / "&Outcome{}" / "pointer to ..."

sqlair.go
```diff
@@ -373,7 +394,8 @@ func (tx *TX) setDone() error {
 	return nil
 }
 
-// Begin starts a transaction.
+// Begin starts a transaction. A transaction must be ended
+// with a Commit or Rollback.
```
letFunny: Same comment about godoc.

letFunny (commented):
Great addition, thanks! I only have a bunch of nitpicks :P

sqlair.go
Aflynn50: You can run it multiple times, but it should only be run once, mainly because we store the context in the query so it should be used straight away. It being designed to be run once was something Gustavo and I decided ages ago.

sqlair.go
Aflynn50: Oh nice, didn't know you could link like this, have added it all over the place.

sqlair.go
Aflynn50: It is an implementation detail, but one the user should know about, it makes it obvious whats going on and is very easy to understand. I've swapped the sentences.

sqlair.go

sqlair.go
letFunny: Right, then keep it as is.

sqlair.go
letFunny: The only problem would be if we decide to change the inner working of the function in the future because `Run is an alias for [Query.Get] that takes no arguments` involves much more than `Run is used to run a query on a database and disregard any results.`

letFunny (approved):
Thanks! Definitely a step on the right direction, the documentation is starting to look much better.

PR #97, opened 2023-08-30
Add support for basic types

Aflynn50:
Add support for numeric, boolean and string types for use in input and output expressions. Specifically the types: `string`, `bool`, `uint`, `uint8`, `uint16`, `uint32`, `uint64`, `int`, `int8`, `int16`, `int32`, `int64`, `float32`, `float64`.

The types `complex64` and `complex128` are not supported by the `Scan` method of `database/sql` and I see no strong need to add them here. The type `byte` and `rune` are aliases for `uint8` and `int32` respectively and cannot be distinguished from the aliases using the reflect library so are not added here.

This change will allow queries such as `SELECT name AS &string FROM ...` and `SELECT id AS &int FROM ...`. Previously these queries would have required a struct or map to extract the result into.

A limitation of being able to reference primitive types in I/O expressions is that each type can only be referenced once in the inputs and once in the outputs. For example the query `SELECT name AS &string, address AS &string FROM ...` is not valid. This is because when the user calls `db.Query(...).Get(str1, str2)` there is no way for SQLair to tell which column corrosponds to which variable. To get around this one can define new types that have the same underlying type.
For example:
```go
type nameStr string
type addressStr string
stmt, err := sqlair.Prepare(
    "SELECT name AS &nameStr, address AS &addressStr FROM ...",
    (nameStr)(""),
    (addressStr)(""),
)
var name nameStr
var address addressStr
err := db.Query(ctx, stmt).Get(&name, &address)
```

This PR also includes #98 which fixes a bug in detecting duplicate types.


jameinel:
From a conceptual perspective, I don't have any problem with the awkwardness around multiple base types. And this is because you really want to be using mappings that have more meaning to them.
Put a different way, you could write a function that is `myfunc(bool, bool, bool)`, but that is usually a terrible function, because for an end user what is the difference between `myfunc(true, true, false)` and `myfunc(true, true, true)`. You almost always would rather have:
```
is_open := true
check_size := true
verbose := false
myfunc(is_open, check_size, verbose)
```

Similarly here, having a SQL statement that just returns a 'string' makes it hard to be clear that you've lined them up correctly.
```
SELECT name, address FROM people;
db.Query().Get(&name, &address)
```

With only plain string types it would be:
```
SELECT name as &string, address as &string FROM people;
```
But both of these lines work without errors:
```
db.Query().Get(&name, &address)
db.Query().Get(&address, &name)
```
So it is clearly better to have
```
type address string
type name string
...
SELECT name as &name, address as &address
```

I think it makes sense to have a fast path for "I just have 1 string value to return", but as soon as there is confusion, it is better to be clear rather than rely on position to disambiguate.

In fact, I would probably document the type aliasing with something more realistic than 'string1' and 'string2' because it is fairly likely that people might actually think they should be using those integer variants.

internal/expr/expr_test.go
```diff
@@ -307,6 +311,24 @@ AND z = @sqlair_0 -- The line with $Person.id on it
 	"[Bypass[INSERT INTO arr VALUES (ARRAY[[1,2],[] Input[HardMaths.x] Bypass[,4]], ARRAY[[5,6],[] Input[HardMaths.y] Bypass[,8]]);]]",
 	[]any{HardMaths{}},
 	"INSERT INTO arr VALUES (ARRAY[[1,2],[@sqlair_0,4]], ARRAY[[5,6],[@sqlair_1,8]]);",
+}, {
+	"all primitive types",
+	"SELECT (col1, col2, col3, col4, col5, col6, col7, col8, col9, col10, col11, col12, col13, col14) AS (&string, &bool, &uint, &uint8, &uint16, &uint32, &uint64, &int, &int8, &int16, &int32, &int64, &float32, &float64) FROM t",
```
jameinel: This test is actually one that clearly demonstrates my concern.
What is the type of `col8`, how long does it take you to figure out whether it is actually a correct statement, and that you are actually parsing the right value into the correct field.

If you even only did:
```
SELECT (col1 AS &string), (col2 AS &bool)...
```
you would have much more locality and a better understanding of whether the statement is actually correct. Put another way, can you tell that this statement is wrong:
```
"SELECT (col1, col2, col3, col4, col5, col6, col7, col8, col9, col10, col11, col12, col13, col14) AS (&string, &bool, &uint, &uint, &uint8, &uint16, &uint32, &uint64, &int, &int8, &int16, &int32, &int64, &float32, &float64) FROM t"
```

internal/expr/expr_test.go
```diff
@@ -510,6 +536,54 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
+		prepareArgs: []any{},
+		err:         `cannot prepare expression: explicit columns required for primitive type e.g. "col AS &string"`,
+	}, {
+		query:       "SELECT &string.* FROM t",
+		prepareArgs: []any{},
+		err:         `cannot prepare expression: explicit columns required for primitive type e.g. "col AS &string"`,
+	}, {
+		query:       "SELECT &string.name FROM t",
+		prepareArgs: []any{},
+		err:         `cannot prepare expression: explicit columns required for primitive type e.g. "col AS &string"`,
```
jameinel: I think for this case you actually wanted to test:
```"SELECT name AS &string.name```
Because you are already testing 'explicit columns' elsewhere, but you aren't testing member of a primitive type as a target. Probably the `&string.*` is also not quite right.

File: internal/expr/expr_test.go [hunk 3]
```diff
@@ -618,6 +692,11 @@ func (s *ExprSuite) TestValidQuery(c *C) {
 		[]any{Person{}, StringMap{}},
 		[]any{Person{ID: 666}, StringMap{"street": "Highway to Hell"}},
 		[]any{sql.Named("sqlair_0", "Highway to Hell"), sql.Named("sqlair_1", 666)},
+	}, {
+		"SELECT foo FROM t WHERE x = $string, y = $int",
+		[]any{0, ""},
+		[]any{"arg1", 2},
+		[]any{sql.Named("sqlair_0", "arg1"), sql.Named("sqlair_1", 2)},
```
jameinel, line 723: Should you be allowed to use `Named("sqlair_0", ...)` ?
That feels (to me) like it goes against the whole point of Named. Named is that you are mapping the human name present in the SQL that was supplied to an input parameter. But this suddenly changes it that "positional args also have names".
Especially what does it mean if you did:
```[]any{sql.Named("myparam", "arg1"), sql.Named("sqlair_0", 2)}```
How do you know what parameter `sqlair_0` is if it is positional, or named, or?

File: internal/expr/prepare.go [hunk 4]
```diff
@@ -113,17 +124,27 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
 		}
 		return info, nil
 	}
-
-	addColumns := func(info typeInfo, tag string, column fullName) error {
+	addColumns := func(info typeInfo, member string, column fullName) error {
 		var tm typeMember
 		switch info := info.(type) {
```
jameinel: I'm not a big fan of switch ...(type), vs using interfaces that have defined interactions.
eg:
```
type TypeMember interface {
  TypeMember(member string) (Res, error)
}

func (simpleType) TypeMember(member string) {
  if member != "" {
    return nil, fmt.Errorf(`cannot specify member of primitive type...')
  }
  return simpleType{...}, nil
}
```

That makes it far easier to extend your functionality in the future, because you can implement the interfaces, without having to change the hard coded switch statements everywhere.

jameinel (commented):
I realize I'm coming into this a bit late, so I don't want my opinions to overwhelm things, but I do feel there are a couple parts that seem a bit awkward.

internal/expr/expr_test.go
Aflynn50: That's very fair, I had written this in the less readable form to make the test more concise. You make a very good point that people will use these tests to get a feel for how to use the new constructs. I have changed it to the other form.

internal/expr/expr_test.go
Aflynn50: True, I've added the intended test as well.

internal/expr/expr_test.go
Aflynn50: So this tests the generated arguments that are designed to be passed to the function. It is part of the public api of the `expr` package but the results of this function are not exposed at all by the `sqlair` package, they are passed directly to the database.

Using named parameters in the way you suggest above is not at all supported by SQLair. The line you refer to above is the expected output of this function, it only ever generates arguments with the prefix "_sqlair_".

internal/expr/prepare.go
Aflynn50: Very good point, I've refactored as suggested.

internal/expr/expr_test.go
```diff
@@ -307,6 +311,24 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+	"[Bypass[SELECT (] Output[[stringCol] [string]] Bypass[), (] Output[[boolCol] [bool]] Bypass[), (] Output[[uintCol] [uint]] Bypass[), (] Output[[uint8Col] [uint8]] Bypass[), (] Output[[uint16Col] [uint16]] Bypass[), (] Output[[uint32Col] [uint32]] Bypass[), (] Output[[uint64Col] [uint64]] Bypass[), (] Output[[intCol] [int]] Bypass[), (] Output[[int8Col] [int8]] Bypass[), (] Output[[int16Col] [int16]] Bypass[), (] Output[[int32Col] [int32]] Bypass[), (] Output[[int64Col] [int64]] Bypass[), (] Output[[float32Col] [float32]] Bypass[), (] Output[[float64Col] [float64]] Bypass[) FROM t]]",
+	[]any{},
+	"SELECT (stringCol AS _sqlair_0), (boolCol AS _sqlair_1), (uintCol AS _sqlair_2), (uint8Col AS _sqlair_3), (uint16Col AS _sqlair_4), (uint32Col AS _sqlair_5), (uint64Col AS _sqlair_6), (intCol AS _sqlair_7), (int8Col AS _sqlair_8), (int16Col AS _sqlair_9), (int32Col AS _sqlair_10), (int64Col AS _sqlair_11), (float32Col AS _sqlair_12), (float64Col AS _sqlair_13) FROM t",
+}, {
+	"custom primitive types",
+	"SELECT name AS &nameStr, address AS &addressStr, id AS &idInt FROM t WHERE name = $nameStr AND address = $addressStr AND id = $idInt",
```
jameinel, line 322: Interesting that you can have nameStr as both an input and an output.

Aflynn50: Yea, the same applies to the builtin types. It may make it a little harder to read but it is hopefully convenient enough to negate that.

internal/expr/expr_test.go
```diff
@@ -442,6 +464,10 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 		query:       "SELECT (&Person.*, &Person.*) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
 		err:         `cannot prepare expression: member "address_id" of type "Person" appears more than once`,
+	}, {
+		query:       "SELECT (&M.id, &M.id) FROM t",
+		prepareArgs: []any{sqlair.M{}},
+		err:         `cannot prepare expression: member "id" of type "M" appears more than once`,
```
letFunny: Is the intention here to disallow select with two columns of the same name? It seems that it is supported by the dbs at least, see https://www.db-fiddle.com/f/u237xG4h57xXiRRuJTQmeW/1.

internal/expr/parser.go
```diff
@@ -401,7 +401,7 @@ func (p *Parser) parseGoFullName() (fullName, bool, error) {
 
 	if id, ok := p.parseIdentifier(); ok {
 		if !p.skipByte('.') {
-			return fullName{}, false, fmt.Errorf("column %d: unqualified type, expected %s.* or %s.<db tag>", p.pos, id, id)
+			return fullName{prefix: id}, true, nil
```
letFunny: Should we update the comment above the function to reflect the fact that we are no longer parsing `&TypeName.col_name` but now we also allow `&TypeName`. If you update it, maybe we should clarify that this function does not parse the `&` or the `$` but the identifier that follows.

internal/expr/typeinfo.go
```diff
@@ -9,17 +9,31 @@ import (
 	"sync"
 )
 
+// typeMember should be implemented without pointer recivers as it is used as a
```
letFunny: Minor nitpick, the spelling here for "receivers"

internal/expr/expr_test.go
Aflynn50: This is actually added as part of the [bugfix here](https://github.com/canonical/sqlair/pull/98) which is included in this PR. We throw an error here because even though its perfectly possible to select the same column twice, here it being read into the same place twice. This would have no effect so is almost certainly a mistake, hence the error.

internal/expr/parser.go
Aflynn50: Fixed the comment.

internal/expr/typeinfo.go
Aflynn50: Good spot!

doc.go
```diff
@@ -1,8 +1,11 @@
 /*
-SQLair is a convenience layer for SQL databases that embeds Go structs and maps directly into SQL queries.
+SQLair is a convenience layer for SQL databases that embeds Go types directly into SQL queries.
 
-The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query that correspond to Go structs or maps.
-This allows the user to specify the Go structs/maps they want in the SQL query itself whilst allowing the full power of SQL to be utilised.
+The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query that correspond to Go types.
+This allows the user to specify the Go types they want in the SQL query itself whilst allowing the full power of SQL to be utilised.
+The Go types in the input and output expressions can be structs, maps or primitive types.
+SQLair is designed around for use with the struct type.
```
letFunny: I don't understand this sentence, are we saying struct types are more special?

doc.go
```diff
@@ -1,8 +1,11 @@
...
+Struct fields are tagged with column names allowing query results to be read directly into them.
```
letFunny: Should we remove this from the general introduction? I think it makes more sense to talk about it when we are present structs later where we already say:
> Note that in the SQLair `db` tags (i.e. the column names) appear in the input/output expressions, not the field names.

doc.go
```diff
@@ -39,27 +42,32 @@ SQLair input expressions take the format:
...
-    - This form cannot be used with maps.
+    - This form cannot be used with maps or primitive types.
 
  3. table.* AS &Type.*
     - Does the same as 2 but prepends all columns with the table name.
 
  4. (t1.col_name1, t2.col_name2) AS &Type.*
     - Fetches and sets only the specified columns (the table is optional).
     - If Type is a map they will be stored at "col_name1" and "col_name2".
+	- This form cannot be used with primitive types.
```
letFunny: Spacing is inconsistent with previous options.

doc.go
```diff
@@ -39,27 +42,32 @@ SQLair input expressions take the format:
...
 
  5. (col_name1, col_name2) AS (&Type.other_col1, &Type.other_col2)
     - Fetches the columns from the database and stores them at other_col1 and other_col2 in Type.
+	- If Type is a primitive type then the "other_col" is dropped e.g. "(col_name AS &string)".
```
letFunny: Spacing is inconsistent with previous options.

internal/expr/expr_test.go
```diff
@@ -391,26 +410,21 @@ comment */ WHERE x = $Address.&d`,
...
 		query: `SELECT (name, id) AS (&Person.name, /* multiline
 comment */
 
 &Person.id`,
 		err: `cannot parse expression: line 1, column 22: missing closing parentheses`,
-	}, {
-		query: `SELECT (name, id) WHERE name = 'multiline
-string
-of three lines' AND id = $Person.*`,
-		err: `cannot parse expression: line 3, column 26: asterisk not allowed in input expression "$Person.*"`,
```
letFunny: Can we keep this test? It is really testing the multiline string of three lines for the line + column reporting. You can just change it to any other error and maybe include a comment in the code saying what this is really testing.

internal/expr/expr_test.go
```diff
@@ -537,6 +555,54 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 		query:       "SELECT &NoTags.* FROM t",
 		prepareArgs: []any{NoTags{}},
 		err:         `cannot prepare statement: output expression: no "db" tags found in struct "NoTags": &NoTags.*`,
+	}, {
+		query:       "SELECT &Person FROM t",
+		prepareArgs: []any{Person{}},
+		err:         `cannot prepare statement: output expression: type "Person" missing struct db tag: &Person`,
```
letFunny: As we discussed offline, can we reword this error to:
```
unqualified type "Person" must be a primitive type
```

internal/expr/expr_test.go
```diff
@@ -537,6 +555,54 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
+	}, {
+		query:       "SELECT (*) AS (&Person) FROM t",
+		prepareArgs: []any{Person{}},
+		err:         `cannot prepare statement: output expression: type "Person" missing struct db tag: (*) AS (&Person)`,
```
letFunny: Ditto

internal/expr/expr_test.go
```diff
@@ -537,6 +555,54 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
+	}, {
+		query:       "SELECT foo FROM t WHERE x = $Person",
+		prepareArgs: []any{Person{}},
+		err:         `cannot prepare statement: input expression: type "Person" missing struct db tag: $Person`,
```
letFunny: Ditto.

letFunny (commented):
A couple of things we discussed during the Sprint.

internal/expr/expr_test.go
```diff
@@ -307,6 +311,24 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+	"custom primitive types",
+	"SELECT name AS &nameStr, address AS &addressStr, id AS &idInt FROM t WHERE name = $nameStr AND address = $addressStr AND id = $idInt",
+	"[Bypass[SELECT ] Output[[name] [nameStr]] Bypass[, ] Output[[address] [addressStr]] Bypass[, ] Output[[id] [idInt]] Bypass[ FROM t WHERE name = ] Input[nameStr] Bypass[ AND address = ] Input[addressStr] Bypass[ AND id = ] Input[idInt]]",
+	[]any{(idInt)(0), (addressStr)(""), (nameStr)("")},
+	"SELECT name AS _sqlair_0, address AS _sqlair_1, id AS _sqlair_2 FROM t WHERE name = @sqlair_0 AND address = @sqlair_1 AND id = @sqlair_2",
+}, {
+	"primitive types passed to prepare unnecessarily",
+	"SELECT id AS &int64, id AS &uint16 FROM t",
+	"[Bypass[SELECT ] Output[[id] [int64]] Bypass[, ] Output[[id] [uint16]] Bypass[ FROM t]]",
+	[]any{int64(0), uint16(0)},
```
letFunny, line 330: I am now wondering if this should be an error. It makes sense that both, passing the types to Prepare, and omitting them work, but at the same time one of the project ethos seems to be to have only one way of doing things. What do you think?

internal/expr/expr_test.go
```diff
@@ -512,23 +530,23 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 	}, {
 		query:       "SELECT * AS &Person.* FROM t",
 		prepareArgs: []any{[]any{Person{}}},
-		err:         `cannot prepare statement: need struct or map, got slice`,
+		err:         `cannot prepare statement: need struct, map, or primitive type, got slice`,
```
letFunny: I think Gustavo suggested in another PR to change the errors to be:
```
need valid type: got ...
```
and I agree with him now that we are adding slices and primitive types.

internal/expr/parser.go
```diff
@@ -510,16 +513,13 @@ func (p *Parser) parseTargetType() (valueAccessor, bool, error) {
...
+// asterisk). For example "Type" or "Type.col_name".
 func (p *Parser) parseTypeName() (valueAccessor, bool, error) {
 	cp := p.save()
-
-	// The error points to the skipped & or $.
-	identifierCol := p.colNum() - 1
 	if id, ok := p.parseIdentifier(); ok {
 		if !p.skipByte('.') {
-			return valueAccessor{}, false, errorAt(fmt.Errorf("unqualified type, expected %s.* or %s.<db tag>", id, id), p.lineNum, identifierCol, p.input)
+			return valueAccessor{typeName: id}, true, nil
```
letFunny, line 522: I think it would be great here to discern between primitive types and regular types. We have that information in the parsing stage and we can throw and error early on instead of having to wait to check the type information. One idea would be to just check the type name here against the list of primitive types.

internal/expr/prepare.go
```diff
@@ -123,10 +123,13 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []columnAccessor,
...
 
 		for _, t := range p.targetTypes {
-			info, ok := ti[t.typeName]
+			info, ok := lookupType(ti, t.typeName)
 			if !ok {
 				return nil, nil, typeMissingError(t.typeName, getKeys(ti))
 			}
+			if _, ok := info.(*primitiveTypeInfo); ok {
+				return nil, nil, fmt.Errorf(`column not specified for primitive type`)
+			}
```
letFunny: Similar to my comment above we can move this error to the parsing stage because it is not valid SQLair syntax. In this case, we can better separate the concerns of parsing vs type checking to have a clear separation between stages (also goes in line with the pipeline redesign).

internal/expr/prepare.go
```diff
@@ -154,10 +157,13 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) (outCols []columnAccessor,
...
 	// Case 2: Explicit columns, single asterisk type e.g. "(col1, t.col2) AS &P.*".
 	if starTypes == 1 && numTypes == 1 {
-		info, ok := ti[p.targetTypes[0].typeName]
+		info, ok := lookupType(ti, p.targetTypes[0].typeName)
 		if !ok {
 			return nil, nil, typeMissingError(p.targetTypes[0].typeName, getKeys(ti))
 		}
+		if _, ok := info.(*primitiveTypeInfo); ok {
+			return nil, nil, fmt.Errorf(`cannot use star with primitive type %q in expression`, info.typ().Name())
+		}
```
letFunny: See my comment about making this a parsing error.

internal/expr/query.go
```diff
@@ -46,19 +46,19 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-		}
 		if _, ok := typeValue[t]; ok {
 			return nil, fmt.Errorf("type %q provided more than once", t.Name())
 		}
 		typeValue[t] = v
 		typeNames = append(typeNames, t.Name())
 		if !inQuery[t] {
+			if t.Name() == "" {
+				return nil, fmt.Errorf("invalid input type: %q", t.String())
+			}
```
letFunny: Why is this check instead the `if` statement? It looks like we should catch this error earlier on and return directly (i.e. `if` just after defining `t`).

internal/expr/query.go
```diff
@@ -79,13 +79,17 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 		}
 		var val reflect.Value
 		switch tm := typeMember.(type) {
-		case *structField:
```
letFunny, line 82: We have to mark this PR as dependent on the duplicate types bug #98  and wait until we reach a decision there. Either that or decouple them.

internal/expr/query.go
```diff
@@ -118,25 +122,25 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 	outputVals := []reflect.Value{}
 	for _, outputArg := range outputArgs {
 		if outputArg == nil {
-			return nil, nil, fmt.Errorf("need map or pointer to struct, got nil")
+			return nil, nil, fmt.Errorf("need map or pointer to struct/primitive type, got nil")
```
letFunny: It is probably also worth it to change the error here so that we do not need to enumerate all the valid types.

internal/expr/query.go
```diff
@@ -118,25 +122,25 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
...
 		}
 		outputVal := reflect.ValueOf(outputArg)
 		k := outputVal.Kind()
 		if k != reflect.Map {
 			if k != reflect.Pointer {
-				return nil, nil, fmt.Errorf("need map or pointer to struct, got %s", k)
+				return nil, nil, fmt.Errorf("need map or pointer to struct/primitive type, got %s", k)
```
letFunny: Ditto

internal/expr/query.go
```diff
@@ -118,25 +122,25 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
...
 			}
 			if outputVal.IsNil() {
 				return nil, nil, fmt.Errorf("got nil pointer")
 			}
 			outputVal = outputVal.Elem()
-			k = outputVal.Kind()
-			if k != reflect.Struct && k != reflect.Map {
-				return nil, nil, fmt.Errorf("need map or pointer to struct, got pointer to %s", k)
-			}
```
letFunny, line 136: Why are we removing this check? Is it redundant now?

internal/expr/query.go
```diff
@@ -166,26 +170,38 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
...
-		case *structField:
-			val := outputVal.Field(tm.index)
-			if !val.CanSet() {
-				return nil, nil, fmt.Errorf("internal error: cannot set field %s of struct %s", tm.name, tm.structType.Name())
+		case primitiveType, structField:
+			switch tm := tm.(type) {
+			case primitiveType:
+				if !outputVal.CanSet() {
+					return nil, nil, fmt.Errorf("internal error: cannot set %s", tm.primitiveType.Name())
+				}
```
letFunny: [Nit]: If this is an internal error I think we should either do not re-check it here or make the code as minimal as possible, we do not need two different messages here to distinguish between structs and primitive types, especially if it means adding another branch in the switch. The last benefit is that we can now simplify the switch to an if statements.

internal/expr/typeinfo.go
```diff
@@ -9,17 +9,31 @@ import (
...
+type primitiveType struct {
+	primitiveType reflect.Type
+}
+
+func (st primitiveType) outerType() reflect.Type {
+	return st.primitiveType
+}
+
+func (st primitiveType) memberName() string {
+	panic("internal error: memberName called on primitiveType")
```
letFunny: Here we are panicking and for other internal errors we are returning an error. The fact that this function cannot return an error/bool means that maybe it is the wrong abstraction for slices and primitive types.

Alternatively if we are changing the code to make better use of interfaces we may define a more specialized version of typeMember which satisfies typeMember and that has the `memberName` function. In the code we could check that the value satisfies this new interface.

internal/expr/typeinfo.go
```diff
@@ -70,14 +84,17 @@ type structInfo struct {
...
 }
 
 func (si *structInfo) typ() reflect.Type {
 	return si.structType
 }
 
 func (si *structInfo) typeMember(member string) (typeMember, error) {
+	if member == "" {
+		return nil, fmt.Errorf(`type %q missing struct db tag`, si.structType.Name())
+	}
```
letFunny: The fact that this error has to be reported here once we have done all the validation and type checking tells me it should really be move to the parser.

internal/expr/typeinfo.go
```diff
@@ -108,7 +125,10 @@ func (mi *mapInfo) typ() reflect.Type {
 }
 
 func (mi *mapInfo) typeMember(member string) (typeMember, error) {
-	return &mapKey{name: member, mapType: mi.mapType}, nil
+	if member == "" {
+		return nil, fmt.Errorf(`type %q missing map key`, mi.mapType.Name())
+	}
```
letFunny, line 130: Ditto.

internal/expr/typeinfo.go
```diff
@@ -117,6 +137,73 @@ func (mi *mapInfo) getAllMembers() ([]typeMember, error) {
...
+
+func (pti *primitiveTypeInfo) typ() reflect.Type {
+	return pti.primitiveType
+}
+
+func (pti *primitiveTypeInfo) typeMember(member string) (typeMember, error) {
+	if member != "" {
+		return nil, fmt.Errorf(`cannot specify member of primitive type %q`, pti.primitiveType.Name())
+	}
+	return primitiveType{primitiveType: pti.primitiveType}, nil
```
letFunny: Same concern, this error should be moved to the parser and not reported here which would indicate that the type does not conform to the interface, or that the interface is not the correct abstraction.

internal/expr/typeinfo.go
```diff
@@ -117,6 +137,73 @@ func (mi *mapInfo) getAllMembers() ([]typeMember, error) {
...
+}
+
+func (pti *primitiveTypeInfo) getAllMembers() ([]typeMember, error) {
+	return nil, fmt.Errorf(`primitive type used in invalid context`)
```
letFunny: Missing test for this error.

package_test.go
```diff
@@ -492,6 +499,10 @@ func (s *PackageSuite) TestNulls(c *C) {
 }
 
 func (s *PackageSuite) TestValidGet(c *C) {
+	stringVar := ""
+	intVar := 0
+	stringExpected := "Fred"
+	intExpected := 30
```
letFunny:
```go
func getPtr[T any](val T) *T {
	return &val
}
...
        }, {
		summary:  "primitive types",
		query:    "SELECT name AS &string, id AS &int FROM person WHERE name = $string AND id = $int",
		types:    []any{},
		inputs:   []any{"Fred", 30},
		outputs:  []any{getPtr(""), getPtr(0)},
		expected: []any{getPtr("Fred"), getPtr(30)},
	}}
```
this applies to all the tests.

package_test.go
```diff
@@ -743,28 +769,28 @@ func (s *PackageSuite) TestGetAllErrors(c *C) {
...
 	}, {
-		summary: "wrong slice type (pointer to int)",
+		summary: "wrong slice type (pointer to func)",
 		query:   "SELECT * AS &Person.* FROM person",
 		types:   []any{Person{}},
 		inputs:  []any{},
-		slices:  []any{&[]*int{}},
-		err:     `cannot populate slice: need slice of structs/maps, got slice of pointer to int`,
+		slices:  []any{&[]*func(){}},
+		err:     `cannot populate slice: need slice of structs, maps, or primitive types, got slice of pointer to func`,
```
letFunny: Consider:
```
need slice of valid type, got ...
```

package_test.go
```diff
@@ -1358,6 +1384,128 @@ AND    l.model_uuid = $JujuLeaseKey.model_uuid`,
 	}
 }
 
+func (s *PackageSuite) TestPrimativeTypes(c *C) {
```
letFunny:
```suggestion
func (s *PackageSuite) TestPrimitiveTypes(c *C) {
```

package_test.go
```diff
@@ -1383,6 +1531,7 @@ func (s *PackageSuite) TestRaceConditionFinalizer(c *C) {
 	// Assert that sql.Stmt was not closed early.
 	c.Assert(q.Run(), IsNil)
 }
+
```
letFunny: The missing newline here was intentional :) so that the comment above applied to the two tests.

package_test.go
```diff
@@ -1358,6 +1384,128 @@ AND    l.model_uuid = $JujuLeaseKey.model_uuid`,
...
+	o13 := string("")
+	stmt, err = sqlair.Prepare("SELECT name AS &string FROM person WHERE name = $string")
+	c.Assert(err, IsNil)
+	err = db.Query(nil, stmt, v13).Get(&o13)
+	c.Assert(err, IsNil)
+	c.Assert(v13, Equals, o13)
+
+	v14 := bool(true)
+	o14 := bool(false)
+	stmt, err = sqlair.Prepare("SELECT true AS &bool FROM person WHERE id > $bool")
```
letFunny: Why are we using `>` with bools?

letFunny (commented):
I really like the idea and the implementation. I left a lot of comments considering the pipeline rework and some possible changes in the code structure.

doc.go
Aflynn50: Yea, this looks like it may be here by mistake. I've removed it

doc.go
Aflynn50: Agreed

doc.go
Aflynn50: Fixed, probably a result of me messing around with vim settings

doc.go
Aflynn50: Fixed.

internal/expr/expr_test.go
Aflynn50: This is a good point. I don't think we should required that people pass them because they genuinely are special cases, unlike the normal user defined types, passing them would imply you could do weird things like name your string type `int64` and have it work, which I don't think we should allow.

My thinking for having this error is that it is something a user would only ever hit once, then they would know that they don't need to pass the primitive types.  

There is some inconstancy though, we actually let people pass types to `Prepare` that are not used in the query, this error does not fit well with that.

I think I could also change the wording of this error to make it better, how does 
`primitive types do not need to be passed to prepare` sound?

internal/expr/expr_test.go
Aflynn50: Added back

internal/expr/expr_test.go
Aflynn50: Agreed, changed.

internal/expr/expr_test.go
letFunny: I agree with you that we should NOT allow people to pass primitive types to Prepare. I would try to frame the error message to say that you cannot pass types with the name of a primitive one, not that it is not needed, more like not allowed. I also agree with you that we need to be clear that the error is passing it to .Prepare, and not including it in the query string. What do you think of?
```
cannot prepare statement: type samples cannot contain primitive types
```

internal/expr/expr_test.go
Aflynn50: The errors from prepare all end with the expression that the error occurred in, so I've changed it to:

`unqualified type "Person" must be a primitive type in: &Person`

Weather it is a good idea to be appending all errors with this in prepare is something I am undecided on. This error message is in `typeInfo.go` so having it end with the word "in" feels a bit weird. Let me know if you have any better solutions than mine.

internal/expr/expr_test.go

internal/expr/expr_test.go

internal/expr/parser.go
Aflynn50: We allow named primitive types, so it could still be an unknown name, but I agree that we should encode this information in our type system.

It would be great to have `valueAccessor` as an interface where the different members could be asterisk accessors, primitive accessors and regular qualified accessors, this would also remove all the manuel checking for asterisks in `prepare.go`. I think that could be brought as a separate PR before this goes in.

internal/expr/prepare.go
Aflynn50: We could have this error in the parse stage, but it would only work for primitive types with default names, we also types defined as primitive types. In `&Person` the `Person` could be an string type `type Person string` which would be perfectly valid.

internal/expr/prepare.go
Aflynn50: We only know that the type is primitive in prepare (unless we did a text match against the builtin names, but that would not catch aliased/user defined types).

internal/expr/query.go
Aflynn50: True, this error should be up a few line, moved.

internal/expr/query.go
Aflynn50: It is already marked as dependent in the description

internal/expr/query.go

internal/expr/query.go

internal/expr/query.go
Aflynn50: This checks that the Kind of the output matches the expected map or struct. This check should now cover all the numerous different kinds that have become valid (there is a separate kind for each size/sign of int).

This error message was actually always a bit redundant, if the user passes a type with an invalid kind it will not be in the query so will be caught there. This message gave a more accurate list of what was valid, though now we would have to change it to "need valid pointer" or something like that since we are not listing all the valid types in error messages elsewhere. The message "need valid pointer" is not really any more useful that "type does not appear in query" so I had decided to remove this check completely.

If you think it would add value for the users I'm happy to add it back in.

internal/expr/query.go
Aflynn50: Changed to an if statement.

Thinking about it, I'm actually not even sure we need these checks for `CanSet`. It returns false if you have done `reflect.ValueOf` on a none pointer value, but we only accept pointers and then call `Elem` which should be fine.

Though I'm going to leave it in because there might be cases I don't know about.

internal/expr/typeinfo.go
Aflynn50: I agree, we need to replace this abstraction. Since we are refactoring the code anyway I am going to wait until the refactor is done to design the new interfaces. Hopefully then it will be clearer what is needed.

internal/expr/typeinfo.go
Aflynn50: It would be great to but at the parse stage we don't know if an unqualified named is a named primitive or an actual unqualified map/struct.

internal/expr/typeinfo.go
Aflynn50: Same as above.

internal/expr/typeinfo.go
Aflynn50: Same comment as above.

internal/expr/typeinfo.go
Aflynn50: This error is actually unreachable in the current code. This is another place where the abstraction does not quite fit the reality. I've changed it to an internal error for now, but when the refactoring is in we should have another look at this.

package_test.go
```diff
@@ -376,28 +376,24 @@ func (s *PackageSuite) TestIterGetErrors(c *C) {
 	}
 }
 
-type ScannerInt struct {
-	SI int
-}
+type ScannerInt int
```
Aflynn50, line 383: [Note] These changes to the tests for implementing scanner interfaces are not directly related to this PR but the test design makes a lot more sense with primitive types.

package_test.go
Aflynn50: Good point, changed.

package_test.go

package_test.go
Aflynn50: I've changed it to an `= 1` check. To be honest I'm not really sure how you would use a bool in a SQL query

package_test.go
Aflynn50: Added back in.

PR #98, opened 2023-08-30
bugfix: Fix duplicate type members in query catch 

Aflynn50:
Add a test for duplicate map keys and check for duplicates using equality of the string representation of the `typeMember` rather than equality of the pointer.

In #75 a bug was unwittingly introduced. `mapKey` and `structField` are both structs that implemented the `typeMember` interface. #75 change it so their a pointer to them implemented the interface so that the types could be passed around by reference rather than by value. The only problem is, that the map `typeMemberPresent` is indexed by `typeMembers`. This was therefore now indexed by pointers to structs rather than the structs themselves.

An error is thrown if two of the same `typeMember` appear in the outputs as this could create ambiguity. The test for this error still passed because in the case of `structField`s if two fields of the same struct are present then both will have the same pointer (as they are added from `tagToField` in `structInfo` which was a `map[string]*structField`. This is not the case for maps however.


internal/expr/typeinfo.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 )
 
+// typeMember should be implemented without pointer recivers as it is used as a
```
letFunny: Typo in recivers. I am not sure if I would reword the comment to say something like "as it is used as a key in maps and, even if the underlying structures are equal, their pointers will not be".

letFunny (commented):
All good, it seems to be basically a revert of the previous PR #75. I only included a comment for a typo and a possible rewording of a comment

internal/expr/typeinfo.go
Aflynn50: Changed.

letFunny (approved):

internal/expr/typeinfo.go
```diff
@@ -9,6 +9,9 @@ import (
 	"sync"
 )
 
+// typeMember should be implemented without pointer receivers as it is used as
+// a key in maps and, even if the underlying structs are equal, the pointers
+// may not be.
```
niemeyer: I'm probably missing something here. These values were designed to be a long term cache of type data, and as such using them as pointers should be fine. In the original design, it also meant that these values were always unique, so we might even use them as keys in a map, and that's a pretty good key since this is a pointer.

Now it sounds like this has changed, maybe after introducing `mapKey`?  But this outcome sounds a bit like previous conversations we had: it's an unintended design change that is spreading abstraction breaking issues.  I would instead prefer to restore the original intention of this logic: to cache struct information.

Just as some backing evidence for this, the data that composes the `structField` type below looks like a pretty bad way to have a unique map key.

Happy to discuss this further.

niemeyer (changes requested):

letFunny (changes requested):
@Aflynn50 and me talked about Gustavo's concern and how to solve it. How it works at the moment for structs is that we get the struct information and field from the type samples. For maps that is not the case because the keys are not included in the reflection information and we have to use a string, that is why the pointers wont be equal and to check for equality we actually need to check the struct value.

We discussed two options:
1. Use pointer receivers for struct to avoid copying them and because conceptually they are unique and long lived, but use value receiver for maps because of the reasons above.
2. Generate a unique key out and use that to cache information in the map. For structs we could generate the string `Struct.Field`, and `Map.key` for maps. Both will be unique per SQLair constraints.

We think the best solution is 2. because we do not have to mix receiver types and it is much simpler conceptually when reading the code.

Aflynn50:
This bugfix has been changed to use the approach detailed above. The map `typeMemberPresent` is now indexed by a string representation of the type member. This avoids the need to replace the mapKey methods with value receivers which would introduce a bad smell (even if there was a justification) into the code.

PR #99, opened 2023-09-01
Include expression in prepare errors

Aflynn50:
In `prepare.go`, include the I/O expression in which the error was found where possible. The errors are changed to reflect that the full expression is now included in the error message. The phrase "cannot prepare expression" is changed to "cannot prepare statement" to reflect that expression is used to refer only to the I/O expression.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -543,17 +543,17 @@ func (s *ExprSuite) TestPrepareMapError(c *C) {
 		"all output into map star",
 		"SELECT &M.* FROM person WHERE name = 'Fred'",
 		[]any{sqlair.M{}},
-		"cannot prepare statement: output expression: &M.*: asterisk cannot be used with maps when no column names are specified",
+		"cannot prepare statement: output expression: &M.*: asterisk cannot be used with map when no column names are specified",
 	}, {
 		"all output into map star from table star",
 		"SELECT p.* AS &M.* FROM person WHERE name = 'Fred'",
 		[]any{sqlair.M{}},
-		"cannot prepare statement: output expression: p.* AS &M.*: asterisk cannot be used with maps when no column names are specified",
+		"cannot prepare statement: output expression: p.* AS &M.*: asterisk cannot be used with map when no column names are specified",
 	}, {
 		"all output into map star from lone star",
 		"SELECT * AS &CustomMap.* FROM person WHERE name = 'Fred'",
 		[]any{CustomMap{}},
-		"cannot prepare statement: output expression: * AS &CustomMap.*: asterisk cannot be used with maps when no column names are specified",
+		"cannot prepare statement: output expression: * AS &CustomMap.*: asterisk cannot be used with map when no column names are specified",
 	}, {
 		"invalid map",
 		"SELECT * AS &InvalidMap.* FROM person WHERE name = 'Fred'",
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -421,95 +421,95 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 	}{{
 		query:       "SELECT (p.name, t.id) AS (&Address.id) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         `cannot prepare query: output expression: (p.name, t.id) AS (&Address.id): mismatched number of columns and target types`,
+		err:         `cannot prepare statement: output expression: (p.name, t.id) AS (&Address.id): mismatched number of columns and target types`,
 	}, {
 		query:       "SELECT (p.name) AS (&Address.district, &Address.street) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare query: output expression: (p.name) AS (&Address.district, &Address.street): mismatched number of columns and target types",
+		err:         "cannot prepare statement: output expression: (p.name) AS (&Address.district, &Address.street): mismatched number of columns and target types",
 	}, {
 		query:       "SELECT (&Address.*, &Address.id) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         `cannot prepare query: member "id" of type "Address" appears more than once`,
+		err:         `cannot prepare statement: member "id" of type "Address" appears more than once`,
 	}, {
 		query:       "SELECT (p.*, t.name) AS (&Address.*) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare query: output expression: (p.*, t.name) AS (&Address.*): invalid asterisk in expression columns",
+		err:         "cannot prepare statement: output expression: (p.*, t.name) AS (&Address.*): invalid asterisk in expression columns",
 	}, {
 		query:       "SELECT (name, p.*) AS (&Person.id, &Person.*) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         "cannot prepare query: output expression: (name, p.*) AS (&Person.id, &Person.*): invalid asterisk in expression columns",
+		err:         "cannot prepare statement: output expression: (name, p.*) AS (&Person.id, &Person.*): invalid asterisk in expression columns",
 	}, {
 		query:       "SELECT (&Person.*, &Person.*) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         `cannot prepare query: member "address_id" of type "Person" appears more than once`,
+		err:         `cannot prepare statement: member "address_id" of type "Person" appears more than once`,
 	}, {
 		query:       "SELECT (p.*, t.*) AS (&Address.*) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare query: output expression: (p.*, t.*) AS (&Address.*): invalid asterisk in expression columns",
+		err:         "cannot prepare statement: output expression: (p.*, t.*) AS (&Address.*): invalid asterisk in expression columns",
 	}, {
 		query:       "SELECT (id, name) AS (&Person.id, &Address.*) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         "cannot prepare query: output expression: (id, name) AS (&Person.id, &Address.*): invalid asterisk in expression types",
+		err:         "cannot prepare statement: output expression: (id, name) AS (&Person.id, &Address.*): invalid asterisk in expression types",
 	}, {
 		query:       "SELECT (name, id) AS (&Person.*, &Address.id) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         "cannot prepare query: output expression: (name, id) AS (&Person.*, &Address.id): invalid asterisk in expression types",
+		err:         "cannot prepare statement: output expression: (name, id) AS (&Person.*, &Address.id): invalid asterisk in expression types",
 	}, {
 		query:       "SELECT (name, id) AS (&Person.*, &Address.*) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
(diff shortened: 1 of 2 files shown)


letFunny (approved):
Looking good

niemeyer (changes requested):
Per conversation, let's please keep the variable length at the end of the error messages, otherwise people would need to get through a potentially long SQL piece before they even know what they are looking for.

internal/expr/expr_test.go
```diff
@@ -421,67 +421,67 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 	}{{
 		query:       "SELECT (p.name, t.id) AS (&Address.id) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         `cannot prepare statement: output expression: (p.name, t.id) AS (&Address.id): mismatched number of columns and target types`,
+		err:         `cannot prepare statement: mismatched number of columns and target types in output expression: (p.name, t.id) AS (&Address.id)`,
```
letFunny: I think it was clearer when the output/input was mentioned just after statement, that way you can know which part of the query is wrong at a glance. It also has the added benefit that it is always at the same place, after `cannot prepare statement` which is always the same number of characters, unlike the other error.

internal/expr/expr_test.go
```diff
@@ -421,67 +421,67 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
 	}, {
 		query:       "SELECT &Address.road FROM t",
 		prepareArgs: []any{Address{}},
-		err:         `cannot prepare statement: output expression: &Address.road: type "Address" has no "road" db tag`,
+		err:         `cannot prepare statement: type "Address" has no "road" db tag in output expression: &Address.road`,
 	}, {
 		query:       "SELECT street FROM t WHERE x = $Address.street",
 		prepareArgs: []any{Person{}},
-		err:         `cannot prepare statement: input expression: $Address.street: type "Address" not passed as a parameter, have: Person`,
+		err:         `cannot prepare statement: type "Address" not passed as a parameter, have: Person. Type mentioned in input expression: $Address.street`,
```
letFunny: I am not sure `Type mentioned ...` makes the error more clear, I would leave it as:
```
cannot prepare statement: type "Address" not passed as a parameter, have "Person": $Address.street
```
I think the intention here was to be explicit about what the raw query refers to but I think it is already clear as all error messages follow the same structure of having it at the end.

internal/expr/expr_test.go
Aflynn50: Very true, changed back

internal/expr/expr_test.go
Aflynn50: I've removed the full stop, its a lot nicer with only the raw expression on the end

internal/expr/expr_test.go
```diff
@@ -421,95 +421,95 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
 	}{{
 		query:       "SELECT (p.name, t.id) AS (&Address.id) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare expression: mismatched number of columns and targets in output expression: (p.name, t.id) AS (&Address.id)",
+		err:         `cannot prepare statement: mismatched number of columns and target types in output expression: (p.name, t.id) AS (&Address.id)`,
```
TheSignPainter98: Very nitpicky, but do these quote marks need to change? I don't see any double-quotes here to avoid (unlike elsewhere)

internal/expr/expr_test.go
```diff
@@ -421,95 +421,95 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
 	}, {
 		query:       "SELECT (p.name) AS (&Address.district, &Address.street) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare expression: mismatched number of columns and targets in output expression: (p.name) AS (&Address.district, &Address.street)",
+		err:         "cannot prepare statement: mismatched number of columns and target types in output expression: (p.name) AS (&Address.district, &Address.street)",
 	}, {
 		query:       "SELECT (&Address.*, &Address.id) FROM t",
 		prepareArgs: []any{Address{}, Person{}},
-		err:         `cannot prepare expression: member "id" of type "Address" appears more than once`,
+		err:         `cannot prepare statement: member "id" of type "Address" appears more than once in output expressions`,
```
TheSignPainter98, line 432: Just to confirm, can there be multiple output expressions?

internal/expr/expr_test.go
```diff
@@ -421,95 +421,95 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
 	}, {
 		query:       "SELECT (p.*, t.name) AS (&Address.*) FROM t",
 		prepareArgs: []any{Address{}},
-		err:         "cannot prepare expression: invalid asterisk in output expression columns: (p.*, t.name) AS (&Address.*)",
+		err:         "cannot prepare statement: invalid asterisk in columns in output expression: (p.*, t.name) AS (&Address.*)",
```
TheSignPainter98: I think as the offending part is given, perhaps 'output expression' is redundant? The same is true with 'input expression' elsewhere

internal/expr/prepare.go
```diff
@@ -99,16 +107,15 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
...
 	fetchInfo := func(typeName string) (typeInfo, error) {
 		info, ok := ti[typeName]
 		if !ok {
 			ts := getKeys(ti)
 			if len(ts) == 0 {
-				return nil, fmt.Errorf(`type %q not passed as a parameter`, typeName)
+				return nil, fmt.Errorf(`type %q not passed as a parameter. Type mentioned`, typeName)
 			} else {
-				return nil, fmt.Errorf(`type %q not passed as a parameter, have: %s`, typeName, strings.Join(ts, ", "))
+				return nil, fmt.Errorf(`type %q not passed as a parameter, have: %s. Type mentioned`, typeName, strings.Join(ts, ", "))
```
TheSignPainter98: This looks incomplete, but I'm also not sure whether this will be removed due to other reviews

internal/expr/expr_test.go
```diff
@@ -543,27 +543,27 @@ func (s *ExprSuite) TestPrepareMapError(c *C) {
 		"all output into map star",
 		"SELECT &M.* FROM person WHERE name = 'Fred'",
 		[]any{sqlair.M{}},
-		"cannot prepare expression: &M.* cannot be used for maps when no column names are specified",
+		"cannot prepare statement: columns must be specified for asterisk map in output expression: &M.*",
```
TheSignPainter98: 'Asterisk map' sounds a little strange of me (it sounds like a strange data structure containing only asterisks). When I read `&M.*`, I hear 'm-dot-star.' Perhaps this error could read 'map-star' or 'starred map'?

internal/expr/prepare.go
```diff
@@ -146,10 +153,10 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
 			if t.name == "*" {
 				switch info := info.(type) {
 				case *mapInfo:
-					return nil, nil, fmt.Errorf(`&%s.* cannot be used for maps when no column names are specified`, info.typ().Name())
+					return nil, nil, fmt.Errorf(`columns must be specified for asterisk map`)
 				case *structInfo:
 					if len(info.tags) == 0 {
-						return nil, nil, fmt.Errorf("type %q in %q does not have any db tags", info.typ().Name(), p.raw)
+						return nil, nil, fmt.Errorf("no db tags found for type %q", info.typ().Name())
```
TheSignPainter98: Perhaps more emphasis should be placed on 'db' and the fact that these are struct fields could be highlighted? I'm thinking just to make sure it's Go-tags that the person reading this thinks of, not some other unrelated database tags. Would something like this work?
```
`no "db:" tags found on struct %q`
```

TheSignPainter98 (commented):
These error messages look better to me, just a few notes

internal/expr/expr_test.go
Aflynn50: Yes, and this error is one that occurs across output expressions. i.e. if someone has put `&Address.id` twice in their query.

internal/expr/prepare.go
Aflynn50: Sounds like a good idea, changed.

internal/expr/expr_test.go

internal/expr/expr_test.go
Aflynn50: Changed back haha

internal/expr/expr_test.go
Aflynn50: Changed with Albertos review

internal/expr/prepare.go
Aflynn50: Removed already.

internal/expr/expr_test.go
```diff
@@ -421,95 +421,95 @@ func (s *ExprSuite) TestPrepareErrors(c *C) {
...
 	}, {
 		query:       "SELECT street AS &Address.street FROM t",
 		prepareArgs: []any{},
-		err:         `cannot prepare expression: type "Address" not passed as a parameter`,
+		err:         `cannot prepare statement: output expression: type "Address" not passed as a parameter: street AS &Address.street`,
 	}, {
 		query:       "SELECT street AS &Address.id FROM t",
 		prepareArgs: []any{Person{}},
-		err:         `cannot prepare expression: type "Address" not passed as a parameter, have: Person`,
+		err:         `cannot prepare statement: output expression: type "Address" not passed as a parameter, have Person: street AS &Address.id`,
```
niemeyer: In this case the original one was significantly more clear. There's some mangling at the end which makes it look confusing: "have Person: street"?  Also, consider what happens if the type ends up lowercased: "have pain: Much" :)

internal/expr/prepare.go
```diff
@@ -61,14 +61,21 @@ func starCount(fns []fullName) int {
...
 	info, ok := ti[p.sourceType.prefix]
 	if !ok {
 		ts := getKeys(ti)
 		if len(ts) == 0 {
 			return nil, fmt.Errorf(`type %q not passed as a parameter`, p.sourceType.prefix)
 		} else {
-			return nil, fmt.Errorf(`type %q not passed as a parameter, have: %s`, p.sourceType.prefix, strings.Join(ts, ", "))
+			// "%s" is used instead of %q to print double quotes within the
+			// joined string correctly.
+			return nil, fmt.Errorf(`type %q not passed as a parameter, have "%s"`, p.sourceType.prefix, strings.Join(ts, `", "`))
```
niemeyer: This seems to be the offending place for the earlier comment.

internal/expr/prepare.go
```diff
@@ -108,7 +117,7 @@ func prepareOutput(ti typeNameToInfo, p *outputPart) ([]fullName, []typeMember,
 			if len(ts) == 0 {
 				return nil, fmt.Errorf(`type %q not passed as a parameter`, typeName)
 			} else {
-				return nil, fmt.Errorf(`type %q not passed as a parameter, have: %s`, typeName, strings.Join(ts, ", "))
+				return nil, fmt.Errorf(`type %q not passed as a parameter, have %s`, typeName, strings.Join(ts, ", "))
```
niemeyer: Here too.

niemeyer (changes requested):
Just one more detail:

internal/expr/expr_test.go
Aflynn50: Changed to use parentheses for "have" section of error.

internal/expr/prepare.go
Aflynn50: Changed as suggested above

internal/expr/prepare.go
Aflynn50: Changed as suggested above

niemeyer (approved):
Thanks!

PR #101, opened 2023-09-06
Replace fullName with two seperate types

Aflynn50:
`fullName` is replaced with two new types: `typeName` and `columnName`. This separation allows for more intuition when looking at type signatures throughout the codebase and enforces separation of the separate uses of `fullName`.

This change also moves the code from `parts.go` into `parser.go` since `parts.go` only contained type definitions.


letFunny (approved):
I think this is a great change for readability of the code. It makes more sense now as it is both self-documenting and enforced by the type system if you are dealing with column names or type names.

internal/expr/parts.go
```diff
@@ -1,69 +0,0 @@
-package expr
-
```
niemeyer, line 2: Usually we try not to move code and refactor it at the same time, as it makes it hard to review what actually changed here. It ends up as just 70 lines more and 70 lines less on the other side.  It's also unclear how this is improving things.

internal/expr/parser.go
```diff
@@ -368,64 +439,64 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
 	cp.restore()
-	return fullName{}, false, nil
+	return typeName{}, false, nil
 }
 
-// parseList takes a parsing function that returns a fullName and parses a
+// parseList takes a parsing function that returns a T and parses a
 // bracketed, comma seperated, list.
-func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]fullName, bool, error) {
+func parseList[T any](p *Parser, parseFn func(p *Parser) (T, bool, error)) ([]T, bool, error) {
```
niemeyer, line 491: So what's the actual goal of the change?  On a quick sight, it seems to now be using generics while before we didn't (changes min version of Go), and is splitting code that was common into independent functions so that they can handle the different types.

Curiously, this is how things were originally, until we decided to unify to make logic common. So has something changed again to make it feel like it should be split once more?

niemeyer (changes requested):

internal/expr/parser.go
letFunny: The goal of the change is to differentiate between column names and type names in the type system. Before this change we were using `fullName` for both. The problem is that it was not clear where a function or some part of the code was expecting one or the other. Now, it is clear just by looking at the type.

The downsides, as you correctly state, is that this function has to be generic because we want it to work with both types, and that there is additional code duplication (i.e. `starCountTypes` and `starCountColumns`). For the latter, even though the code *seems* duplicated because several functions are similar, in reality the types are semantically different which leads to some frustration when mixing them just because the implementation can be shared.

internal/expr/parts.go
Aflynn50: This change is primarily to aid readability rather than cut down on the lines of code. Alberto's comment explains in greater depth.

This is motivated by Albertos impressions on the code coming in as a fresh pair of eyes on it.

niemeyer (approved):
Thanks for the change and for caring about readability.

PR #103, opened 2023-09-14
Change switch/case in Prepare methods of typeInfo

Aflynn50:
The switch/case statements in `prepareInput` and `prepareOutput` contain repeated code and repeated error message that can be factored out into methods on `typeInfo`.
- `typeMember(string) (typeMember, error)` is added to get a specific member of a struct or map
- `getAllMembers() ([]typeMember, error)` is added for when asterisk expressions need all members of a type

The local function `addColumns` in `prepareOutput` is removed and calls are replaced with the explicit code since with these changes it only added cognitive overhead.

This means that the error messages change slightly but any context lost here will be added back with PR #99 where we add the relevant raw SQL to the error.

### DIFF
--- a/internal/expr/typeinfo.go
+++ b/internal/expr/typeinfo.go
@@ -58,7 +58,7 @@ type typeInfo interface {
 	typ() reflect.Type
 
 	// typeMember returns the type member associated with a given column name.
-	typeMember(string) (typeMember, error)
+	typeMember(member string) (typeMember, error)
 
 	// getAllMembers returns all members a type associated with column names.
 	getAllMembers() ([]typeMember, error)
--- a/internal/expr/typeinfo.go
+++ b/internal/expr/typeinfo.go
@@ -53,8 +53,15 @@ func (f structField) memberName() string {
 	return f.tag
 }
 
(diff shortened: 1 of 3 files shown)


letFunny (approved):
Looks good, already checked as part of #69

internal/expr/typeinfo.go
```diff
@@ -53,8 +53,15 @@ func (f structField) memberName() string {
 	return f.tag
 }
 
+// typeInfo exposes useful information about types used in SQLair queries.
 type typeInfo interface {
 	typ() reflect.Type
+
+	// typeMember returns the type member associated with a given column name.
+	typeMember(string) (typeMember, error)
```
niemeyer: Having the parameter name here makes the code readable without having to read the docs above, so it's typically a good convention even if it's not strictly required.

niemeyer (approved):
Nice refactoring, thanks. Just one nitpick:

internal/expr/typeinfo.go
Aflynn50: Good point, changed.

PR #105, opened 2023-09-20
Refactor fetchInfo

Aflynn50:
The logic for looking up a type name in `typeNameToInfo` appears in both `prepareInput` and `prepareOutput` including repeated error messages. The logic for the two cases of the error message has been moved to the function: `typeNotPassedError`. The internal function in `prepareOutput` that used to contain this code has also been removed and replaced with inline map lookups.

The error message in `query.go` that is triggered when two types have the same name is clarified. It is included in this PR because the old error message is of the same form as those that have been factored into the new function.


internal/expr/prepare.go
```diff
@@ -60,17 +60,27 @@ func starCount(fns []fullName) int {
 	return s
 }
 
-// prepareInput checks that the input expression corresponds to a known type.
-func prepareInput(ti typeNameToInfo, p *inputPart) (typeMember, error) {
-	info, ok := ti[p.sourceType.prefix]
+type typeNameToInfo map[string]typeInfo
+
+func (ti typeNameToInfo) lookupInfo(typeName string) (typeInfo, error) {
```
letFunny: [Nitpick] I think the naming of the function could be improved by keeping it simple and changing it to `lookup` because the type system has the rest of the information.

Another problem with the spelling right now is that you are really looking up a typeName, not the info, in the same way that you look up a word in a dictionary and not the meaning.

letFunny (commented):
I like the refactor, it makes the code more concise and allows us to remove one of the closures that was cluttering the `prepareOutput` function.

internal/expr/prepare.go
Aflynn50: Fair enough, I've changed it

internal/expr/expr_test.go
```diff
@@ -716,7 +716,7 @@ func (s *ExprSuite) TestQueryError(c *C) {
 		query:       "SELECT street FROM t WHERE y = $Person.name",
 		prepareArgs: []any{outerP},
 		queryArgs:   []any{shadowedP},
-		err:         "invalid input parameter: type expr_test.Person not passed as a parameter, have expr_test.Person",
+		err:         `invalid input parameter: type "expr_test.Person" not passed as a parameter (have "expr_test.Person")`,
```
letFunny: I find this error a little bit confusing because it does not tell the user the type was shadowed, something that we know. If it is doable as part of this or another PR, can we change the error message?

letFunny (commented):
I like the idea of having a function to avoiding repeat the code for errors, looking good.

internal/expr/expr_test.go
Aflynn50: So in this case the test is a little misleading. I would expect that when this error was normally thrown, shadowed type and the original would be from different packages which would make this error message a lot clearer. Unfortunately its a bit more complicated to simulate that in the test.

You are right in its current form the message is misleading. I will change it to be more explict about the issue, but I won't go so far as to add a specific case for the package *and* the type name being the same. We could change it but then this would be a error for a subcase of a subcase, the first of which we would already expect to be pretty rare (that of the type name being the same but the package name different). My instinct would be to change this if we find out that there is a way people could encounter this error that's more common that the contrived example in the tests.

niemeyer (commented):
This is a bit of a curious refactoring. Supposedly it's reducing duplicated code, and yet we end up with almost exactly the same number of lines, more functions, and more methods, so a larger (undocumented?) API surface for people to read and maintain.

If you both feel like this is an improvement, we can go ahead and merge this, but not repeating code is not necessarily a good thing, despite the common DRY wisdom.

internal/expr/prepare.go
niemeyer: > Another problem with the spelling right now is that you are really looking up a typeName, not the info, in the same way that you look up a word in a dictionary and not the meaning.

This logic doesn't necessarily lead to a more clear API. When we go for an address book, are we looking for a person or for an address?  Phonebook, person or phone?  Etc.

So the answer is that, as usual, it depends.

With that said, it's great to see that there's interest in producing good APIs with sensible naming. Thanks for that.

Aflynn50: Following our discussion this morning @letFunny has done the changes to remove the method on `typeNameToInfo` and keep simply the map lookup, its looking a lot better.

internal/expr/prepare.go
```diff
@@ -60,22 +60,27 @@ func starCount(fns []fullName) int {
 	return s
 }
 
+type typeNameToInfo map[string]typeInfo
+
+func typeNotPassedError(typeName string, allTypeNames []string) error {
```
niemeyer: Can we please rename the parameters to `missingType` and `existingTypes`, to make the function more clear?

Also, maybe the function name migth be shorter?  `typeMissingError`?

internal/expr/query.go
```diff
@@ -62,7 +62,7 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 			// Check if we have a type with the same name from a different package.
 			for _, typeMember := range pe.inputs {
 				if t.Name() == typeMember.outerType().Name() {
-					return nil, fmt.Errorf("type %s not passed as a parameter, have %s", typeMember.outerType().String(), t.String())
+					return nil, fmt.Errorf("type %q not found, have type with same name: %q", typeMember.outerType().String(), t.String())
```
niemeyer: If that's the exact same issue, the message should also be the same?

internal/expr/prepare.go
```diff
@@ -60,22 +60,27 @@ func starCount(fns []fullName) int {
...
+	if len(allTypeNames) == 0 {
+		return fmt.Errorf(`type %q not passed as a parameter`, typeName)
```
niemeyer: Since we're here, should we change this message to something slightly nicer? Maybe `parameter with type %q missing, have: ...`?

niemeyer (approved):
Looks good, just some suggestions:

internal/expr/query.go
letFunny: It was changed because of my comment above https://github.com/canonical/sqlair/pull/105#discussion_r1335522636, the problem is that when the type name is the same it is not clear in the error message why it is not working.

Aflynn50: I've changed it to the new message that we use in `typeMissingError`.

internal/expr/prepare.go
Aflynn50: Sounds good, changed to that.

internal/expr/prepare.go
Aflynn50: Changed, I think those do make it clearer.

niemeyer (approved):
Thanks for the changes and the PR itself. Nice polishing.

PR #107, opened 2023-09-21
Include raw expression in `Query` errors

Aflynn50:
Add information to errors in `query.go` about the expression in the query that the error pertains to. Extra information is also added to the error that reports duplicate type members in the query, it now includes the expressions in which the two instances occur.

In the `PreparedExpr` a list of typeMembers was used to store the query input and outputs . This is replaced by a list of `typeLocations`. The `typeLocation` struct contains the `typeMember` and the raw string. `typeMember` structs are reused across statements so it would be strange to include information specific to the statement inside them. The `typeLocation` struct can also be used in the future to retain other contextual information about where the `typeMember` occurs in the query.

### DIFF
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -17,7 +17,7 @@ import (
 // TypeBoundExpr represents a SQLair statement bound to concrete Go types. It
 // contains information used to generate the underlying SQL query and map it to
 // the SQLair query.
-type TypeBoundExpr []any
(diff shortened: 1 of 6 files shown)


Aflynn50:
Note that this PR is affected by the bug fixed in #98.

letFunny:
Now that we are changing the pipeline we should hold this PR. We would not be needing `typeLocation`, only to embed the raw string in the tree replacing PreparedExpr.

PR #108, opened 2023-09-22
Add line numbers and columns to error messages

letFunny:
Add line number and columns to errors taking into account the line breaks in the input. Also, added missing tests to get full coverage of the parser errors. Lastly, change column calculations for some errors, examples below (B: before, A: after):
* `unqualified type, expected Address.* or Address.<db tag>`
```
SELECT foo FROM t WHERE x = $Address
----------------------------^-------^
                            A       B
```
* `unexpected parentheses around types after "AS"`
```
SELECT name AS (&Person.*)
---------------^----------^
               A          B
```
* `missing parentheses around types after "AS"`
```
SELECT (name) AS &Person.*
-----------------^--------^
                 A        B
```
* `missing closing parentheses`
```
SELECT (name, id) AS (&Person.name, &Person.id
---------------------^------------------------^
                     A                        B
```


internal/expr/parser.go
```diff
@@ -15,10 +15,17 @@ type Parser struct {
 	// under pos. We maintain partStart >= prevPart.
 	partStart int
 	parts     []queryPart
+	// lineNum is the number of the current line of the input.
+	lineNum int
+	// firstByteOfLine is the position of the first byte of the current line in
+	// the input.
+	firstByteOfLine int
```
Aflynn50: Could this be called `lineStart` or something else a bit shorter?

internal/expr/parser.go
```diff
@@ -15,10 +15,17 @@ type Parser struct {
...
 }
 
 func NewParser() *Parser {
-	return &Parser{}
+	return &Parser{
+		lineNum: 1,
```
Aflynn50: Is there a reason you are not just letting `init` set this?

internal/expr/parser.go
```diff
@@ -132,6 +146,9 @@ func (p *Parser) skipComment() bool {
 						}
 					}
 					return true
+				} else if p.input[p.pos] == '\n' {
```
Aflynn50: Having this here feels a bit weird. It makes me thing we might want to redesign some of the other bits of the parser to make this more natural.

A possible alternative might be to have a method that replaces p.pos++ (`nextChar` or something) which checks for a newline and is used everywhere in the parser instead of p.pos++, this would stop people adding a new function and accidentally forgetting to update this.

If that turns out to be a can of worms even just putting this code in a function called `checkNewLine` or writing a comment would be good.

internal/expr/parser.go
```diff
@@ -239,7 +260,7 @@ func (p *Parser) skipStringLiteral() (bool, error) {
 
 		// Reached end of string and didn't find the closing quote
 		cp.restore()
-		return false, fmt.Errorf("column %d: missing closing quote in string literal", cp.pos)
+		return false, fmt.Errorf("line %d, column %d: missing closing quote in string literal", p.lineNum, p.colNum())
```
Aflynn50: Would it be possible to defer adding the `line %d, column %d` in `Parse` rather than adding this to every error message in the parser?

internal/expr/parser.go
```diff
@@ -423,7 +450,8 @@ func (p *Parser) parseList(parseFn func(p *Parser) (fullName, bool, error)) ([]f
 		return nil, false, nil
 	}
 
-	parenPos := p.pos
+	// Error points to first parentheses skipped above.
+	parenCol := p.colNum() - 1
```
Aflynn50: This could maybe be called `openingParenCol` to make it more explicit.

Also, should you also worry about the line number here? It is possible to have a multi line comment in a list, also `skipBlanks` skips new lines.

internal/expr/parser.go
```diff
@@ -282,7 +303,11 @@ func (p *Parser) skipBlanks() bool {
 			continue
 		}
 		switch p.input[p.pos] {
-		case ' ', '\t', '\r', '\n':
+		case '\n':
+			p.firstByteOfLine = p.pos + 1
+			p.lineNum++
```
Aflynn50: Integrating this check into a function to replace `p.pos++` would mean you didn't have to copy this code for the third time.

internal/expr/parser.go
```diff
@@ -550,3 +581,8 @@ func (p *Parser) parseInputExpression() (*inputPart, bool, error) {
 	cp.restore()
 	return nil, false, nil
 }
+
+// colNum calculates the current column number taking into account line breaks.
+func (p *Parser) colNum() int {
```
Aflynn50: Could this function go at the top of the file closer to the defn of the parser. The file is currently (very loosely) order in terms of low level functions to high level functions.

Aflynn50 (commented):
Very good! The library will feel a lot more professional with this in.

internal/expr/parser.go
letFunny: It is so funny when you cannot come up with a name for 15 min and then someone just thinks about it instantly :)

internal/expr/parser.go
letFunny: Changed

internal/expr/parser.go
letFunny: I have thought about this and about your above comment, the problem is that we are not always using `p.pos` as there are other instances in the code like `advance` where we check the *previous* character instead. We can probably change the parser but that will involve a more general refactor which I am not sure belongs to this PR.

I will do it locally and if it is not too lengthy we can include it here.

internal/expr/parser.go
letFunny: You are right about both, added a test and corrected the second one.

internal/expr/parser.go
Aflynn50: I was thinking that you could even use this new function instead of the p.pos++ in `advance`. There is only one other case I can think of and that is the fold in skipString (which we use because it is case insensitive), but honestly I think it would be fine to just not worry about that. You could put a comment above that function noting that it doesn't handle line breaks but I think it would also be very strange if we added code to this library that hit that case so its not necessary really.

Are there any other cases you were worried about or other concerns for this? I think if not then this PR is as good a place as any to include it, rather than doing a full refactor.

internal/expr/parser.go
letFunny: Very good idea, I refactored the code and updated the PR.

internal/expr/parser.go
letFunny: I created a helper function so as to reuse the message. Defering it to the Parse() function is more challenging because the errors would have to carry the line and column information.

internal/expr/parser.go
letFunny: Done

internal/expr/parser.go
letFunny: I refactored the code, the only place left is `skipStringLiteral` which I will finish soon, I am thinking about the edge cases.

letFunny: Done as well, I had to change one comment which was a little bit misleading but the rest was straightforward.

niemeyer (approved):
Looks like an improvement indeed,  thank you.

One small request: can we please have errorAt only reporting "line N, " if N > 1?  This will make the error slightly nicer for the common and boring short-string cases.

letFunny:
@niemeyer in your suggestion to make the errors clearer, do you mean to omit the line if the input is only one line? or, do you mean to omit the line if it is the first one regardless of the total number of lines of the input?

I think the former makes the error clearer, although in the code we would need to look for newlines in the rest of the input to see that there are none because the parsing error could be reported before encountering the first one.

niemeyer (approved):
The PR is ready, thank you.

But that bug in PR #101 clearly raises an issue that needs addressing. We have name in both, and yet in one it's the prefix and the other it's the suffix. Let's please fix this convention so that we don't do this mistake again.

PR #110, opened 2023-10-06
Replace typeName and columnName

Aflynn50:
Change `typeName` to `valueAccessor` and `columnName` to `columnAccessor`. This change reduces the ambiguity introduced in #101 where both `typeName` and `columnName` had a `name` field that represented a very different thing. The names of the structs and their fields are changed here to reflect more what the values represent in the context of the program rather than what they literally are.


niemeyer (approved):
Thanks for cleaning this up, Alastair.

PR #111, opened 2023-10-10
bugfix: race condition, statement was closed early

letFunny:
Because the Query struct did not contain references to either Statement or DB, if either of those goes out of scope the underlying sql.Stmt was closed. If the user then proceeds to use the Query struct, the result will be an error.

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -1357,3 +1357,49 @@ AND    l.model_uuid = $JujuLeaseKey.model_uuid`,
 		}
 	}
 }
+
+func (s *PackageSuite) TestRaceConditionFinalizer(c *C) {
+	var q *sqlair.Query
+	// Drop all the values except the query itself.
+	func() {
+		sqldb, err := setupDB()
+		c.Assert(err, IsNil)
+
+		db := sqlair.NewDB(sqldb)
+
+		selectStmt := sqlair.MustPrepare(`SELECT 'hello'`)
+		q = db.Query(nil, selectStmt)
+	}()
+
+	// Try to run their finalizers by calling GC several times.
+	for i := 0; i <= 10; i++ {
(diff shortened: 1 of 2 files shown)


package_test.go
```diff
@@ -1357,3 +1357,49 @@ AND    l.model_uuid = $JujuLeaseKey.model_uuid`,
...
+	}()
+
+	// Try to run their finalizers by calling GC several times.
+	for i := 0; i <= 10; i++ {
+		runtime.GC()
+		time.Sleep(0)
+	}
+
+	c.Assert(q.Run(), IsNil)
+}
```
letFunny: At this moment this test does not fail even if we remove the reference to the statement in the tx Query. This is because if we are using a transaction, we use https://pkg.go.dev/database/sql#Tx.Stmt that creates a new statement for the transaction just before executing it.

Still, I have added the test and implementation just in case something changes the future and the finalizers are called. This way, the test will start failing.

package_test.go
```diff
@@ -1357,3 +1357,49 @@ AND    l.model_uuid = $JujuLeaseKey.model_uuid`,
 		}
 	}
 }
+
```
niemeyer, line 1360: Would be nice to have a comment on these. We can pretty much just rephrase the description of the PR I guess.

sqlair.go
```diff
@@ -141,6 +141,10 @@ type Query struct {
 	ctx     context.Context
 	err     error
 	tx      *TX // tx is only set for queries in transactions.
+	// Persist statement so it is not closed until the Query is dropped.
```
niemeyer: This can be just "statement and db"  I guess. :)

niemeyer (approved):
Good catch, thanks!

PR #112, opened 2023-10-11
Move SQL generation from the Type Bind stage to Query stage.

Aflynn50:
Implement the pipeline described in `internal/expr/doc.go`. 

- Move the SQL generation to the query stage to generate it alongside the query parameters. 
- Change `TypedExpr` from containing lists of inputs outputs and the SQL to containing a typed versions of the expressions. This means it also retains the same shape as the `ParsedExpr`.
- Move the functions `bindOutputTypes` and `bindInputTypes` methods on parsed input and output expressions.
- Change `TypedExpr` to `TypeBoundExpr`. This is to remove the confusion with the name of the type constituents  `typedExpression`.


internal/expr/query.go
```diff
@@ -12,35 +13,53 @@ func (qe *QueryExpr) QueryArgs() []any {
...
+// markerIndex returns the int X from the string "_sqlair_X".
+func markerIndex(s string) (int, bool) {
+	if strings.HasPrefix(s, markerPrefix) {
+		n, err := strconv.Atoi(s[len(markerPrefix):])
+		if err == nil {
+			return n, true
+		}
+	}
+	return 0, false
+}
```
letFunny: The fact that we are creating a pattern just to later parse it again seems like an anti-pattern. Is there a way to keep the information and only create the actual marker later on?

I know this function existed beforehand but if we are redesigning the information flow it seems like an opportunity for improvement.

internal/expr/query.go
```diff
@@ -12,35 +13,53 @@ func (qe *QueryExpr) QueryArgs() []any {
...
+// QueryExpr represents a complete SQLair query, ready for execution on a
+// database.
 type QueryExpr struct {
 	sql     string
 	args    []any
 	outputs []typeMember
 }
 
-// Query returns a query expression ready for execution, using the provided values to
+const markerPrefix = "_sqlair_"
```
letFunny: Can we change the name to `outputMarkerPrefix`, `outputMarkerName`, etc.? or even shorter like `outMarkerPrefix`, `outMarkerName`, etc.

internal/expr/query.go
```diff
@@ -58,38 +77,80 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-					return nil, fmt.Errorf("parameter with type %q missing, have type with same name: %q", typeMember.outerType().String(), t.String())
+	}
+
+	// Query parameteres.
+	qargs := make([]any, 0)
+	outputs := make([]typeMember, 0)
+	inQuery := make(map[reflect.Type]bool)
+	inCount := 0
+	outCount := 0
+	var sqlStr bytes.Buffer
```
letFunny: [Nit] Avoid `var`.

internal/expr/query.go
```diff
@@ -58,38 +77,80 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 		typeValue[t] = v
 		typeNames = append(typeNames, t.Name())
-		if !inQuery[t] {
-			// Check if we have a type with the same name from a different package.
-			for _, typeMember := range pe.inputs {
-				if t.Name() == typeMember.outerType().Name() {
-					return nil, fmt.Errorf("parameter with type %q missing, have type with same name: %q", typeMember.outerType().String(), t.String())
+	}
+
+	// Query parameteres.
```
letFunny:
```suggestion
	// Query parameters.
```

internal/expr/query.go
```diff
@@ -58,38 +77,80 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+	qargs := make([]any, 0)
+	outputs := make([]typeMember, 0)
+	inQuery := make(map[reflect.Type]bool)
```
letFunny: [Nit] Can we write a very small comment or change the name to reflect that this is checking whether `args` have more types than what the query expects?

internal/expr/query.go
letFunny: I see that we are basically using it to get the `typeMember` when scanning the args. We could change query to also output something that has the proper association between types and columns. We can leave this for a future PR as it does not have a lot to do with changing Prepare.

internal/expr/query.go
```diff
@@ -58,38 +77,80 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
+		switch pp := pp.(type) {
+		case *preparedInput:
+			typeMember := pp.inputValue
+			outerType := typeMember.outerType()
+			v, ok := typeValue[outerType]
+			if !ok {
+				argTypes := make([]reflect.Type, 0)
+				for argType, _ := range typeValue {
+					argTypes = append(argTypes, argType)
 				}
```
letFunny: I think this copy is unnecessary because the function below, `checkShadowedType` is not really mutating the slice.

internal/expr/query.go
```diff
@@ -58,38 +77,80 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-				return nil, fmt.Errorf(`map %q does not contain key %q`, outerType.Name(), tm.name)
-			}
+	}
+
+	return &QueryExpr{outputs: outputs, sql: sqlStr.String(), args: qargs}, nil
+}
+
+// checkShadowedType returns an error if a query type and some argument type
+// have the same name but are from a different package.
+func checkShadowedType(queryType reflect.Type, argTypes []reflect.Type) error {
```
letFunny: I like that you have created a function for this. I was little skeptical reading the code because it is only used in one place but I think it aids readability. This way, we have less indentation levels in the above function and, from this function name, it is clear what it does.

internal/expr/prepare.go
```diff
@@ -1,43 +1,46 @@
...
+// information to generate the relevent SQL and access any Go types pertaining
+// to the part.
+type preparedPart interface {
+	preparedPart()
 }
 
-const markerPrefix = "_sqlair_"
+type preparedOutput struct {
+	queryColumns []columnAccessor
+	outputValues []typeMember
```
letFunny: [Nit] What do you think of naming it `outputTypes` or `columnTypes`.

internal/expr/prepare.go
```diff
@@ -1,43 +1,46 @@
...
-	return pe.sql
+// preparedPart represents a part of a SQLair statement. It contains
+// information to generate the relevent SQL and access any Go types pertaining
+// to the part.
+type preparedPart interface {
+	preparedPart()
 }
 
-const markerPrefix = "_sqlair_"
+type preparedOutput struct {
```
letFunny: I know in the code we are only traversing each list independently **right now**, but the `queryColumns` and `outputValues` are really in a 1:1 correspondence, can we create a list that contains the column and type together?

This way, we would not be losing the association between column and type which might prove useful later on.

internal/expr/prepare.go
```diff
@@ -1,43 +1,46 @@
...
-	inputs  []typeMember
-	sql     string
+	preparedParts []preparedPart
 }
 
-// SQL returns the SQL ready for execution.
-func (pe *PreparedExpr) SQL() string {
-	return pe.sql
+// preparedPart represents a part of a SQLair statement. It contains
+// information to generate the relevent SQL and access any Go types pertaining
```
letFunny:
```suggestion
// information to generate the relevant SQL and access any Go types pertaining
```

internal/expr/expr_test.go
```diff
@@ -44,43 +45,45 @@ type IntMap map[string]int
...
-	summary          string
-	query            string
-	expectedParsed   string
-	prepareArgs      []any
-	expectedPrepared string
+	summary        string
+	query          string
+	expectedParsed string
+	prepareArgs    []any
+	queryArgs      []any
```
letFunny: [Nit] Can we explicitly include the empty variable in the test cases, it makes them easier to read if they all have the same fields.

letFunny (commented):
As we discussed, I think the idea behind this refactor is very natural and I really like how it is shaping up. On that note, I think we should reword the PR description to give a high level overview of what we are doing:
* Moving the SQL generation from Prepare to Query.
* Changing the output of Prepare so that it becomes an intermediate representation of the AST. Because we are keeping the structure of the AST we no longer need to keep seemingly unrelated arrays in sync by keeping track of a myriad of indexes.

Also, I would state that the motivation behind this PR is that for several other changes, like IN clauses or omitempty, we had to create ad-hoc solutions to pass information around that was already in the AST from the beginning. And obviously we are moving the SQL generation out of Prepare because we do not know the slice length in Prepare.

Lastly, I would only mention this changes. Although I think the low level details that you included in the description are useful for the reviewer, at the end of the day they are in the diff itself, and I think they are not important while browsing Git history.

internal/expr/query.go
Aflynn50: When we retrieve the column names from the results use `markerIndex` to parse them. The int returned is used to find the struct field/map key that we want to scan the result in that column into. It would be good to remember how the columns relate to the `typeMember` values simply by the position/index of the column but then if a non-SQLair column was returned in the query it would mess up our system.

We could store the `typeMember` values currently stored in `qe.outputs` in a map indexed by the column names or something along those lines, which may make the code a little simpler, but I think you're right that it would be best in a separate PR.

internal/expr/query.go
Aflynn50: I would rather not change this function here because we are also moving it and it makes the changes hard to see. It is also not directly related to what this PR does. Relating to the comment above, if you wanted to bring a PR for that and change this there as well that would be great

internal/expr/query.go

internal/expr/query.go

internal/expr/query.go
Aflynn50: Changed name to `typeUsed`.

internal/expr/query.go
Aflynn50: Is there a way to not copy it? `typeValue` is a map not a slice. I have changed this to make it a tiny bit more efficient by putting a capacity on the new slice.

internal/expr/prepare.go
Aflynn50: I originally went with `outputValues` because we have got `valueAccessor`. But you're right, they're not really values. I think the problem with calling them types is that they are not really types, they are struct fields, or map keys which is slightly different.

I'm very much struggling to come up with a good name here. I've changed them to just `columns` and `outputs` for now.

Aflynn50: I think it would be good to discuses this further, I think we still haven't found a great consistent name for this concept across the codebase.

internal/expr/prepare.go

internal/expr/expr_test.go
Aflynn50: I based this pattern off PL007 - (the Test Discipline section). In my opinion it makes it clearer, each test entry in the list of tests can be examined individually to see what is expected, if something is not expected or not inputted then it is left out to reduce visual clutter.

internal/expr/prepare.go
Aflynn50: This is a very good point. I've had a go at unifying them and it works nicely. Though I think it could be grouped very well in a PR with the point you had about the remembering the column names above. I would love to include that in this PR but it is already massive and I don't really want to make it any bigger. We should very much bring a PR for this afterwards though.

internal/expr/query.go
Aflynn50: Reading your later comments, I think it would be great to store the `typeMember` values and column names together which would eliminate the need for this function. But I think it would be best brought as a subsequent PR since this one is already rather large.

internal/expr/prepare.go
```diff
@@ -1,43 +1,49 @@
...
+// referenced in the part.
+type preparedPart interface {
+	preparedPart()
 }
 
-const markerPrefix = "_sqlair_"
+// preparedOutput contains the columns to fetch from the database and
+// information about the Go values to read the results into.
+type preparedOutput struct {
+	columns []columnAccessor
```
Aflynn50, line 25: I would like to change this type to `string` instead of `columnAccessor` since at this point we have "prepared" the column and only need it for the SQL generation. I think this should be in a separate PR though.

niemeyer (changes requested):
During this week's call we've jointly reviewed a PR with a proposal that was clearly lacking in terms of general abstraction organization, with several layers being blended and somewhat disregarded. So we agreed to go slower and to do more clear steps that can be evaluated and merged independently.

In the latter sense, this PR is a good move, breaking down one aspect and attempting to addresss it. But in the former sense, it's hard to tell if there's now some clear sense of direction which will result in a better code base or if we're still just moving things around without clear goals. It simply says "move SQL generation" from one stage to the next. But, what do those stages mean?  Why do we have a prepared expression which is not actually prepared now?  What is the responsibility of each layer, before and after this move?   What invariants do we want to keep, and when does it make sense to drop or repurpose them?

Unless we have a clear view of the design, the outcome will not be good.

As such, can we please produce a written down version of how the pipeline is organized today, and clearly outline which responsibilities each stage has, and which responsibilities each stage does NOT have, so that these lines can be respected now and in the future?  If we struggle to produce such an outline, we should fix the design until that's easy.

internal/expr/expr_test.go
letFunny: Then leave it as it is, thanks for commenting about the reasoning.

internal/expr/prepare.go
letFunny: Okay, let's do that. Can you write a comment in this PR saying what we are leaving out for next one. That will aid other reviewers in not focusing too much on it.

internal/expr/query.go
letFunny: True, I did not realize it was a map. Then my suggestion would be to use `copy` directly.

internal/expr/prepare.go
Aflynn50: I've decided we might just as well change it here, I had thought it would affect some other things but I was wrong, the change is now in.

internal/expr/query.go
Aflynn50: There doesn't seem to be an obvious way to use `copy` here, we are copying the keys of a map into a slice and `copy` copies a slice into another slice.

internal/expr/expr_test.go
```diff
@@ -44,43 +45,45 @@ type IntMap map[string]int
 type StringMap map[string]string
 
 var tests = []struct {
```
letFunny: As we discussed, we cannot really see the tests that were added or deleted here because we are moving code around and also modifying it. This is why #115 needs to be merged first.

internal/expr/prepare.go
```diff
@@ -237,56 +252,37 @@ func (pe *ParsedExpr) Prepare(args ...any) (expr *PreparedExpr, err error) {
...
-	var sql bytes.Buffer
-
-	var inCount int
-	var outCount int
-
-	var outputs = make([]typeMember, 0)
-	var inputs = make([]typeMember, 0)
-
-	var typeMemberPresent = make(map[typeMember]bool)
+	typeMemberPresent := make(map[typeMember]bool)
```
letFunny: Can we have a comment here saying what `typeMemberPresent` is checking? If I remember correctly it was checking that there were no extra types.

letFunny (changes requested):
As we agreed we are going to hold this PR until we get the spec approved. Publishing a few minor comments from yesterday's discussion.

internal/expr/prepare.go
Aflynn50: Changed the name to `outputMemberUsed` to make it clearer

internal/expr/bindinputs.go
```diff
@@ -10,35 +11,20 @@ import (
...
 	}()
 
-	var inQuery = make(map[reflect.Type]bool)
-	for _, typeMember := range te.inputs {
-		inQuery[typeMember.OuterType()] = true
-	}
-
-	var typeValue = make(map[reflect.Type]reflect.Value)
-	var typeNames []string
+	var typeToValue = make(map[reflect.Type]reflect.Value)
```
letFunny: [Nit]: Can we follow the project's convention and use `:=` now that we are already changing it?

internal/expr/bindinputs.go
```diff
@@ -10,35 +11,20 @@ import (
...
-	outputs []typeinfo.Member
-	inputs  []typeinfo.Member
-	sql     string
-}
-
-// SQL returns the SQL ready for execution.
-func (te *TypedExpr) SQL() string {
-	return te.sql
-}
+type TypedExprs []typedExpression
```
letFunny: I am not sure `TypedExprs` is the best name. To me it could imply that we have a list of expressions that might not be connected.

internal/expr/bindinputs.go
```diff
@@ -49,41 +35,111 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+	outCount := 0
+	sqlStr := bytes.Buffer{}
+	for _, te := range *tes {
+		switch te := te.(type) {
+		case *typedInputExpr:
+			typeMember := te.input
+			outerType := typeMember.OuterType()
+			v, ok := typeToValue[outerType]
+			if !ok {
+				return nil, inputMissingError(outerType, typeToValue)
```
letFunny: What about renaming it to `typeNotInQueryError` or even `typeNotReferencedInQueryError`? I think it is clearer.

internal/expr/bindinputs.go
```diff
@@ -49,41 +35,111 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
-			for _, typeMember := range te.inputs {
-				if t.Name() == typeMember.OuterType().Name() {
-					return nil, fmt.Errorf("parameter with type %q missing, have type with same name: %q", typeMember.OuterType().String(), t.String())
+		typeToValue[t] = v
+	}
+
+	// Generate SQL and query parameters.
+	params := []any{}
+	outputs := []typeinfo.Member{}
+	argTypeUsed := map[reflect.Type]bool{}
```
letFunny, line 49: I think I said this in another review but can we really add a comment here? Maybe it's only me but I find difficult to know what it does without reading the implementation or being very familiar with the code.

Something like:
```
// This map is used to check that all the query parameters are referenced in the query.
```
will make it easier to read in my opinion.

internal/expr/bindinputs.go
```diff
@@ -49,41 +35,111 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
-		if err != nil {
-			return nil, err
-		}
+	return &PrimedQuery{outputs: outputs, sql: sqlStr.String(), params: params}, nil
+}
 
-		params = append(params, sql.Named("sqlair_"+strconv.Itoa(i), val.Interface()))
+// typedExpression represents a expression bound to a type. It contains
+// information to generate the SQL for the part and to access Go types
+// referenced in the part.
```
letFunny: I think we should remove the word `part` which does no longer mean anything in the code. Also, I think the sentence `access Go types referenced in the part` could be improved. In fact, I think `information to generate the SQL` is enough for the interface, the bit about accessing values is not strictly relevant for bypass expressions for example.

internal/expr/bindinputs.go
```diff
@@ -49,41 +35,111 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+type typedExpression interface {
+	// typedExpr is a marker method.
+	typedExpr()
+}
+
+// outputColumn stores the name of a column to fetch from the database and the
+// type member to scan the result into.
```
letFunny: `type member` does not have any meaning in itself anymore, right? Can we say `typeinfo.Member` instead?

internal/expr/bindinputs.go
letFunny: Another [Nit], I find "typedExpression represents a expression bound to a type" difficult to parse. Maybe we can say:
```go
// typedExpression represents a expression whose type names have been bound to Go types. It ...
```

internal/expr/bindinputs.go
```diff
@@ -49,41 +35,111 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+type typedOutputExpr struct {
+	outputColumns []outputColumn
+}
+
+// typedExpr is a marker method.
+func (*typedOutputExpr) typedExpr() {}
+
+// checkUsed checks if the members in the typed output expression have already
+// been used in the query. It updates the memberUsed map with its own members.
+func (toe *typedOutputExpr) checkUsed(memberUsed map[typeinfo.Member]bool) error {
```
letFunny: Something about this function seems wrong to me. It is called `checkSomething` but it is also mutating a map passed as an argument. Could we change the map in the call site if there is no error or, alternatively, and less ideal in my opinion, could we change the name so that it is clear this function updates the map?

internal/expr/bindinputs.go
```diff
@@ -100,3 +156,17 @@ func markerIndex(s string) (int, bool) {
 	}
 	return 0, false
 }
+
+// inputMissingError returns an error message for a missing input type.
+func inputMissingError(missingType reflect.Type, typeToValue map[reflect.Type]reflect.Value) error {
+	// check if the missing type and some argument type have the same name but
```
letFunny:
```suggestion
	// Check if the missing type and some argument type have the same name but
```

internal/expr/bindtypes.go
```diff
@@ -5,24 +5,21 @@ import (
...
 	"github.com/canonical/sqlair/internal/typeinfo"
 )
 
-// ParsedExpr is the AST representation of SQLair query. It contains only
+// ParsedExprs is the AST representation of SQLair query. It contains only
 // information encoded in the SQLair query string.
-type ParsedExpr struct {
-	exprs []expression
-}
+type ParsedExprs []expression
```
letFunny: Same comment about `ParsedExpressions` seemingly implying it is a list of unrelated expressions.

internal/expr/bindtypes.go
```diff
@@ -75,63 +72,68 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
+			return nil, err
+		}
 
-	// Check and expand each query expr.
-	for _, expr := range pe.exprs {
-		switch e := expr.(type) {
-		case *inputExpr:
-			inLoc, err := bindInputTypes(ti, e)
+		if toe, ok := te.(*typedOutputExpr); ok {
+			err = toe.checkUsed(outputMemberUsed)
```
letFunny: That is what I mean above. When I see this I check for `outputMemberUsed` to see where the data is coming from because I will never expect it to be mutated inside.

letFunny: Also, it feels a little bit weird for this function to be a method and be located in a different file. I think it should be a private function in this file instead.

internal/expr/bindtypes.go
```diff
@@ -75,63 +72,68 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
-		default:
-			return nil, fmt.Errorf("internal error: unknown query expr type %T", expr)
 		}
+
+		typedExprs = append(typedExprs, te)
 	}
 
-	return &TypedExpr{inputs: inputs, outputs: outputs, sql: sql.String()}, nil
+	typedExpr := TypedExprs(typedExprs)
+	return &typedExpr, nil
```
letFunny: Combine this into one line.

internal/expr/bindtypes.go
```diff
@@ -75,63 +72,68 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
+// expression represents a parsed node of the SQLair query's AST.
+type expression interface {
+	// String returns a text representation for debugging and testing purposes.
+	String() string
+
+	// bindTypes generates a typed expression from the query type information.
+	bindTypes(typeNameToInfo) (typedExpression, error)
+}
+
+// bypass represents part of the expression that we want to pass to the backend
```
letFunny: I will remove the word `part`. Maybe something like:
```go
// bypass represents an expression that we want to pass to the backend database verbatim

// bypass is a node of the AST that we want to pass to the backend database verbatim
```
I prefer the second one.

internal/expr/bindtypes.go
```diff
@@ -75,63 +72,68 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
+}
+
+// bindTypes is part of the expression interface. bypass expressions have no
+// types so the same expression is returned.
+func (e *bypass) bindTypes(typeNameToInfo) (typedExpression, error) {
+	return e, nil
+}
+
+// inputExpr represents a named parameter that will be sent to the database
+// while performing the query.
```
letFunny: It does not represent a named parameter, that is what the struct is going to be used for. It represents a node in the AST into which the user can input a value.

internal/expr/bindtypes.go
```diff
@@ -75,63 +72,68 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
+	sourceType valueAccessor
+	raw        string
+}
+
+func (e *inputExpr) String() string {
+	return fmt.Sprintf("Input[%+v]", e.sourceType)
+}
+
+// bindTypes binds the input expression to a query type and
+// returns a typed input expression.
```
letFunny: [Nit]: inconsistent line lengths.

internal/expr/bindtypes.go
```diff
@@ -143,17 +145,28 @@ func bindInputTypes(ti typeNameToInfo, e *inputExpr) (tm typeinfo.Member, err er
...
-	return tm, nil
+	return &typedInputExpr{tm}, nil
 }
 
-// bindOutputTypes binds the output expression to concrete types. It then checks
-// the expression is formatted correctly and generates the columns for the query
-// and the typeMembers the columns correspond to.
-func bindOutputTypes(ti typeNameToInfo, e *outputExpr) (outCols []columnAccessor, typeMembers []typeinfo.Member, err error) {
+// outputExpr represents a named target output variable in the SQL expression,
+// as well as the source table and column where it will be read from.
```
letFunny:
```gp
// outputExpr represents a named target output variable in the SQL expression.
// It contains the source table and column where it will be read from and the value
// where it will be read to.
```
It does not **represent** the source table and column ...

internal/expr/bindtypes.go
```diff
@@ -143,17 +145,28 @@ func bindInputTypes(ti typeNameToInfo, e *inputExpr) (tm typeinfo.Member, err er
...
+	raw           string
+}
+
+func (e *outputExpr) String() string {
+	return fmt.Sprintf("Output[%+v %+v]", e.sourceColumns, e.targetTypes)
+}
+
+// bindTypes binds the output expression to concrete types. It then checks the
+// expression is formatted correctly and generates a typed output expression.
+func (e *outputExpr) bindTypes(ti typeNameToInfo) (te typedExpression, err error) {
```
letFunny: Can we change the return of the function to `*typedOutputExpr` which makes it clearer that it will only return that implementor of the interface?

internal/expr/bindtypes.go
```diff
@@ -143,17 +145,28 @@ func bindInputTypes(ti typeNameToInfo, e *inputExpr) (tm typeinfo.Member, err er
...
+	targetTypes   []valueAccessor
+	raw           string
+}
+
+func (e *outputExpr) String() string {
+	return fmt.Sprintf("Output[%+v %+v]", e.sourceColumns, e.targetTypes)
+}
+
+// bindTypes binds the output expression to concrete types. It then checks the
+// expression is formatted correctly and generates a typed output expression.
```
letFunny: What is "formatted" correctly? We could say something like,
```go
// bindTypes binds the output expression type names to concrete Go types. It then checks that
// the result is still valid and generates a typed output expression.
```

internal/expr/bindtypes.go
```diff
@@ -176,50 +191,50 @@ func bindOutputTypes(ti typeNameToInfo, e *outputExpr) (outCols []columnAccessor
...
 		}
 		for _, c := range e.sourceColumns {
 			tm, err := info.TypeMember(c.columnName)
 			if err != nil {
-				return nil, nil, err
+				return nil, err
 			}
-			typeMembers = append(typeMembers, tm)
-			outCols = append(outCols, c)
+			oc := outputColumn{sql: c.String(), tm: tm}
```
letFunny: [Nit]: Now that I see that the usage of `c.String()` is to output SQL maybe it will be good to rename it to `SQL()`.

letFunny (commented):
I reviewed the main two files. Tomorrow, I will review the tests and the parser and leave more comments.

In general I think it looks pretty good and there are definitely quality of life changes here. I was a little bit pedantic about the comments and terminology because we are changing a lot of things very fast and sometimes we use the old names which make it confusing.

I also have a couple general comments:
* Can we add type assertions to make sure the types implement the interfaces (e.g. `var _ Interface = (*MyType)nil`.
* It is difficult to know which tests were added/removed by looking at the diff. In general we should not refactor code and add things at the same time.

internal/expr/bindinputs.go
Aflynn50: I don't think its a great name either but I'm struggling to come up with a better one. The reason for this change is that having a `TypedExpr` made up of `typedExpressions` is very confusing, especially when it comes to variable names in the code.

We could maybe rename one of the to just "typed"? We could also change `TypedExprs` to `TypeBoundExpr`? What do you think of either of those?

internal/expr/bindinputs.go
Aflynn50: The error is that one of the types in the type bound expression cannot be found in the input arguments, rather than the other way around. I've changed the names and function description to reflect this better.

internal/expr/bindinputs.go
Aflynn50: I've changed it for typeBoundExpr for now, but let me know what you think.

internal/expr/bindinputs.go
Aflynn50: Added comment

internal/expr/bindinputs.go
Aflynn50: I've changed the first sentence as suggested. For the second sentence I've actually removed it and put a modified version on `TypeBoundExpr` itself because I think it belongs more there. The sentence that was there before was a bit vague. We now have:
```
// TypeBoundExpr represents a SQLair query bound to concrete Go types. It
// contains information used to generate the underlying SQL query and map it to
// the SQLair query.

// typedExpression represents a expression with the type names bound to Go
// types.
```

internal/expr/bindinputs.go
Aflynn50: I've changed it to just "type" rather than "type member". A member is still a type and in the future it may not be a member at all.

internal/expr/bindinputs.go
Aflynn50: That's fair, I'm also not sure about it being a method on typedOutputExpr. I added this function because I didn't like that we were digging into the typedOutputExpr in BindTypes, if we update the map there we still have to do that really. 

I think the choice is between in-lining this function and having it update the map. I've gone with updating the map for now because I think it makes the code easier to in bindInputs, but I'm also fine with just in-lining it.

internal/expr/bindtypes.go
Aflynn50: I've changed this back for now.

internal/expr/bindinputs.go
letFunny: I prefer `typeBoundExpr` over `TypedExprs` but I also like `TypedTree` or something that indicates that it is a tree/query/"AST"/IR that happens to be representable by a slice.

internal/expr/bindinputs.go
letFunny: What about `typeNotBoundError`?

internal/expr/bindinputs.go
letFunny: I love the new comments, they are crystal clear to me

letFunny: My only comment would be against using the words "SQLair query" as it seems to be a heated topic.

internal/expr/bindtypes.go
Aflynn50: Agreed, I've moved it to this file and make it not a method.

internal/expr/bindtypes.go
Aflynn50: Unfortunately this can't be, Go won't let you directly take the address when you do a type assertion

internal/expr/bindinputs.go
letFunny: Is it really a type then? I now it is a little bit pedantic but you cannot "scan something into a type". Maybe:
```go
// outputColumn stores the name of a column to fetch from the database and the
// information on how to scan the result into a Go value.
```

internal/expr/bindtypes.go
Aflynn50: Yea, this sentence could do with rewording. I've changed it to
```
// bypass is an expression representing a chunk of SQL that we want to pass to
// the backend database verbatim.
```
I want to talk about it in terms of the types we have rather than as a node in the AST, even though it is that. We have the types above the we can uses as names for it to make it clear both how we use it and what it is at the same time.

internal/expr/bindtypes.go
Aflynn50: Comment changed.

internal/expr/bindtypes.go
Aflynn50: Go doesn't allow you to do this and still implement the interface.

internal/expr/bindtypes.go
Aflynn50: It is in this case, but this is a parser type, it represents a column that was in the SQLair query so it can contain an asterisk which we would not want the SQL method to return.

internal/expr/expr_test.go
```diff
@@ -89,245 +92,286 @@ var tests = []struct {
...
+	summary:        "insert array",
+	query:          "INSERT INTO arr VALUES (ARRAY[[1,2],[$HardMaths.x,4]], ARRAY[[5,6],[$HardMaths.y,8]]);",
+	expectedParsed: "[Bypass[INSERT INTO arr VALUES (ARRAY[[1,2],[] Input[HardMaths.x] Bypass[,4]], ARRAY[[5,6],[] Input[HardMaths.y] Bypass[,8]]);]]",
+	typeSamples:    []any{HardMaths{}},
+	inputArgs:      []any{HardMaths{X: 1, Y: 2}},
+	expectedParams: []any{1, 2},
+	expectedSQL:    "INSERT INTO arr VALUES (ARRAY[[1,2],[@sqlair_0,4]], ARRAY[[5,6],[@sqlair_1,8]]);",
 }}
 
 func (s *ExprSuite) TestExpr(c *C) {
```
letFunny: What does `Expr` mean in this test? That we are testing the whole `expr` package?

internal/expr/expr_test.go
```diff
@@ -604,124 +648,61 @@ func (s *ExprSuite) TestMapError(c *C) {
 	}
 }
 
-func (s *ExprSuite) TestValidBindInputs(c *C) {
```
letFunny, line 607: I suppose all of these tests are redundant now and do not have to be migrated, right?

internal/expr/bindtypes.go
```diff
@@ -176,50 +208,50 @@ func bindOutputTypes(ti typeNameToInfo, e *outputExpr) (outCols []columnAccessor
...
 				// Generate asterisk columns.
 				allMembers, err := info.GetAllMembers()
 				if err != nil {
-					return nil, nil, err
+					return nil, err
 				}
-				typeMembers = append(typeMembers, allMembers...)
 				for _, tm := range allMembers {
-					outCols = append(outCols, columnAccessor{pref, tm.MemberName()})
+					oc := outputColumn{sql: colString(pref, tm.MemberName()), tm: tm}
```
letFunny: Feels a little bit weird to use a private function on another file, and that the function `colString` looks internal. What about creating a constructor for outputColumn?

letFunny (commented):
Finished reviewing the renaming files, looking good.

internal/expr/bindtypes.go
letFunny: Is it idiomatic in Go to use `String()` to provide a stable representation of a value vs providing a user facing string representation. If it is the former then I agree, if it is the latter then we should rename it to `Raw` or something similar.

internal/expr/bindtypes.go
letFunny: Sad, thanks

internal/expr/expr_test.go
Aflynn50: Exactly that, though the name was from when all three major types were Expr types. I'll change it to `TestExprPkg`

internal/expr/expr_test.go
Aflynn50: They have been integrated into the `TestExprPkg` where they are not already redundant.

internal/expr/bindtypes.go
Aflynn50: Fair, I wasn't sure about that colString method either. I've created a constructor

internal/expr/bindinputs.go
letFunny: Re-reading this comment I think it does not add to much value, it is merely saying what the name implies. What do you think of stating its purpose instead:
```go
// argTypeUsed is used to check that all the query parameters are referenced in the query.
```

internal/expr/bindtypes.go
```diff
@@ -44,80 +41,113 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
-		default:
-			return nil, fmt.Errorf("internal error: unknown query expr type %T", expr)
+	typedExpr := TypeBoundExpr(typedExprs)
+	return &typedExpr, nil
+}
+
+// trackUsedOutputs tracks the outputs used already and returns an error if an
+// output has been used twice. It updates outputUsed with the outputs from the
+// typedOutputExpr.
+func trackUsedOutputs(outputUsed map[typeinfo.Output]bool, toe *typedOutputExpr) error {
```
letFunny: I am not sure this should be a separate function, to me it is clearer reading the 6 lines than having to click on the IDE to see the function comment and read it.

internal/expr/bindinputs.go
Aflynn50: I'll leave it as `TypeBoundExpr` then to keep the `Expr` part consistent.

internal/expr/bindinputs.go
Aflynn50: This error is now in the `typeinfo` package. Its called `valueNotFoundError`.

internal/expr/bindinputs.go
Aflynn50: Comment added. Now with the changes from #124 the definition and usage of it fit on the screen at the same time, hopefully that also makes it a lot clearer.

internal/expr/bindinputs.go
Aflynn50: Changed to statement.

internal/expr/bindinputs.go
Aflynn50: I had changed it to "type location" before, but I've just changed it again to be more precise without being verbose.

internal/expr/bindtypes.go
Aflynn50: This call has disappeared now with the `newOutputColumn` function.

internal/expr/bindtypes.go
Aflynn50: I've inlined it again. I'd originally made it a method of the `typedOutputExpr` because I didn't want to break abstraction levels in the `BindInputs` function (since this relies on the internals of the `typedOutputExpr`), but I agree that as a standalone function it doesn't do much.

PR #113, opened 2023-10-12
Remove unnecessary allocation when scanning unused columns

letFunny:
We need to add pointers for all of the columns returned by the query but we only care about the ones that contain output expressions. For those unused, we use sql.RawBytes that avoids extra allocations.

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -150,9 +150,11 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 	for _, column := range columns {
 		idx, ok := markerIndex(column)
 		if !ok {
-			// Columns not mentioned in output expressions are scanned into x.
-			var x any
-			ptrs = append(ptrs, &x)
+			// Columns not mentioned in output expressions are scanned into a
+			// dummy variable because sql.Scan needs pointers for all columns.
+			// We use sql.RawBytes to avoid unnecesary allocations.
+			var dummy sql.RawBytes
+			ptrs = append(ptrs, &dummy)
 			continue
 		}
 		if idx >= len(qe.outputs) {


internal/expr/query.go
```diff
@@ -150,9 +150,11 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
 	for _, column := range columns {
 		idx, ok := markerIndex(column)
 		if !ok {
-			// Columns not mentioned in output expressions are scanned into x.
-			var x any
-			ptrs = append(ptrs, &x)
+			// Columns not mentioned in output expressions are scanned into a
+			// dummy variable because sql.Scan needs pointers for all columns.
+			// We use sql.RawBytes to avoid unnecesary allocations.
+			var dummy sql.RawBytes
```
niemeyer, line 156: If we don't care about the result, any reason not to use an empty struct here (`struct{}`)?

niemeyer (commented):
Sounds reasonable. Just a quick question:

internal/expr/query.go
letFunny: Per the [documentation](https://pkg.go.dev/database/sql#Rows.Scan) you need to provide  `Scan` with one of the following values:
```
*string
*[]byte
*int, *int8, *int16, *int32, *int64
*uint, *uint8, *uint16, *uint32, *uint64
*bool
*float32, *float64
*interface{}
*RawBytes
*Rows (cursor value)
any type implementing Scanner (see Scanner docs)
```
and, if the provided container is compatible, it will **copy** the data to that container. The only way of avoiding copying the data is to use a pointer to `RawBytes` which uses the internal byte buffer owned by the sql library which is discarded after requesting the next row. If we passed an `interface{}` instead, it will always be a valid conversion but it will copy the data as well.

niemeyer: If it doesn't allow the empty struct alone, we can do an empty struct that implements the Scanner interface, and does nothing with the argument.

Otherwise we're replacing one issue by the next. If you look in convert.go, you'll notice that RawBytes actually causes conversions and allocations in several cases. So it's arguably not much better than using `any` as far as allocations go.

letFunny: You are right, if you look at the source code the allocation cost is actually pretty similar and that is without taking into account that RawBytes holds a lock for longer.

I am tempted to go for the empty struct with the Scanner as the sql library only does one call without assignments ([ref](https://cs.opensource.google/go/go/+/master:src/database/sql/convert.go;l=385-387)) but it is still not clear to me if there is any benefit because we are still boxing the type to convert it to the Scanner interface requiring one allocation.

I feel this PR would actually be a form of premature optimization for not a clear win so I am closing it until the need arises.

letFunny:
> You are right, if you look at the source code the allocation cost is actually pretty similar and that is without taking into account that RawBytes holds a lock for longer.
> 
> I am tempted to go for the empty struct with the Scanner as the sql library only does one call without assignments ([ref](https://cs.opensource.google/go/go/+/master:src/database/sql/convert.go;l=385-387)) but it is still not clear to me if there is any benefit because we are still boxing the type to convert it to the Scanner interface requiring one allocation.
>
> I feel this PR would actually be a form of premature optimization for not a clear win so I am closing it until the need arises.

https://github.com/canonical/sqlair/pull/113#discussion_r1369944251

PR #114, opened 2023-10-12
feat: parse slice expressions

letFunny:
Introduces the parsing scaffolding that will eventually allow SQLair to support slices as input expressions. This PR adds support for parsing slice expressions via a range of the form 'Slice[:]'.


internal/expr/expr_test.go
```diff
@@ -411,6 +411,18 @@ comment */
...
+	}, {
+		query: "SELECT &S[0] FROM t",
+		err:   `cannot parse expression: column 8: cannot use slice syntax "S[0]" in output expression`,
+	}, {
+		query: "SELECT &S[1:5] FROM t",
+		err:   `cannot parse expression: column 8: cannot use slice syntax "S[1:5]" in output expression`,
+	}, {
+		query: "SELECT col1 AS &S[1:5] FROM t",
+		err:   `cannot parse expression: column 16: cannot use slice syntax "S[1:5]" in output expression`,
 	}}
```
Aflynn50, line 494: It looks like you're missing tests for a few of the error messages in `parseSliceAccessor`. These one's only relate to output expressions.

internal/expr/expr_test.go
```diff
@@ -411,6 +411,18 @@ comment */
 string
 of three lines' AND id = $Person.*`,
 		err: `cannot parse expression: line 3, column 26: asterisk not allowed in input expression "$Person.*"`,
+	}, {
+		query: "SELECT &S[:] FROM t",
+		err:   `cannot parse expression: column 8: cannot use slice syntax "S[:]" in output expression`,
+	}, {
+		query: "SELECT &S[0] FROM t",
+		err:   `cannot parse expression: column 8: cannot use slice syntax "S[0]" in output expression`,
```
Aflynn50: Is there any need for this case? It is testing an invalid form of slice syntax in an invalid context. The other tests should already cover this case.

internal/expr/parser.go
```diff
@@ -17,17 +18,33 @@ type queryPart interface {
...
-// valueAccessor stores information for accessing a Go value. It consists of a
-// type name and some value within it to be accessed. For example: a field of a
-// struct, or a key of a map.
-type valueAccessor struct {
+// valueAccessor stores information about how to access a Go value. For example,
+// by index, by key, or by slice syntax.
+type valueAccessor interface {
+	TypeName() string
+
+	// marker method
```
Aflynn50: Maybe change to `accessor is a marker method.`

internal/expr/parser.go
```diff
@@ -17,17 +18,33 @@ type queryPart interface {
...
+	accessor()
+}
+
+// keyedAccesor stores information for accessing a keyed Go value. It consists
```
Aflynn50: Is there a good reason to use `keyedAccessor` instead of `memberAccessor` as the name? I'm not married to the name `member` but unless there is a good reason I think it would make more sense here to be consistent across the codebase.

internal/expr/parser.go
```diff
@@ -17,17 +18,33 @@ type queryPart interface {
...
+func (va keyedAccesor) String() string {
 	return va.typeName + "." + va.memberName
 }
 
+func (va keyedAccesor) TypeName() string {
+	return va.typeName
+}
+
+func (va keyedAccesor) accessor() {
+}
```
Aflynn50, line 114: Can be on the line above

internal/expr/parser.go
```diff
@@ -40,6 +57,49 @@ func (ca columnAccessor) String() string {
...
+	}
+	strHigh := ""
+	if st.high != nil {
+		strHigh = strconv.FormatUint(*st.high, 10)
+	}
+	return fmt.Sprintf("%s[%s:%s]", st.typ, strLow, strHigh)
+}
+
+func (st sliceRangeAccessor) accessor() {
+}
```
Aflynn50: Can also be on the line above.

internal/expr/parser.go
```diff
@@ -502,35 +577,100 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+	return keyedAccesor{}, false, nil
+}
+
+// parseUNumber parses a non-negative number composed of one or more digits.
+func (p *Parser) parseUNumber() (*uint64, bool) {
+	mark := p.pos
+	if !p.skipNumber() {
+		return nil, false
+	}
+	n, _ := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
```
Aflynn50: I know it should be a number because we did `skipNumber` but it still feels weird to ignore the error here. Maybe we could panic with an internal error if the err is not nil?

internal/expr/parser.go
```diff
@@ -17,17 +18,33 @@ type queryPart interface {
...
+}
+
+// keyedAccesor stores information for accessing a keyed Go value. It consists
+// of a type name and some value within it to be accessed. For example: a field
+// of a struct, or a key of a map.
+type keyedAccesor struct {
 	typeName, memberName string
 }
 
-func (va valueAccessor) String() string {
```
Aflynn50, line 87: It may be better to implement the `valueAccessor` interface with the pointers rather than the pure struct. That way we're not copying them around all the time. Unless there is a good reason to not do that (such as using them in the key of a map/comparing them in places).

internal/expr/parser.go
```diff
@@ -40,6 +57,49 @@ func (ca columnAccessor) String() string {
...
 }
 
+// sliceRangeAccessor represents the expression typeName[low:high].
+type sliceRangeAccessor struct {
+	typ string
+	// Using pointers to represent that the both low and high are optional.
+	low, high *uint64
+}
+
+func (st sliceRangeAccessor) TypeName() string {
```
Aflynn50: Same point about implementing with pointers.

internal/expr/expr_test.go
letFunny: They are in the parse helper instead, I thought about having them in both places but I feel this way the code is less cluttered.

internal/expr/expr_test.go
letFunny: It is a valid form of slice syntax, just the index kind instead of the range one. I wanted to include one example for each of the broad categories.

internal/expr/parser.go
```diff
@@ -40,6 +57,49 @@ func (ca columnAccessor) String() string {
...
+func (st sliceIndexAccessor) TypeName() string {
+	return st.typ
+}
+
+func (st sliceIndexAccessor) String() string {
+	return fmt.Sprintf("%s[%d]", st.typ, st.index)
+}
+
+func (st sliceIndexAccessor) accessor() {
+}
```
Aflynn50: Also line above.

internal/expr/parser.go
```diff
@@ -502,35 +577,100 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+	if !p.skipNumber() {
+		return nil, false
+	}
+	n, _ := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
+	return &n, true
+}
+
+// parseSliceAccessor parses a slice range composed of two indexes of the form
+// "[low:high]"
+func (p *Parser) parseSliceAccessor() (sr valueAccessor, ok bool, err error) {
```
Aflynn50: `sr` is a little misleading because this can also be a `sliceIndex`. Do the return arguments need to be named here?

internal/expr/parser.go
```diff
@@ -502,35 +577,100 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+func (p *Parser) parseUNumber() (*uint64, bool) {
+	mark := p.pos
+	if !p.skipNumber() {
+		return nil, false
+	}
+	n, _ := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
+	return &n, true
+}
+
+// parseSliceAccessor parses a slice range composed of two indexes of the form
```
Aflynn50: It might be worth pointing out in the comment that it can return a `sliceIndex` or a `sliceRange`, since the `valueAccessor` is an interface.

internal/expr/parser.go
```diff
@@ -611,20 +751,19 @@ func (p *Parser) parseTargetTypes() (types []valueAccessor, parentheses bool, ok
...
 	} else if ok {
 		return &outputPart{
 			sourceColumns: []columnAccessor{},
-			targetTypes:   []valueAccessor{targetType},
+			targetTypes:   []keyedAccesor{targetType},
 			raw:           p.input[start:p.pos],
 		}, true, nil
 	}
 
-	cp := p.save()
```
Aflynn50: Is there a good reason to move this to the start here? If `parseTargetType` is doing what we expect this should be safe.

internal/expr/parser_helper_test.go
```diff
@@ -169,3 +175,65 @@ func (s *ExprInternalSuite) TestRemoveComments(c *C) {
 		}
 	}
 }
+
+func (s *ExprInternalSuite) TestParseSliceRange(c *C) {
```
Aflynn50: I think these tests would really fit better in `expr_test.go` rather than here. The current convention is that the `skip` functions are tested here with white box testing and and the `parse` functions are tested more a black box strategy in `expr_test`. These should really be done with a black box imo since they are public error messages.

internal/expr/parser.go
```diff
@@ -17,17 +18,33 @@ type queryPart interface {
...
 }
 
-// valueAccessor stores information for accessing a Go value. It consists of a
-// type name and some value within it to be accessed. For example: a field of a
-// struct, or a key of a map.
-type valueAccessor struct {
+// valueAccessor stores information about how to access a Go value. For example,
+// by index, by key, or by slice syntax.
+type valueAccessor interface {
+	TypeName() string
```
Aflynn50: Should this be `typeName`? i.e. not exported?

internal/expr/prepare.go
```diff
@@ -88,15 +88,24 @@ func prepareInput(ti typeNameToInfo, p *inputPart) (tm typeMember, err error) {
...
+		return nil, typeMissingError(p.sourceType.TypeName(), getKeys(ti))
 	}
 
-	tm, err = info.typeMember(p.sourceType.memberName)
-	if err != nil {
-		return nil, err
+	switch t := p.sourceType.(type) {
+	case keyedAccesor:
+		if t.memberName == "*" {
+			return nil, fmt.Errorf(`asterisk used with %s in invalid context`, info.typ().Kind())
```
Aflynn50: Is this case ever hit? Don't we still have the check for an asterisk in `parseInputExpr`?

internal/expr/prepare.go
```diff
@@ -88,15 +88,24 @@ func prepareInput(ti typeNameToInfo, p *inputPart) (tm typeMember, err error) {
...
+		}
+		tm, err = info.typeMember(t.memberName)
+		if err != nil {
+			return nil, err
+		}
+	case sliceRangeAccessor:
+		return nil, fmt.Errorf("internal error: slice support not implemented")
```
Aflynn50: imo this doesn't need the internal error prefix, this is something we want to tell the user if they encounter it.

Aflynn50 (commented):
Looking very good!

internal/expr/expr_test.go
Aflynn50: My mistake for pressing add single comment instead of start review, though I've left a comment in the review relating to this.

internal/expr/expr_test.go
Aflynn50: Fair enough, would be good to include all the other error messages we can trigger here as well.

internal/expr/parser.go
letFunny: None other than I wanted to use a verb at the beginning of the name, I will change it to `memberAccesor`.

internal/expr/parser.go
letFunny: Good idea, I think it makes it clearer

internal/expr/parser.go
letFunny: The problem with using a pointer is that I also have to allocate the struct on the heap if we are returning a `valueAccessor` from a function, in this case from `parseSliceAccessor`. And, I would have to check that the returned pointer is not null before using it. Do you still think pointer receiver is better even with theses caveats?

internal/expr/parser.go
letFunny: You are right

internal/expr/parser.go
letFunny: None, it was a remnant of the previous layout of the code. I left it there just for consistency of having the checkpoints at the top of the file, but I will leave it out of this PR.

internal/expr/parser_helper_test.go
letFunny: I think the best thing would be to move the parsing error tests and keep the other ones here as they are directly validating the output struct, what do you think?

This had the advantage of testing the correct parsing directly and being able to get the values back. We will also have tests that are more end-to-end including slices.

internal/expr/parser.go
letFunny: The benefit of having it in upper case is that it does not clash with the field name `typeName` of `memberAccessor`, and following the same naming convention as `queryPart` in the same file which has the `String` method.

internal/expr/prepare.go
letFunny: You are right, removing it here. That is the downside of this type system, I am not sure if we have to check or not without following the logic in the code.

internal/expr/prepare.go
letFunny: True. FYI, at the moment we never reach this code because prepare errors out if you pass it a slice.

internal/expr/expr_test.go
letFunny: I think these two are the only error messages we can trigger for output expressions, the rest are already tested in the ParseErrors bit, at least until we add support for slices in prepare/query.

internal/expr/parser.go
Aflynn50: The `String` method is public so that it implements the `Stringer` interface https://go.dev/tour/methods/17. That type is used by the `fmt` package when printing things.
To me it feels wrong to use a public method here, it implies that there is a reason for it. You could change the field name in `memberAccessor` to `outerType` maybe? As it is in `typeMember`. Its a shame they decided not to allow interfaces to be implemented by struct fields in Go, it would help a lot here.

internal/expr/parser.go
```diff
@@ -601,13 +598,17 @@ func (p *Parser) parseUNumber() (*uint64, bool) {
 	if !p.skipNumber() {
 		return nil, false
 	}
-	n, _ := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
+	n, err := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
+	if err != nil {
+		panic("internal error: skipNumber did not return a valid number")
```
Aflynn50: Maybe put `skip a valid number` rather than `return a valid number`

internal/expr/parser.go
Aflynn50: I think it would be fine not to be checking for nil for this value, we would only return a nil pointer to one in an error case. As for assigning on the heap, I'm not sure what would happen exactly in the case/version of the compiler, but interface variables with an interface type go on the heap according to this https://research.swtch.com/interfaces.

I think its fair enough to not change this because of the extra complexity that it would add, especially since it contains a very small amount of information.

internal/expr/parser.go
Aflynn50: Cool ok. There was a discussion a while back about weather to keep the checkpoints consistent in each function (i.e. always start and end) or weather to do the save as late as possible. It was decided that we should generally do the most efficient option and leave the save as late as we can.

internal/expr/parser_helper_test.go
Aflynn50: That's fair, I think this layout makes sense.

internal/expr/prepare.go
Aflynn50: Yea, I would like at some point to tighten up the types in parser so that information like this does not have to be re-validated in the prepare stage and is instead stored in parser types.

internal/expr/prepare.go
Aflynn50: Ah yea, good point.

internal/expr/parser.go
letFunny: After all the discussions we had on how to name the accessors, I feel that they are now looking good and I don't want to change it back to `outerType`. I have a few ideas, what do you think of the following?
1. Using `getTypeName` for the interface. Even if it is discouraged in general, it would make sense in this case and there are examples of using the `get` prefix for name clashes like protobuf.
2. Changing the field name to `outerTypeName`.
3. Changing the field name to `typName`.

Aflynn50: Using `getTypeName` makes the most sense here imo.

internal/expr/parser.go
```diff
@@ -502,35 +574,104 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+
+// parseSliceAccessor parses a slice range composed of two indexes of the form
+// "[low:high]". On success it returns either a sliceIndexAccessor or a
+// sliceRangeAccessor.
+func (p *Parser) parseSliceAccessor() (va valueAccessor, ok bool, err error) {
+	cp := p.save()
+
+	id, ok := p.parseIdentifier()
+	if !ok {
+		return sliceRangeAccessor{}, false, nil
```
Aflynn50: You can return `nil` here instead of sliceRangeAccessor. Interfaces can be `nil` and you are returning a `valueAccessor`.

You can also do it for all the other error returns here.

internal/expr/parser.go
```diff
@@ -502,35 +574,104 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+			return memberAcessor{}, false, errorAt(fmt.Errorf("cannot use slice syntax %q in output expression", st), startLine, startCol, p.input)
+		}
 		return p.parseTypeName()
 	}
 
-	return valueAccessor{}, false, nil
+	return memberAcessor{}, false, nil
+}
+
+// parseUNumber parses a non-negative number composed of one or more digits.
```
Aflynn50: Maybe worth pointing out why this value is a pointer rather than a straight integer here. Or changing it to return a plain integer and doing the pointer bits in the call site.

internal/expr/parser.go
```diff
@@ -17,17 +18,32 @@ type queryPart interface {
...
+type memberAcessor struct {
 	typeName, memberName string
 }
 
-func (va valueAccessor) String() string {
+func (va memberAcessor) String() string {
 	return va.typeName + "." + va.memberName
 }
 
+func (va memberAcessor) TypeName() string {
```
Aflynn50: Should these variable names be `ma` not `va`?

internal/expr/parser.go
```diff
@@ -17,17 +18,32 @@ type queryPart interface {
...
+type memberAcessor struct {
 	typeName, memberName string
 }
 
-func (va valueAccessor) String() string {
+func (va memberAcessor) String() string {
 	return va.typeName + "." + va.memberName
 }
 
+func (va memberAcessor) getTypeName() string {
```
Aflynn50: Should this be `ma` not `va`?

internal/expr/parser.go
```diff
@@ -17,17 +18,32 @@ type queryPart interface {
 	part()
 }
 
-// valueAccessor stores information for accessing a Go value. It consists of a
-// type name and some value within it to be accessed. For example: a field of a
-// struct, or a key of a map.
-type valueAccessor struct {
+// valueAccessor stores information about how to access a Go value. For example,
```
Aflynn50: Colon after "For example"

Aflynn50 (commented):
Amazing, pretty much ready to go.

internal/expr/parser.go
```diff
@@ -40,6 +56,47 @@ func (ca columnAccessor) String() string {
 	return ca.tableName + "." + ca.columnName
 }
 
+// sliceRangeAccessor represents the expression typeName[low:high].
+type sliceRangeAccessor struct {
+	typ string
+	// Using pointers to represent that the both low and high are optional.
+	low, high *uint64
```
letFunny: [INFO FOR REVIEWER] These should really be `Optional<uint64>` but Go does not have support for such types. There are several approaches:
1. Store `low` and `high` as numbers and have two flags to indicate whether they are empty or not.
2. Use pointers which are nil if the values are empty.

The advantage of the second approach is that it is enforced by the type system, i.e. you cannot access the number without dereferencing the pointer, thus being reminded to check for null. In the first one, you might forget to check the flag and use the zero value instead.

manadart:
Can we get this one rebased?

The description could do with more details, including the kind of statements it enables support for.

letFunny:
> Can we get this one rebased?
> 
> The description could do with more details, including the kind of statements it enables support for.

@manadart I rebased the PR and also added some more comments in the description for the long term goal of supporting slices as input expressions. Let me know if that is enough detail.

internal/expr/parser.go
```diff
@@ -77,17 +78,32 @@ type expression interface {
...
 }
 
-// valueAccessor stores information for accessing a Go value. It consists of a
-// type name and some value within it to be accessed. For example: a field of a
-// struct, or a key of a map.
-type valueAccessor struct {
+// valueAccessor stores information about how to access a Go value. For example:
+// by index, by key, or by slice syntax.
+type valueAccessor interface {
+	getTypeName() string
```
manadart, line 84: The only implementations of this appear to be property accessors. The Go convention would in that case be `typeName()`.

internal/expr/parser.go
```diff
@@ -77,17 +78,32 @@ type expression interface {
...
+
+	// accessor is a marker method.
+	accessor()
```
manadart: I really don't want to add more of these. A group of types either has a common indirection or they don't.

In this case, there are no implementers of `getTypeName` other than those that also implement `accessor` so it can just be deleted.

You could add `String` though, or embed `Stringer`.

internal/expr/parser.go
```diff
@@ -100,6 +116,49 @@ func (ca columnAccessor) String() string {
 	return ca.tableName + "." + ca.columnName
 }
 
+// sliceRangeAccessor stores information for accessing a slice using the
+// expression "typeName[low:high]".
+type sliceRangeAccessor struct {
+	typ string
```
manadart: I'd be inclined to use `typeName` as you have for `memberAccessor`. Same in `sliceIndexAccessor` below.

internal/expr/bindtypes.go
```diff
@@ -138,15 +138,21 @@ func bindInputTypes(ti typeNameToInfo, e *inputExpr) (tm typeinfo.Member, err er
...
-	if err != nil {
-		return nil, err
+	switch t := e.sourceType.(type) {
+	case memberAcessor:
+		tm, err = info.TypeMember(t.memberName)
+		if err != nil {
+			return nil, err
+		}
+	case sliceRangeAccessor:
+		return nil, fmt.Errorf("slice support not implemented")
```
manadart, line 98: How do you envision this working ultimately?

We should either use type switching or interface indirection. A mix of both is not cogent.

internal/expr/parser.go
```diff
@@ -77,17 +78,32 @@ type expression interface {
...
+
+// memberAcessor stores information for accessing a keyed Go value. It consists
+// of a type name and some value within it to be accessed. For example: a field
+// of a struct, or a key of a map.
+type memberAcessor struct {
 	typeName, memberName string
 }
 
-func (va valueAccessor) String() string {
+func (va memberAcessor) String() string {
```
manadart: Differing receiver names.

internal/expr/parser.go
```diff
@@ -117,7 +176,7 @@ func (p *inputExpr) expr() {}
 // as well as the source table and column where it will be read from.
 type outputExpr struct {
 	sourceColumns []columnAccessor
-	targetTypes   []valueAccessor
+	targetTypes   []memberAcessor
```
manadart: Even though these aren't exported, comments would be beneficial.

internal/expr/parser.go
```diff
@@ -473,35 +547,106 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+
+// parseUNumber parses a non-negative number composed of one or more digits.
+func (p *Parser) parseUNumber() (uint64, bool) {
+	mark := p.pos
+	if !p.skipUNumber() {
+		return 0, false
+	}
+	n, err := strconv.ParseUint(p.input[mark:p.pos], 10, 64)
+	if err != nil {
+		panic("internal error: skipUNumber did not skip a valid number")
```
manadart: We don't want to panic in a library. This could be quite disruptive for a Juju controller for example.

manadart (commented):
Some comments from a first pass.

internal/expr/bindtypes.go
letFunny: Well for this PR I wanted to make it as small as possible and to stop the flow of execution early. You are right in the future this will be a method on the `sliceRangeAccessor`. I can remove the type switching here and add the method which will only return an error.

internal/expr/parser.go
letFunny: It is only prefixed by `get` because it clashes with the `typeName` field of `memberAccessor`. The problem is it is difficult to come up with a better name for the field and/or the method.

internal/expr/parser.go
letFunny: In the initial implementation I was just ignoring the error because it meant that `skipUNumber` was failing internally, but after a few revisions of this PR I changed it to a panic to be explicit about why we were ignoring the error. The other option would be to return `(int, bool, error)` which seems a little bit overkill. What do you think is the best approach?

internal/expr/bindtypes.go
letFunny: My idea was to limit this PR to the bare minimum to get the parsing done. In this case I think it is useful to stop execution here and, in the next PR, we can rework the type information functions (e.g. at the moment `TypeMember` is the entrypoint but it expects the name of a column, a notion which does not translate to slices).

If you agree I would like to postpone this discussion until the next PR (which Alastair is writing at the moment) and then, we can define the correct abstraction.

letFunny:
Status update: we have changed the spec to reflect the fact that we want the implementation to be simpler initially. We can always support the extended syntax in the future when the need arises.

Additionally, we will put this PR on hold until the pipeline redesign is completed.

internal/expr/parser.go
```diff
@@ -100,44 +100,19 @@ func (ca columnAccessor) String() string {
...
 
-// outputExpr represents a named target output variable in the SQL expression,
-// as well as the source table and column where it will be read from.
-type outputExpr struct {
-	sourceColumns []columnAccessor
-	targetTypes   []valueAccessor
-	raw           string
+// sliceRangeAccessor stores information for accessing a slice using the
+// expression "typeName[:]".
+type sliceRangeAccessor struct {
```
Aflynn50: Should this be called sliceAccessor now since we are not doing ranges?

internal/expr/parser.go
```diff
@@ -68,26 +68,26 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
...
-	// String returns a text representation for debugging and testing purposes.
-	String() string
-
-	// marker method
-	expr()
+// valueAccessor stores information about how to access a Go value. For example:
+// by index, by key, or by slice syntax.
+type valueAccessor interface {
+	fmt.Stringer
+	valueAccessorMarker()
```
Aflynn50: What was the reason for removing the method `getTypeName` and having a marker method instead? There are some places where the type name is useful such as error messages.

internal/expr/parser.go
```diff
@@ -473,35 +453,68 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+		}
 		return p.parseTypeName()
 	}
 
-	return valueAccessor{}, false, nil
+	return memberAccessor{}, false, nil
+}
+
+// parseSliceAccessor parses a slice range composed of the form "[:]".
+func (p *Parser) parseSliceAccessor() (va sliceRangeAccessor, ok bool, err error) {
```
Aflynn50: [Nit] could change va to sra.

internal/expr/parser.go
```diff
@@ -473,35 +453,68 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+	p.skipBlanks()
+	if !p.skipByte(']') {
+		return sliceRangeAccessor{}, false, errorAt(fmt.Errorf("invalid slice: expected ']'"), p.lineNum, p.colNum(), p.input)
+	}
+	return sliceRangeAccessor{typeName: id}, true, nil
 }
 
 // parseTypeName parses a Go type name qualified by a tag name (or asterisk)
 // of the form "TypeName.col_name".
-func (p *Parser) parseTypeName() (valueAccessor, bool, error) {
```
Aflynn50, line 486: Maybe we could change this function name to something like `parseTypeMember`, or `parseOutputType` or something.

Aflynn50 (commented):
Just a few small things, will give another quick pass once its rebased

internal/expr/parser.go
letFunny: That simplifies it, we can always add it back in the future.

internal/expr/parser.go
letFunny: What would you return in `getTypeName` for `columnAccessor`?

internal/expr/parser.go
letFunny: What about `parseTypeMemberAccessor`?

internal/expr/parser.go
Aflynn50: Hang on, I don't think we want `columnAccessor` to implement `valueAccessor`. The value accessor should be the stuff on the right hand side of the output expression and the column accessor on the left e.g. in `(t.co1, p.col2) AS (&Person.id, &Person.address)` . A value accessor would always have a `&` or `$` before it and a column accessor would not.

internal/expr/parser.go
Aflynn50: Yea maybe, though what do you think of `parseTypeWithMember` or `parseTypeAndMember`? Since they say what it is a bit more and you can see the `memberAccessor` from the type

internal/expr/parser.go
letFunny: You are, in one of the merges `columnAccessor` was moved in between the other two and I mistakenly implemented the interface for that one. I have uploaded a new commit that reverts that.

letFunny: And also removed the marker method.

internal/expr/parser.go
letFunny: I like `parseTypeAndMember`, much better than what I suggested. Changed now!

internal/expr/expr_test.go
```diff
@@ -455,6 +458,39 @@ comment */
...
+		err:   `cannot parse expression: column 8: cannot use slice syntax in output expression`,
+	}, {
+		query: "SELECT col1 AS &S[1:5] FROM t",
+		err:   `cannot parse expression: column 16: cannot use slice syntax in output expression`,
+	}, {
+		query: "SELECT col1 AS &S[] FROM t",
+		err:   `cannot parse expression: column 16: cannot use slice syntax in output expression`,
+	}, {
+		query: "SELECT * FROM t WHERE id IN $ids[:-1]",
+		err:   `cannot parse expression: column 35: invalid slice: expected ']'`,
```
Aflynn50: Just as an idea, how would you feel about this error message being: `expected "$id[:]"`. I assume that when we throw the error we would have all the info to output something like that. Its also fine as it is though.

internal/expr/parser.go
```diff
@@ -473,35 +503,68 @@ func (p *Parser) parseColumn() (columnAccessor, bool, error) {
...
+	if !ok {
+		return sliceAccessor{}, false, nil
+	}
+	if !p.skipByte('[') {
+		cp.restore()
+		return sliceAccessor{}, false, nil
+	}
+	p.skipBlanks()
+	if !p.skipByte(':') {
+		return sliceAccessor{}, false, errorAt(fmt.Errorf("invalid slice: expected ':'"), p.lineNum, p.colNum(), p.input)
```
Aflynn50: This and the error below are the ones mentioned in the comment in expr_test.

Aflynn50 (approved):
Looks excellent

internal/expr/expr_test.go
letFunny: You are on fire lately! I like that suggestion. Because we now know there is only one accepted format that would make it clearer.

letFunny: I have changed the error message. The only confusing aspect is with whitespace because we are saying `expected 'slice[:]'` when in reality `'slice[ : ]'` is also accepted, do you think that is confusing?

manadart (approved):
Thanks for seeing all these changes through. This is good to iterate on.

PR #115, opened 2023-10-12
Update field names in expr_test.go

Aflynn50:
Add field names to the list of table driven tests for `TestExpr`.
Change the field names:
`expectedPrepared` -> `expectedSQL`
`prepareArgs` -> `typeSamples`
`queryArgs` -> `inputArgs`

Adding field names to the table driven tests means that tests which do not require every field can omit some (this is done in #112). It also aids readability.

The name changes keep the names consistent with recent changes elsewhere in the codebase.


manadart (approved):
This does read better; thanks.

PR #116, opened 2023-10-12
Move marker methods from prepare.go to query.go

Aflynn50:
Move the const `markerPrefix` and the methods `markerName` and `markerIndex` to `query.go`.

This change is motivated by PR #112. In it, the SQL will be generated in `query.go` so all calls to these functions will come from `query.go`.

### DIFF
--- a/internal/expr/prepare.go
+++ b/internal/expr/prepare.go
@@ -21,23 +21,6 @@ func (pe *PreparedExpr) SQL() string {
 	return pe.sql
 }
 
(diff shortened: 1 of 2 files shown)


niemeyer (changes requested):
Let's please hold this until #112 itself has been approved, as the underlying assumption is that logic is moving from PrepareExpr to QueryExpr.

Aflynn50:
Closed as this is covered by #123

PR #118, opened 2023-10-23
bugfix: Iter.Get check number of args to prevent panic

letFunny:
In `Iter.Get` we need to explicitly check the length of args before a check at the zero-index in order to prevent a panic.


Aflynn50 (commented):
LGTM

manadart:
Need a description on this one before we can merge it.

manadart (approved):

PR #119, opened 2023-10-26
Test for prepared statement leaks

letFunny:
Created wrapper driver to monitor statement creation and deletion while calling SQLite behind the scenes.

Before this change the tests did not flag when statements were not closed properly, for example if you commented [the following line](https://github.com/canonical/sqlair/blob/main/sqlair.go#L279) no tests would have failed. After this change, the tests are more robust and they can now verify that the cache and finalizers are working correctly. This is especially important now that we want to overhaul the cache implementation and we want to verify it is still correct.

### DIFF
--- a/driver_test.go
+++ b/driver_test.go
@@ -72,14 +72,18 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
 	return c.PrepareContext(context.Background(), query)
 }
 
-var testNameTag = "testName"
+const testNameTag = "testName"
 
 func (d *Driver) Open(name string) (driver.Conn, error) {
 	var testName string
-	for _, p := range strings.Split(name, "&") {
+	parameters := strings.Split(name, "?")[1]
+	for _, p := range strings.Split(parameters, "&") {
 		if strings.HasPrefix(p, testNameTag) {
 			testName = strings.Split(p, "=")[1]
 		}
+	}
+	if testName == "" {
+		panic("internal error: testName is not found in the db DSN")
 	}
 
 	baseConn, err := d.Driver.Open(name)
--- a/driver_test.go
+++ b/driver_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"database/sql/driver"
+	"strings"
 	"sync"
 	"unsafe"
 
@@ -19,8 +20,8 @@ import (
 // detect resource leaks. It uses pointers instead of references to the object
 // because if we stored a reference the runtime.Finalizer would not be able to
 // run.
-var openStmts map[uintptr]string = map[uintptr]string{}
-var closedStmts map[uintptr]bool = map[uintptr]bool{}
+var openStmts = map[string]map[uintptr]string{}
+var closedStmts = map[string]map[uintptr]bool{}
 var stmtRegistryMutex sync.RWMutex
 
 // Structs used for wrapping the underlying SQLite driver.
@@ -28,16 +29,22 @@ type Driver struct {
 	driver.Driver
 }
 type Conn struct {
+	testName string
 	*sqlite3.SQLiteConn
 }
 type Stmt struct {
+	testName string
 	*sqlite3.SQLiteStmt
 }
 
 func (s *Stmt) Close() error {
 	stmtRegistryMutex.Lock()
-	closedStmts[uintptr(unsafe.Pointer(s))] = true
-	stmtRegistryMutex.Unlock()
+	defer stmtRegistryMutex.Unlock()
+	_, ok := closedStmts[s.testName]
+	if !ok {
+		closedStmts[s.testName] = map[uintptr]bool{}
+	}
+	closedStmts[s.testName][uintptr(unsafe.Pointer(s))] = true
 
 	return s.SQLiteStmt.Close()
 }
@@ -45,11 +52,15 @@ func (s *Stmt) Close() error {
 func (c *Conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
 	s, err := c.SQLiteConn.PrepareContext(ctx, query)
 	if s, ok := s.(*sqlite3.SQLiteStmt); ok {
-		sPtr := &Stmt{s}
+		sPtr := &Stmt{SQLiteStmt: s, testName: c.testName}
 
 		stmtRegistryMutex.Lock()
-		openStmts[uintptr(unsafe.Pointer(sPtr))] = query
-		stmtRegistryMutex.Unlock()
+		defer stmtRegistryMutex.Unlock()
+		_, ok := openStmts[c.testName]
+		if !ok {
+			openStmts[c.testName] = map[uintptr]string{}
+		}
+		openStmts[c.testName][uintptr(unsafe.Pointer(sPtr))] = query
 
 		return sPtr, err
 	} else {
@@ -61,13 +72,22 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
 	return c.PrepareContext(context.Background(), query)
 }
 
+var testNameTag = "testName"
+
 func (d *Driver) Open(name string) (driver.Conn, error) {
+	var testName string
+	for _, p := range strings.Split(name, "&") {
+		if strings.HasPrefix(p, testNameTag) {
+			testName = strings.Split(p, "=")[1]
+		}
+	}
+
 	baseConn, err := d.Driver.Open(name)
 	if err != nil {
 		return nil, err
 	}
 	if baseConn, ok := baseConn.(*sqlite3.SQLiteConn); ok {
-		return &Conn{baseConn}, err
+		return &Conn{SQLiteConn: baseConn, testName: testName}, err
 	} else {
 		panic("internal error: base driver is not SQLite")
 	}
--- a/driver_test.go
+++ b/driver_test.go
@@ -16,10 +16,10 @@ import (
 // references to said statements. We can later use that information to check
 // for statement leaks.
 
-// The stmt registry keeps the pointers for the open and closed statements to
-// detect resource leaks. It uses unsafe pointers instead of references to the
-// object because if we stored a reference the runtime.Finalizer would not be
-// able to run.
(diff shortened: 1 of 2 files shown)


driver_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"sync"
+	"unsafe"
+
+	"github.com/mattn/go-sqlite3"
+)
+
+// This file contains a wrapper sql.Driver over the SQLite driver which does
```
Aflynn50: No need to say "which does the real work behind the scenes". You can just say what the wrapper does.

driver_test.go
```diff
@@ -0,0 +1,80 @@
...
+// the real work behind the scenes. On top of it, it monitors the creation and
+// closing of prepared statements and stores the references to said statements.
+// We can later use that information to check for leaks.
```
Aflynn50: Maybe specify memory leaks.

driver_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+// The stmt registry keeps the pointers for the open and closed statements to
+// detect resource leaks. It uses pointers instead of references to the object
```
Aflynn50: Maybe say it uses "unsafe pointers" just to make it clear that this is the justification for using `unsafe`.

package_test.go
```diff
@@ -22,8 +22,29 @@ type PackageSuite struct{}
...
+		runtime.GC()
+		time.Sleep(0)
+	}
+
+	stmtRegistryMutex.Lock()
+	defer stmtRegistryMutex.Unlock()
+
+	// Asssert that all the open statements were closed.
+	for sPtr, query := range openStmts {
+		c.Check(closedStmts[sPtr], Equals, true,
```
Aflynn50: Its a bit weird that this checks that "open statements" are "closed", maybe `openStmts` could be renamed?

driver_test.go
```diff
@@ -0,0 +1,80 @@
...
+// because if we stored a reference the runtime.Finalizer would not be able to
+// run.
+var openStmts map[uintptr]string = map[uintptr]string{}
+var closedStmts map[uintptr]bool = map[uintptr]bool{}
```
Aflynn50: Having a global variable here feels very dangerous. I believe this is working right now because go check executes the unit tests sequentially. Go test actually runs them concurrently, so if we did ever switch to go test I believe this would stop working. 

It would be good if it was possible to attach this to the DB (still by overwriting the methods of the driver), since that is the scope we care about in the vast majority of cases, though it may have to be to individual connections since that seems to be the language the driver speaks.

You could maybe have some checks that are global on the Suite to check that the statements are closed when all DBs are closed as expected but you would want to make sure it was concurrent safe.

Aflynn50 (commented):
This is a much better approach than what is currently in the tests. Once this is in we should follow it up with a PR to modify the tests for the prepared statement cache with this approach.

driver_test.go
letFunny: I will mention statement leaks.

driver_test.go
letFunny: I agree with you, because go-check does not say explicitly that it runs the tests sequentially it is better if we change the driver to support it.

Just like we discussed offline, I have added the test name in the DSN when creating a DB connection and that gets propagated to the statement opening and closing functions.

driver_test.go
```diff
@@ -61,13 +72,22 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
 	return c.PrepareContext(context.Background(), query)
 }
 
+var testNameTag = "testName"
```
Aflynn50: You could use `const` here

driver_test.go
```diff
@@ -61,13 +72,22 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
...
+
 func (d *Driver) Open(name string) (driver.Conn, error) {
+	var testName string
+	for _, p := range strings.Split(name, "&") {
+		if strings.HasPrefix(p, testNameTag) {
+			testName = strings.Split(p, "=")[1]
+		}
+	}
+
```
Aflynn50, line 90: It would be good to panic here if the testName is not found, or we will start getting weird problems where the empty test name is used by multiple tests if one does not pass the test name in the argument.

driver_test.go
```diff
@@ -61,13 +72,22 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
 	return c.PrepareContext(context.Background(), query)
 }
 
+var testNameTag = "testName"
+
 func (d *Driver) Open(name string) (driver.Conn, error) {
+	var testName string
+	for _, p := range strings.Split(name, "&") {
```
Aflynn50: If the option we're looking for is the first one then it won't be found

Aflynn50 (commented):
Very nice!

driver_test.go
letFunny: Agree, it is better to explicitly fail and if we don't want to use this for a test then add a flag or something. This way all tests will be using it by default.

driver_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+// This file contains a wrapper sql.Driver over the SQLite driver which
+// monitors the creation and closing of prepared statements and stores the
+// references to said statements. We can later use that information to check
+// for statement leaks.
+
+// The stmt registry keeps the pointers for the open and closed statements to
+// detect resource leaks. It uses unsafe pointers instead of references to the
+// object because if we stored a reference the runtime.Finalizer would not be
+// able to run.
```
Aflynn50: This comment should explain why `openedStmts` and `closedStmts` are maps of maps (i.e. because they are per test).

driver_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/mattn/go-sqlite3"
+)
+
+// This file contains a wrapper sql.Driver over the SQLite driver which
+// monitors the creation and closing of prepared statements and stores the
+// references to said statements. We can later use that information to check
+// for statement leaks.
+
+// The stmt registry keeps the pointers for the open and closed statements to
```
Aflynn50: Maybe define the stmt registery somewhere or rephrase this, its not immediately obvious what it is.

driver_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func (c *Conn) Prepare(query string) (driver.Stmt, error) {
+	return c.PrepareContext(context.Background(), query)
+}
+
+const testNameTag = "testName"
+
+func (d *Driver) Open(name string) (driver.Conn, error) {
```
Aflynn50, line 79: I think it would be good to document that we pass the test name in the DSN here

package_test.go
```diff
@@ -22,25 +22,40 @@ type PackageSuite struct{}
 
 var _ = Suite(&PackageSuite{})
 
-func setupDB() (*sql.DB, error) {
-	return sql.Open("sqlite3", "file:test.db?cache=shared&mode=memory")
-}
+func (s *PackageSuite) TearDownTest(c *C) {
+	// Try to run finalizers by calling GC several times.
+	for i := 0; i <= 10; i++ {
```
Aflynn50, line 27: It should technically only need running twice if the value is out of scope (once to garbage collect it and the second time to run the actual finalizer). But I guess there's no harm in running it a few more times.

package_test.go
```diff
@@ -22,25 +22,40 @@ type PackageSuite struct{}
...
+	stmtRegistryMutex.Lock()
+	defer stmtRegistryMutex.Unlock()
+
+	// Asssert that all the open statements were closed.
+	for sPtr, query := range openedStmts[c.TestName()] {
+		c.Check(closedStmts[c.TestName()][sPtr], Equals, true,
+			Commentf("%s: failed to close statement: %s", c.TestName(), query))
 	}
 
+	// Reset state for next test.
```
Aflynn50: Should we still be doing this if these maps are indexed by the tests? If tests are run in parallel for whatever reason I think this could break it.

They could be cleared on suite teardown though.

Aflynn50 (commented):
I think there were a few things missed when updating the statement registry to be indexed by test.

package_test.go
letFunny: I was treating the finalizers as an implementation detail, hence why running them more times than needed to avoid breaking the tests in the future.

package_test.go
letFunny: You are completely right

driver_test.go
```diff
@@ -0,0 +1,106 @@
...
+var closedStmts = map[string]map[uintptr]bool{}
+var stmtRegistryMutex sync.RWMutex
+
+// Structs used for wrapping the underlying SQLite driver.
+type Driver struct {
+	driver.Driver
+}
+type Conn struct {
+	testName string
+	*sqlite3.SQLiteConn
```
manadart, line 33: I prefer to have embedded things first to make behaviour clear.

package_test.go
```diff
@@ -22,25 +22,45 @@ type PackageSuite struct{}
...
+	}
 
-func createExampleDB(createTables string, inserts []string) (*sql.DB, error) {
-	db, err := setupDB()
-	if err != nil {
-		return nil, err
+	stmtRegistryMutex.Lock()
+	defer stmtRegistryMutex.Unlock()
+
+	// Asssert that all the open statements were closed.
```
manadart, line 35: Typo.

driver_test.go
```diff
@@ -0,0 +1,106 @@
...
+// This file contains a wrapper sql.Driver over the SQLite driver which
+// monitors the creation and closing of prepared statements and stores the
+// references to said statements. We can later use that information to check
+// for statement leaks.
+
+// The maps below store the pointers to the created/closed statements indexed
+// by test case. We use unsafe pointers instead of references to the objects
+// because if we stored a reference the runtime.Finalizer would not be able to
+// run.
+var openedStmts = map[string]map[uintptr]string{}
```
manadart, line 23: It would be nice to have these as members of the suite itself, but I can't see an easy way to supply them to the driver without repeating register/unregister.

manadart (approved):
Nice addition. Let's get this in.

PR #120, opened 2023-11-15
Add year and company to the license

Aflynn50:
Replace the placeholders in the Apache 2.0 license with the year and the company name.


manadart (approved):
We should also add this as a header to each of the source files. See Juju as an example.

PR #121, opened 2023-11-16
Separate concerns relating to reflection and type information

manadart:
SQLair has an `internal` package, but under that a single `expr` package into which everything is lumped. 

This has allowed accretion of logic with no clear delineation of concerns, lack of encapsulation, no principle of least knowledge et al.

Here we create a new package `typeinfo` into which logic around reflection and type information is relocated. In particular:
- The exposed API surface of this logic is now apparent.
- `typeinfo.Member` (previously `typeMember`) implementations assume responsibility for returning their corresponding values within parents, and for generating scan targets. This fixes query and prepare logic so that we can rely on the indirection and eschew type switching, which violates responsibility bounds.

### DIFF
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -97,8 +97,14 @@ func (pq *PrimedQuery) ScanArgs(columnNames []string, outputArgs []any) (scanArg
 			return nil, nil, fmt.Errorf("type %q found in query but not passed to get", typeMember.OuterType().Name())
 		}
 
-		if ptrs, scanProxies, err = typeMember.AddScanTarget(outputVal, ptrs, scanProxies); err != nil {
+		ptr, scanProxy, err := typeMember.GetScanTarget(outputVal)
+		if err != nil {
 			return nil, nil, err
+		}
+
+		ptrs = append(ptrs, ptr)
+		if scanProxy != nil {
+			scanProxies = append(scanProxies, *scanProxy)
 		}
 	}
 
--- a/internal/expr/query.go
+++ b/internal/expr/query.go
@@ -1,10 +1,11 @@
 package expr
 
 import (
-	"database/sql"
 	"fmt"
 	"reflect"
 	"strings"
+
+	"github.com/canonical/sqlair/internal/typeinfo"
 )
 
 // PrimedQuery contains all concrete values needed to run a SQLair query on a
@@ -14,7 +15,7 @@ type PrimedQuery struct {
 	// params are the query parameters to pass to the database.
 	params []any
 	// outputs specifies where to scan the query results.
-	outputs []typeMember
+	outputs []typeinfo.Member
 }
 
 // Params returns the query parameters to pass with the SQL to a database.
@@ -28,32 +29,23 @@ func (pq *PrimedQuery) HasOutputs() bool {
 	return len(pq.outputs) > 0
 }
 
-var scannerInterface = reflect.TypeOf((*sql.Scanner)(nil)).Elem()
-
 // ScanArgs produces a list of pointers to be passed to rows.Scan. After a
 // successful call, the onSuccess function must be invoked. The outputArgs will
 // be populated with the query results. All the structs/maps/slices mentioned in
 // the query must be in outputArgs.
 func (pq *PrimedQuery) ScanArgs(columnNames []string, outputArgs []any) (scanArgs []any, onSuccess func(), err error) {
-	var typesInQuery = []string{}
+	var typesInQuery []string
 	var inQuery = make(map[reflect.Type]bool)
 	for _, typeMember := range pq.outputs {
-		outerType := typeMember.outerType()
+		outerType := typeMember.OuterType()
 		if ok := inQuery[outerType]; !ok {
 			inQuery[outerType] = true
 			typesInQuery = append(typesInQuery, outerType.Name())
 		}
 	}
 
-	type scanProxy struct {
-		original reflect.Value
-		scan     reflect.Value
-		key      reflect.Value
-	}
-	var scanProxies []scanProxy
-
 	var typeDest = make(map[reflect.Type]reflect.Value)
-	outputVals := []reflect.Value{}
+	var outputVals []reflect.Value
 	for _, outputArg := range outputArgs {
 		if outputArg == nil {
 			return nil, nil, fmt.Errorf("need map or pointer to struct, got nil")
@@ -84,7 +76,8 @@ func (pq *PrimedQuery) ScanArgs(columnNames []string, outputArgs []any) (scanArg
 	}
 
 	// Generate the pointers.
-	var ptrs = []any{}
+	var ptrs []any
+	var scanProxies []typeinfo.ScanProxy
 	var columnInResult = make([]bool, len(columnNames))
 	for _, column := range columnNames {
 		idx, ok := markerIndex(column)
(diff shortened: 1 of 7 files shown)


internal/expr/query.go
```diff
@@ -61,61 +63,48 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
-		switch tm := typeMember.(type) {
-		case *structField:
-			val = v.Field(tm.index)
-		case *mapKey:
-			val = v.MapIndex(reflect.ValueOf(tm.name))
-			if val.Kind() == reflect.Invalid {
-				return nil, fmt.Errorf(`map %q does not contain key %q`, outerType.Name(), tm.name)
-			}
+
+		val, err := typeMember.ValueForOuter(v)
```
manadart: This is one key part of the cleanup. If we have types with a unifying indirection, we should use that indirection to limit the knowledge required by other types.

internal/expr/query.go
```diff
@@ -161,53 +151,25 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
...
-				scanProxies = append(scanProxies, scanProxy{original: val, scan: scanVal})
-			} else {
-				ptrs = append(ptrs, val.Addr().Interface())
-			}
-		case *mapKey:
-			scanVal := reflect.New(tm.mapType.Elem()).Elem()
-			ptrs = append(ptrs, scanVal.Addr().Interface())
-			scanProxies = append(scanProxies, scanProxy{original: outputVal, scan: scanVal, key: reflect.ValueOf(tm.name)})
+
+		if ptrs, scanProxies, err = typeMember.AddScanTarget(outputVal, ptrs, scanProxies); err != nil {
```
manadart: Here is the other.

internal/typeinfo/member.go
```diff
@@ -0,0 +1,99 @@
...
+	OuterType() reflect.Type
+
+	// MemberName returns this member's name.
+	MemberName() string
+
+	// ValueForOuter returns the value represented by this
+	// type member within the input outer value.
+	ValueForOuter(reflect.Value) (reflect.Value, error)
+
+	AddScanTarget(reflect.Value, []any, []ScanProxy) ([]any, []ScanProxy, error)
```
SimonRichardson: This requires documentation, along with all the public methods in this file.

internal/typeinfo/scan.go
```diff
@@ -0,0 +1,25 @@
+package typeinfo
+
+import "reflect"
+
+// ScanProxy is a shim for scanning query results
+// into types for which we have information.
+type ScanProxy struct {
```
SimonRichardson, line 7: This feels like it's doing too much. I'm not suggesting fixing it now, as you're doing a lot of mechanical changes, but we should have a `MapSetterProxy` and a `ValueSetterProxy`. Only `mapKey` returns the key, so that should return a `MapSetterProxy`, then `structField` can just return a `ValueSetterProxy`.  The `AddScanTarget` should take and return a `Proxy` interface.

```go
type Proxy interface {
    SetValue()
}
```

That way we remove the need for the if statement in `OnSuccess`, in addition to changing the name which doesn't tell you the intent of what it's actually doing.

manadart: I'm trying to limit this to a lift and shift. 

I actually started to do some other rework, and the change-set got unwieldy very fast. Once we have things a little better organised, I imagine we'll reconsider these implementations wholesale in the context of the features we need to add.

internal/expr/query.go
Aflynn50: I agree, this functionality is effectively the whole raison d'etre of `typeMember`. It makes sense it should be a method of it. For the name of this function, would something like `ValueFromOuter` be better? Because we are getting the member value from the outer value.

internal/expr/query.go
Aflynn50: This certainly works a natural method on `typeMember`. To me it seems a bit obscure passing the `ptrs` and the `scanProxies` to the `AddScanTarget` and doing the append inside the function. It makes sense to hide the details of how the `ScanTarget` is built from the code here, but building the list seems like a local concern, not one of `typeMember`. The two arguments `ptrs` and `scanProxies` are also only used with the `append`. Could this code be:

```
if ptr, scanProxy, err = typeMember.GetScanTarget(outputVal); err != nil {
	return nil, nil, err
}
ptrs = append(ptrs, ptr)
scanProxies = append(scanProxies, scanProxy)
```
I called it `GetScanTarget` above because `ScanTarget` sounds like we are scanning a target (the name `ScanArgs` suffers from a similar problem). Any other name that is unambiguous works.

internal/expr/query.go
```diff
@@ -161,53 +151,25 @@ func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []an
...
-			} else {
-				var val reflect.Value
-				if !sp.scan.IsNil() {
-					val = sp.scan.Elem()
-				} else {
-					val = reflect.Zero(sp.original.Type())
-				}
-				sp.original.Set(val)
-			}
+			sp.OnSuccess()
```
Aflynn50, line 119: Lovely

internal/typeinfo/member.go
Aflynn50: The documentation for this function might be a better place to document the behavior on NULL than where it is currently (in the `AddScanTarget` function of `structField`). The `AddScanTarget` method of `structField` would also work.

internal/typeinfo/scan.go
```diff
@@ -0,0 +1,25 @@
...
+	original reflect.Value
+	scan     reflect.Value
+	key      reflect.Value
+}
+
+func (sp ScanProxy) OnSuccess() {
```
Aflynn50, line 24: Documentation needed to make it clear this needs to be run after `rows.Scan`.

internal/typeinfo/scan.go
```diff
@@ -0,0 +1,25 @@
+package typeinfo
+
+import "reflect"
+
+// ScanProxy is a shim for scanning query results
+// into types for which we have information.
```
Aflynn50: The last part of the sentence is not very clear, maybe add that `ScanProxy` is used when scanning directly into the target is not possible/desired

internal/typeinfo/member.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (mk *mapKey) ValueForOuter(v reflect.Value) (reflect.Value, error) {
+	val := v.MapIndex(reflect.ValueOf(mk.name))
+	if val.Kind() == reflect.Invalid {
+		return val, fmt.Errorf("map %q does not contain key %q", mk.OuterType().Name(), mk.name)
+	}
+	return val, nil
+}
+
+func (mk *mapKey) AddScanTarget(outVal reflect.Value, ptrs []any, proxies []ScanProxy) ([]any, []ScanProxy, error) {
```
Aflynn50: It would be good to note the reason that we need the `ScanProxy` here, namely that Go does not expose pointers to map values (this would maybe be an internal comment rather than on the public function).

Aflynn50 (commented):
This feels like a very natural change, it makes the code a lot cleaner. It also great to see some of the small things tidied too like variable declarations up that have been bugging me ever since I started trying to be consistent.

internal/typeinfo/scan.go
letFunny: I think whether to dispatch or pattern match (see my general comment) comes down to personal opinion in the end, but I also think no paradigm should be absolute. As I said, my fear is that we do not take into account the original design and its quest for "simplicity".

In this concrete case we would be creating one file, one interface, three struct types and three methods to refactor a <20 line function. Is there a reason for that, are we envisioning that we will support more data types or that we need this added complexity?

letFunny: > That way we remove the need for the if statement in OnSuccess, in addition to changing the name which doesn't tell you the intent of what it's actually doing.

We would not be removing the if though, we would be using method dispatch do it instead, effectively we would only be moving the conditional elsewhere.

internal/expr/query.go
letFunny: I completely agree with @Aflynn50, I think the code is clearer if we process then accumulate instead of mutating the state inside each function. Conceptually it is also not the role of a typeMember to mutate the "global" state that is used in this outer function, we are imposing state on a method that is really stateless.

letFunny (commented):
In general I quite like the changes to the code, especially:
* The idea of creating another internal package for `typeinfo` if that is going to lead to better encapsulation especially by restricting the use of reflect only to that module.
* The creation of `typeMember.ValueForOuter(...)` which seems to be the correct abstraction here.

Apart from that, I feel we should have a conversation about the goals of the project and the paradigm that we want to follow. We briefly touched on that in an offline discussion and it comes down to method dispatching and encapsulation vs pattern matching. The project has been mostly following the latter by using interfaces as enums and by performing pattern matching using switch statements on those interfaces. We seem to be now moving towards method dispatching and encapsulation which is all right, it comes down to personal preference after all. My only fear is that we do not take into account the original design of SQLair and its shared vision.

Specifically, one of the main objectives of SQLair was to be simple and straightforward to read and I am not sure that continues to be a shared goal (see [this comment](https://github.com/canonical/sqlair/pull/121/files#r1395838194) or [this one](https://github.com/canonical/sqlair/pull/121/files#r1395715007)). For this reason, if we decide on changing the paradigm to encapsulation and indirection, I think it is important that we continue to balance it against our explicit goal of simplicity and explicitness.

Lastly, if we decide to sacrifice this simplicity for better modularization or any other goal, I will commit fully to the new values, but I think it is a conversation that we need to have.

Aflynn50 (approved):
LGTM

PR #122, opened 2023-11-16
Change names and add documentation for new pipeline

Aflynn50:
Change type/variable names and add documentation for the new SQLair pipeline. A `doc.go` file is added to the `expr` package to detail how the pipeline should look. The names of the types in the pipeline are changed to clarify they purpose.

The full planned changes are described in the `doc.go` file added here. The upcoming changes In short are: the SQL method will be on `PrimedQuery` rather than `TypedExpr` and the `TypedExpr` will be changed to retain the shape of the AST from the `ParsedExpr`.


internal/expr/doc.go
```diff
@@ -0,0 +1,40 @@
...
+# Parsing stage
+
+The parsing stage takes a SQLair query string and parses it into an AST. The
+parser only processes information already encoded in the syntax of the SQLair
+expressions.
+
+# Type Binding stage
+
+The Type Binding stage binds concrete Go types to the type names in the SQLair
+query expressions. Type samples are provided by the user. These are used to
```
letFunny: Should we drop the `query` from `query expressions`? I see above we are saying `SQLair expressions` which seems more natural.

internal/expr/doc.go
```diff
@@ -0,0 +1,40 @@
...
+validate the SQLair query expressions, and to generate an intermediate
```
letFunny: Ditto.

internal/expr/doc.go
```diff
@@ -0,0 +1,40 @@
+/*
```
letFunny, line 1: I love the idea of including the structure of the pipeline in the code itself. I was a bit reluctant but I can see how newcomers and contributors would find it very useful.

I guess the idea is to include the docs as part of this PR and update the implementation to match it in subsequent ones, is that right?

internal/expr/parser.go
```diff
@@ -6,15 +6,13 @@ import (
 	"strings"
 )
 
-// A QueryPart represents a section of a parsed SQL statement, which forms
-// a complete query when processed together with its surrounding parts, in
-// their correct order.
-type queryPart interface {
-	// String returns the part's representation for debugging purposes.
+// expr represents a section of a parsed SQL statement.
```
letFunny: I would prefer avoiding the usage of `statement` which is already very overloaded in SQLair. Additionally, what do you think of changing the description to state it is conceptually part of an AST, not just a mere section of the text in the query. Maybe something along the lines of:
```
// expr are the nodes of the query's AST
```
or
```
// expr are the constructs that constitute a parsed SQL query
```

internal/expr/parser.go
```diff
@@ -40,54 +38,54 @@ func (ca columnAccessor) String() string {
...
-	prevPart int
-	// partStart is the value of pos just before we started parsing the part
-	// under pos. We maintain partStart >= prevPart.
-	partStart int
-	parts     []queryPart
+	// prevExpr is the value of pos when we last finished parsing a expr.
+	prevExpr int
+	// exprStart is the value of pos just before we started parsing the expr
+	// under pos. We maintain exprStart >= prevExpr.
+	exprStart int
```
letFunny: [Nit]: Now that we are changing the names, either as part of this PR or another one can we change the names to `prevExprPos`, `exprStartPos`, or even `prevExprEnd` and `currentExprStart`? We can discuss them offline but something that is self-explanatory would be perfect.

internal/expr/parser.go
```diff
@@ -257,37 +249,36 @@ func (p *Parser) skipComment() bool {
 	return false
 }
 
-// Parse takes an input string and parses the input and output parts. It returns
-// a pointer to a ParsedExpr.
-func (p *Parser) Parse(input string) (expr *ParsedExpr, err error) {
+// Parse takes an SQLair query string and returns a ParsedExpr.
+func (p *Parser) Parse(query string) (pe *ParsedExpr, err error) {
```
letFunny: [Nit]: I am not sure we should rename the argument to `query` here. `query` means something different in various sites in the code base while `input` has a concrete meaning. And then in other PRs (function parsing) we are using the parser to parse things that are not queries which is why I think `input` is better.

internal/expr/query.go
```diff
@@ -87,19 +87,22 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
...
 var scannerInterface = reflect.TypeOf((*sql.Scanner)(nil)).Elem()
 
-// ScanArgs returns list of pointers to the struct fields that are listed in qe.outputs.
-// All the structs and maps mentioned in the query must be in outputArgs.
-func (qe *QueryExpr) ScanArgs(columns []string, outputArgs []any) (scanArgs []any, onSuccess func(), err error) {
+// ScanArgs takes the result column names and the SQLair output arguments. It
+// returns list of pointers to be used as arguments for rows.Scan. They point to
+// locations within the SQLair output arguments. The onSuccess function should
+// be run once the pointers have been scanned to populate output arguments that
+// cannot be written to directly (such as map values).
```
letFunny: I think this comment can be improved a little if we reword it to avoid saying "arguments" for two different things, and if we avoid talking about the implementation (i.e. all we care about is that `onSuccess` needs to be ran after rows.Scan, we do not care why in this comment).

What do you think of the following?
```
// ScanArgs produces a list of pointers to be passed to rows.Scan. After a
// successful call, the onSuccess function must be invoked. The end result is
// that the outputArgs will be populated with the results from the query.
// All the structs/maps/slices mentioned in the query must be in outputArgs.
```
plus changing `columns` to `columnNames`.

sqlair.go
```diff
@@ -50,7 +50,7 @@ var cacheMutex sync.RWMutex
 // It is ready to be run on a SQLair DB.
 type Statement struct {
 	cacheID stmtID
-	pe      *expr.PreparedExpr
+	te      *expr.TypedExpr
```
letFunny: What do you think about including a comments here saying we are caching the `TypeExpr` to be reused for all queries? It used to be that the `sqlair.Statement` had all the values necessary so it was self-explanatory but that is no longer that.

internal/expr/query.go
```diff
@@ -8,36 +8,36 @@ import (
...
+	return len(qv.outputs) > 0
 }
 
-type QueryExpr struct {
+// QueryValues contains all concrete values needed to run a SQLair query on a
+// database.
+type QueryValues struct {
 	sql     string
-	args    []any
+	params  []any
```
letFunny: We have probably talked about this already, but why is this named parameters if the types contained are named `NamedArg` implying arguments (we might have to change the name in the reference spec if we change it here)?

letFunny (commented):
This PR was very helpful for seeing the actual usage of the new names in the code, thanks! I already liked the new names but now seeing them in the actual code, I can see how well they fit.

I left a couple of suggestions mostly regarding small wording changes.

internal/expr/doc.go
Aflynn50: True, changed to just "SQLair expressions".

internal/expr/doc.go
Aflynn50: Ditto.

internal/expr/doc.go
Aflynn50: That's right, I've updated the PR comment to reflect this.

internal/expr/parser.go
Aflynn50: Changed to: `expr represents a parsed node of the SQLair query's AST.`

internal/expr/parser.go
Aflynn50: I like those suggested names, they are a lot clearer than the current names, I'll change them.

internal/expr/parser.go
Aflynn50: I think query makes sense in regard to this being a public method, we want to say what the argument is to the user, not to us. But since this is in the internal package it makes sense to change it back to input.

internal/expr/query.go
Aflynn50: Fair enough, the previous comment was a bit rambly, I've changed it to a more concise version of your suggestion.

sqlair.go
Aflynn50: I've added the comment:
```
// te is the type checked SQLair query. It is used to generate the required
// query values when the Statement is run on a database.
```

internal/expr/query.go
Aflynn50: In the spec we are using "Input arguments" to refer to the structs and maps from the user, and "query parameters" to refer to what is passed to the database. I've changed this name to try and disambiguate the two in the code.

I've added a comment here (and one for `outputs` too)

sqlair.go
letFunny: Make it `type bounded` to be consistent with the terminology. I like the comments but I think we can better convey that we are only caching the tree. We really don't need this in the statement, we have all the information in the .Query and .Get args as well. And `it is used to generate` seems to imply that we are using this alone to generate the values when in reality we have to wait for the .Query step.

Aflynn50: I've changed it from type checked to type bound.

I'm avoiding using the word caching here to talk about the information stored in the type bound expression because the other field in this struct is called `cacheID` and refers to something else. I wouldn't say that we are simply caching the tree here. The SQL string only appears in the call to `Prepare` and we have added the type information, some of which only appears at the `.Get` stage and is needed at the `.Query` stage.

I've changed the comment to:
```
te is the type bound SQLair query. It contains information used to
generate query values from the input arguments when the Statement is 
run on a database.
```

letFunny: I think the new comment you suggested is perfect. And I agree we should not talk about caching in two different contexts which might get muddled.

internal/expr/parser.go
```diff
@@ -6,15 +6,13 @@ import (
...
-type queryPart interface {
-	// String returns the part's representation for debugging purposes.
+// expression represents a parsed node of the SQLair query's AST.
+type expression interface {
+	// String returns a text representation for debugging and testing purposes.
 	String() string
 
 	// marker method
-	part()
+	expression()
```
manadart: Keep this one as `expr` to make it distinct from the type name. I hope we can obviate the meed for it in future anyway.

internal/expr/query.go
```diff
@@ -8,36 +8,38 @@ import (
...
-// substitute the input placeholders in the prepared expression. These placeholders use
-// the syntax "$Person.fullname", where Person would be a type such as:
-//
-//	type Person struct {
-//	        Name string `db:"fullname"`
-//	}
-func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
+// BindInputs takes the SQLair input arguments and returns the PrimedQuery ready
+// for use with the database.
+func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
```
manadart: Move this to be co-located with its type declaration in _typebind.go_

internal/expr/query.go
```diff
@@ -69,9 +71,9 @@ func (pe *PreparedExpr) Query(args ...any) (ce *QueryExpr, err error) {
 		}
 	}
 
-	// Query parameteres.
-	qargs := []any{}
-	for i, typeMember := range pe.inputs {
+	// Query parameters.
```
manadart: Not for this patch, but this is an example of what I mentioned in our meeting. If you are using comments to label the steps of a large method, just make the steps their own methods.

internal/expr/query.go
```diff
@@ -8,36 +8,38 @@ import (
 	"strings"
 )
 
-func (qe *QueryExpr) QueryArgs() []any {
-	return qe.args
+// PrimedQuery contains all concrete values needed to run a SQLair query on a
+// database.
+type PrimedQuery struct {
```
manadart, line 12: I had a thought that `PrimedQuery` is not such a great name due to being a member of `sqlair.Query`, but I think it's OK, because it is distinct by being from the `expr` package, from who's point-of-view the work is done.

manadart (commented):
Almost there.

internal/expr/parser.go
Aflynn50: Changed

internal/expr/query.go
Aflynn50: I'm not 100% happy with it either, but it can stay there for now, if we think of a better name we can change it

internal/expr/query.go
Aflynn50: Moved to `typebind.go`.

Following this pattern the `func (*ParsedExpr) BindTypes(...)` method in `typebind.go` should really be moved to the parser to be colocated with the declaration of `ParsedExpr`. To me though that move would feel a bit wrong, all the actual "type binding" takes place in that method so it should really be in `bindtypes.go`. I have left it as it is for now.

manadart (approved):
This is good to land and iterate on.

PR #124, opened 2023-11-24
Rework typeinfo to allow for slices

Aflynn50:
Change the interfaces in typeInfo to allow for slice type to be naturally supported (this PR does not itself add support for slices).

This PR adds `Input` and `Output` interfaces to the `typeinfo` package. These break up the existing `Member` into more natural separate parts. A new interface `ValueLocator` replaces the `Member` interface (since slice ranges are not "members" of anything).

The `typeinfo.Info` interface is replaced with a private `arg` interface. The `Info` type was previously used to generate the `Member` values. It has been replaced with a struct, `ArgInfo` which can be generated by `typeinfo.GenerateArgInfo(...)`. This `ArgInfo` acts as a black box containing type information which can be requested by the `expr` package via new methods on it.

The testing for `typetnfo` has been completely redone. The old tests have been removed and new, more complete, tests added. The error messages have not been changed in this PR where avoidable.


internal/typeinfo/member.go
```diff
@@ -8,59 +8,78 @@ import (
...
-	// MemberName returns this member's name.
-	MemberName() string
-
-	// ValueFromOuter returns the value represented by this
-	// type member within the input outer value.
-	ValueFromOuter(reflect.Value) (reflect.Value, error)
+// Input is a locator for a SQL query parameter from SQLair input arguments.
+type Input interface {
+	// GetParams returns the query parameters represented
+	// by this Input from the assosiated input argument.
```
letFunny:
```suggestion
	// LocateParams returns the query parameters represented
	// by this Input from the associated input argument.
```
typos

letFunny: Apart from that, I would refrain from using terms such as `query parameters` which we have already defined as the `NamedArg`s passed to the db. I would rephrase it like:
```
// LocateParams returns the Go values from the input arguments.
```

internal/typeinfo/member.go
```diff
@@ -8,59 +8,78 @@ import (
...
-	// OuterType returns the outer type in which this member is present.
-	OuterType() reflect.Type
-
-	// MemberName returns this member's name.
-	MemberName() string
-
-	// ValueFromOuter returns the value represented by this
-	// type member within the input outer value.
-	ValueFromOuter(reflect.Value) (reflect.Value, error)
+// Input is a locator for a SQL query parameter from SQLair input arguments.
```
letFunny: I think this is using `query parameter` incorrectly, we are locating Go values from SQLair input arguments.

internal/expr/bindtypes.go
```diff
@@ -44,69 +39,43 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
 	var inCount int
 	var outCount int
 
-	var outputs = make([]typeinfo.Member, 0)
-	var inputs = make([]typeinfo.Member, 0)
+	var outputs = make([]typeinfo.Output, 0)
+	var inputs = make([]typeinfo.Input, 0)
 
-	var typeMemberPresent = make(map[typeinfo.Member]bool)
+	var outputUsed = make(map[typeinfo.Output]bool)
```
letFunny, line 55:
```suggestion
	var outputsUsed = make(map[typeinfo.Output]bool)
```

internal/expr/bindtypes.go
```diff
@@ -174,50 +138,42 @@ func bindOutputTypes(ti typeNameToInfo, e *outputExpr) (outCols []columnAccessor
...
 
 		for _, t := range e.targetTypes {
-			info, ok := ti[t.typeName]
-			if !ok {
-				return nil, nil, typeMissingError(t.typeName, getKeys(ti))
-			}
 			if t.memberName == "*" {
 				// Generate asterisk columns.
-				allMembers, err := info.GetAllMembers()
+				members, memberNames, err := argInfo.AllOutputMembers(t.typeName)
```
letFunny: I was taking a look at the usage of these functions and I am not sure it is the correct abstraction. It is true that now the implementation lives closer to each class but it seems like a lot of boilerplate is needed. For instance, if we replaced the `AllOutputMembers` with
```go
	arg, err := argInfo.GetArg(t.typeName)
	if err != nil {
		return nil, nil, err
	}
	si, ok := arg.(*structInfo)
	if !ok {
		return nil, nil, fmt.Errorf("columns must be specified for map with star")
	}
	members, memberNames := si.allOutputMembers()
```
it would be clearer because:
1. We can see directly in the code that we need the type to be a struct which is not clear when looking at the method.
2. We remove the `AllOutputMembers` from the interface. I see this as a benefit because it is a method that only makes sense for one of the implementors while the other three will return an error.

internal/typeinfo/info.go
```diff
@@ -13,17 +13,127 @@ import (
...
-// Info exposes useful information about types used in SQLair queries.
-type Info interface {
-	Typ() reflect.Type
+// ArgInfo is used to access type information useful for SQLair. It should only
+// be accessed using it methods, not used directly as a map.
+type ArgInfo map[string]arg
+
+// GenerateArgInfo returns type information useful for SQLair by reflecting on
+// sample instantiations of argument types.
+func GenerateArgInfo(typeSamples ...any) (ArgInfo, error) {
```
letFunny: I think this abstraction is great! I can see that the data from the type samples is created without having to parse all lines that do that along with the error reporting.

internal/typeinfo/info.go
```diff
@@ -13,17 +13,127 @@ import (
...
 
-// Info exposes useful information about types used in SQLair queries.
-type Info interface {
-	Typ() reflect.Type
+// ArgInfo is used to access type information useful for SQLair. It should only
+// be accessed using it methods, not used directly as a map.
+type ArgInfo map[string]arg
+
+// GenerateArgInfo returns type information useful for SQLair by reflecting on
+// sample instantiations of argument types.
```
letFunny: I am not really sure how but I think we should reword this to make it clearer what is doing. Can we change `useful` with something like `type information (e.g. whether the type is a struct or a map, the fields it has, etc.)`

internal/typeinfo/info.go
```diff
@@ -13,17 +13,127 @@ import (
...
+	}
+	var vl ValueLocator
+	switch arg := arg.(type) {
+	case *structInfo:
+		vl, err = arg.member(member)
+	case *mapInfo:
+		vl, err = arg.member(member)
+	default:
+		return nil, fmt.Errorf("internal error: unknown arg type %T", arg)
+	}
```
letFunny: This feels wrong in a way. We can discuss it offline but maybe we can create another interface for the `member()` function and check whether the argument qualifies.

internal/typeinfo/info.go
```diff
@@ -13,17 +13,127 @@ import (
...
+		case reflect.Pointer:
+			return nil, fmt.Errorf("need struct or map, got pointer to %s", t.Elem().Kind())
+		default:
+			return nil, fmt.Errorf("need struct or map, got %s", t.Kind())
+		}
+	}
+	return argInfo, nil
+}
+
+// InputMember returns an input locator for a member of a struct or map.
```
letFunny: It feels wrong that we have a function that only works for some of the types. Normally the flow is:
1. Check what is the type of the object I have.
2. Call the relevant function.

and not,
1. Call the function on the interface which only works for some of the implementors.
2. See if it returns an error.

I also have to admit that it is difficult to come up with a difficult abstraction. I was think of having an interface for types that have the `member` function. We can then change the signature of this function to expect that interface. In the code we would do:
```go
myType, ok := myType.(HasMembers)
if !ok {
     fmt.Errorf("type cannot be indexed by member")
}
argInfo.InputMember(myType, memberName)
```
but at this point isn't it easier to check whether the type is a struct/map directly for better error messages and more locality? (Incidentally that is what we are doing in `getMember`, we are just hiding it away.

letFunny (commented):
I did a very light first pass on some of the files. I think it would be very valuable to dedicate some time to discuss the abstraction face to face. I think in general the code is moving in a good direction, but the whole abstraction about members is challenging and it can probably do with some more discussion.

internal/expr/bindtypes.go
Aflynn50: I actually prefer it as `outputUsed`, its a map rather than a slice so when you look up an output you do `if outputUsed[o]`, or `outputUsed[o] = true` which makes more sense than `if outputsUsed[o]` since you are only checking one.

letFunny: That makes sense as well, I agree with you, let's keep the name as is.

internal/expr/bindtypes.go
Aflynn50: The `AllOutputMembers` method is on the `ArgInfo` struct, not actually on any interface. I think it does need a name change to something like `AllStructOutputs` because it does only work for structs, but the caller does not actually know that they are getting a struct back when they pass it the string with the name. In my opinion the method makes sense where it is but could do with a name change.

The error should also be changed, I didn't change this because  I didn't want to change the errors much in this PR but this one doesn't really make sense anymore.

For the code above, I don't think we should go back to exposing the information for individual info types to the `expr` package, but I think it should be explicit that a struct is expected.

internal/typeinfo/info.go
Aflynn50: I've redone the comments on `GenerateArgInfo` and `ArgInfo` to make them more specific

internal/typeinfo/info.go
Aflynn50: The argument to InputMember is a string of the type name and the member of it. All the caller knows at that point is that they have the name of a type, and the name of a supposed member of it. I think it makes sense that the caller should expect an error if the type name passed is not a type that actually has members.

internal/typeinfo/info.go
Aflynn50: Agreed, there are two simple solutions I can think of:
1. Add a private `hasMember` interface which structInfo and mapInfo implement
2. Inline the member functions

I have changed it to option 1 for now, but I think it could be a bit overkill.

In fact, it could be a good idea to get rid of other methods on the `mapInfo` and `structInfo` all together and just inline them in the functions on `ArgInfo`. That would make things a bit simpler here.

internal/typeinfo/member.go
Aflynn50: Fixed, looks like I forgot to update that comment

internal/expr/bindtypes.go
```diff
@@ -44,69 +39,43 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
-				}
-				return nil, fmt.Errorf("two types found with name %q: %q and %q", t.Name(), dupeInfo.Typ().String(), t.String())
-			}
-			ti[t.Name()] = info
-		case reflect.Pointer:
-			return nil, fmt.Errorf("need struct or map, got pointer to %s", t.Elem().Kind())
-		default:
-			return nil, fmt.Errorf("need struct or map, got %s", t.Kind())
-		}
+	argInfo, err := typeinfo.GenerateArgInfo(args...)
```
manadart: Better.

internal/typeinfo/info.go
```diff
@@ -13,17 +13,131 @@ import (
...
+// arguments. Methods on ArgInfo can be used to generate input and output value
+// locators.
+//
+// ArgInfo should only be accessed using it methods, not used directly as a
+// map.
+type ArgInfo map[string]arg
+
+// GenerateArgInfo takes sample instantiations of argument types and uses
+// reflection to generate an ArgInfo containing the types.
+func GenerateArgInfo(typeSamples ...any) (ArgInfo, error) {
```
manadart: Variadic arg lists are ergonomic on an exported API, but once inside internal/private packages just make them slices. Then you don't have to keep splatting through the call stack.

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
+	// LocateScanTarget locates the output argument associated this Output in
+	// typeToValue and returns a pointer to the Go value within the output
+	// argument for rows.Scan, along with a ScanProxy for the cases where the
+	// output argument cannot be scanned into directly.
+	LocateScanTarget(typeToValue map[reflect.Type]reflect.Value) (any, *ScanProxy, error)
+	ValueLocator
+}
+
+// ValueLocator specifies how to locate a value in a SQLair argument type.
+type ValueLocator interface {
```
manadart: Declare types ahead of usage; funcs/methods in hierarchy of call.

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
-	GetScanTarget(reflect.Value) (any, *ScanProxy, error)
+// Input is a locator for a Go value from SQLair input arguments to be used in
+// a SQL query parameter.
+type Input interface {
+	// LocateParams locates the input argument associated with this Input in
+	// the typeToValue map and then returns the Go values within the input
+	// argument that are to be used in a query parameter. An error is returned
+	// if typeToValue does not contain the input argument.
+	LocateParams(typeToValue map[reflect.Type]reflect.Value) ([]reflect.Value, error)
+	ValueLocator
```
manadart: Embeddings first (general -> specific).

internal/expr/bindtypes.go
```diff
@@ -131,29 +100,24 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
 
 // bindInputTypes binds the input expression to a type and returns the
 // typeMember represented by the expression.
-func bindInputTypes(ti typeNameToInfo, e *inputExpr) (tm typeinfo.Member, err error) {
+func bindInputTypes(argInfo typeinfo.ArgInfo, e *inputExpr) (input typeinfo.Input, err error) {
```
manadart, line 103: If you reorder these arg lists, putting subject ahead of auxiliary data, you can start to see their candidacy for putting on the input/output expression types themselves.

Not advocating for this right now.

internal/expr/bindinputs.go
```diff
@@ -33,12 +33,11 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
 	var inQuery = make(map[reflect.Type]bool)
-	for _, typeMember := range te.inputs {
-		inQuery[typeMember.OuterType()] = true
+	for _, input := range te.inputs {
+		inQuery[input.ArgType()] = true
 	}
 
-	var typeValue = make(map[reflect.Type]reflect.Value)
-	var typeNames []string
+	var typeToValue = make(map[reflect.Type]reflect.Value)
```
manadart: `typeToValue := make...`

internal/expr/bindinputs.go
```diff
@@ -49,37 +48,33 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
-	var params []any
-	for i, typeMember := range te.inputs {
-		outerType := typeMember.OuterType()
-		v, ok := typeValue[outerType]
-		if !ok {
-			return nil, typeMissingError(outerType.Name(), typeNames)
-		}
-
-		val, err := typeMember.ValueFromOuter(v)
+	params := []any{}
```
manadart: `var params []any`

internal/expr/bindtypes.go
```diff
@@ -226,24 +182,21 @@ func bindOutputTypes(ti typeNameToInfo, e *outputExpr) (outCols []columnAccessor
...
 		}
 	} else {
 		return nil, nil, fmt.Errorf("mismatched number of columns and target types")
 	}
 
-	return outCols, typeMembers, nil
+	return outCols, outputs, nil
 }
 
+/*
```
manadart: Just delete if not used.

manadart (approved):
I don't think we have the perfect abstractions yet, but this is a step in the right direction.

I'm happy to land and iterate if @letFunny has his concerns accommodated enough to approve.

internal/expr/bindtypes.go
Aflynn50: Agreed, I make this an interface method on `expression` in #112

internal/expr/bindtypes.go
Aflynn50: Ah yea, was intending to delete this

internal/expr/bindinputs.go
```diff
@@ -67,19 +65,16 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
-		val, err := typeMember.ValueFromOuter(v)
+	inputCount := 0
+	for _, input := range te.inputs {
+		vals, err := input.LocateParams(typeToValue)
 		if err != nil {
 			return nil, err
 		}
-
-		params = append(params, sql.Named("sqlair_"+strconv.Itoa(i), val.Interface()))
+		for _, val := range vals {
```
letFunny, line 74: I like that we are already paving the way for slices!

internal/expr/expr_test.go
```diff
@@ -570,17 +570,17 @@ func (s *ExprSuite) TestMapError(c *C) {
 		"all output into map star",
 		"SELECT &M.* FROM person WHERE name = 'Fred'",
 		[]any{sqlair.M{}},
-		"cannot prepare statement: output expression: columns must be specified for map with star: &M.*",
+		"cannot prepare statement: output expression: cannot generate columns for non-struct type: &M.*",
```
letFunny: I preferred the error message that we had in the past, it clearly says how to fix they issue (i.e. by specifying the columns). I would keep the new `non-struct type`  though.

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
+// messages.
+func (mk *mapKey) String() string {
+	return "key \"" + mk.name + "\" of map \"" + mk.mapType.Name() + "\""
 }
 
-// GetScanTarget returns a pointer for the target of rows.Scan, and a ScanProxy
-// reference for setting that target as the value for this map key.
-func (mk *mapKey) GetScanTarget(outVal reflect.Value) (any, *ScanProxy, error) {
+// LocateScanTarget locates the map specified in mapKey from the provided
+// typeToValue map. It returns a pointer for to pass to rows.Scan, and a
```
letFunny:
```suggestion
// typeToValue map. It returns a pointer to pass to rows.Scan, and a
```

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
+// ArgType returns the type of the map the key is located in.
+func (mk *mapKey) ArgType() reflect.Type {
 	return mk.mapType
 }
 
-// MemberName returns the map key.
-func (mk *mapKey) MemberName() string {
-	return mk.name
+// LocateParams locates the map and then the value of the key specified in
+// mapKey from the provided typeToValue map. An error is returned if the map
```
letFunny: What do you mean with `value of the key [...] from the provided typeToValue map`? Should it be `locates the map using typeToValue and then uses gets the value associated with the key ?

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
+type Output interface {
+	ValueLocator
+	// LocateScanTarget locates the output argument associated this Output in
+	// typeToValue and returns a pointer to the Go value within the output
+	// argument for rows.Scan, along with a ScanProxy for the cases where the
+	// output argument cannot be scanned into directly.
+	LocateScanTarget(typeToValue map[reflect.Type]reflect.Value) (any, *ScanProxy, error)
+}
+
+// mapKey stores information about where to find a key of a particular map.
```
letFunny:
```
// mapKey stores information about where to find a value in a map.
```

internal/typeinfo/info.go
```diff
@@ -9,110 +9,182 @@ import (
...
+		argNames := []string{}
+		for argName := range argInfo {
+			argNames = append(argNames, argName)
+		}
+		// Sort for consistant error messages.
+		sort.Strings(argNames)
+		return nil, typeMissingError(typeName, argNames)
+	}
+	return arg, nil
 }
```
letFunny: [Nit] What is your opinion on refactoring the error to a separate function (I know we already have `typeMissingError`), to change the code in each function to:
```go
        arg, ok := argInfo[typeName]
	if !ok {
		newFunctionForError(typeName, mapToList(argInfo))
	}
```
instead of:
```go
	arg, err := argInfo.getArg(typeName)
	if err != nil {
		return nil, nil, err
	}
```

internal/typeinfo/info.go
```diff
@@ -9,110 +9,182 @@ import (
...
 	}
-	return tm, nil
+	return output, nil
 }
 
-func (si *structInfo) GetAllMembers() ([]Member, error) {
+// AllStructOutputs returns a list of output locators that locate every member
+// of the named type along with the names of the members. If the type is not a
+// struct an error is returned.
+func (argInfo ArgInfo) AllStructOutputs(typeName string) ([]Output, []string, error) {
```
letFunny: I love the new name, that addresses all of my concerns in a better way than what I proposed.

internal/typeinfo/member.go
```diff
@@ -116,3 +150,17 @@ func (f *structField) GetScanTarget(outVal reflect.Value) (any, *ScanProxy, erro
 	}
 	return val.Addr().Interface(), nil, nil
 }
+
+// locateValue locates the value corresponding to the given type in the
+// typeToValueMap and returns an error message if it cannot be found.
+func locateValue(typeToValue map[reflect.Type]reflect.Value, typ reflect.Type) (reflect.Value, error) {
```
letFunny, line 158: Same comment, if we could change this function to report the error and we can have the map access directly in the `LocateScanTarget` functions, that would make it simpler and easier to read.

internal/typeinfo/member.go
```diff
@@ -8,59 +8,82 @@ import (
...
+	String() string
 }
 
+// Input is a locator for a Go value from SQLair input arguments to be used in
+// a SQL query parameter.
+type Input interface {
+	ValueLocator
+	// LocateParams locates the input argument associated with this Input in
+	// the typeToValue map and then returns the Go values within the input
+	// argument that are to be used in a query parameter. An error is returned
```
letFunny:
```suggestion
	// argument that are to be used in query parameters. An error is returned
```

internal/typeinfo/member.go
```diff
@@ -79,32 +102,43 @@ type structField struct {
...
-// Rows.Scan will return an error if it tries to scan NULL into a type that
+// LocateScanTarget locates the struct specified in structField from the
+// provided typeToValue map. It returns a pointer for the target of rows.Scan,
+// and a ScanProxy reference in the event that we need to coerce that pointer
+// into a struct field.
+// rows.Scan will return an error if it tries to scan NULL into a type that
 // cannot be set to nil, so for types that are not a pointer and do not
 // implement sql.Scanner, a pointer to them is generated and passed to
 // Rows.Scan. If Scan has set this pointer to nil the value is zeroed by
 // ScanProxy.OnSuccess.
```
letFunny: I think this comment is really necessary when you read the code. I am wondering if we should move it to the `LocateScanTarget` in the interface itself instead. What do you think?

internal/typeinfo/member.go
```diff
@@ -79,32 +102,43 @@ type structField struct {
 	omitEmpty bool
 }
 
-// OuterType returns the reflected type of struct in
-// which this Member implementation is a field.
-func (f *structField) OuterType() reflect.Type {
+// ArgType returns the type of struct in this field is located in.
```
letFunny:
```suggestion
// ArgType returns the type of the struct this field is located in.

File: internal/typeinfo/member.go [hunk 28]
```diff
@@ -79,32 +102,43 @@ type structField struct {
...
+func (f *structField) ArgType() reflect.Type {
 	return f.structType
 }
 
-// MemberName returns the name of this struct field.
-func (f *structField) MemberName() string {
-	return f.tag
+// LocateParams locates the struct the field is located in from the typeToValue
```
letFunny: [Nit]: Maybe `the struct that contains the field`?

letFunny (commented):
Alastair and me synced offline and he explained how he planned to make slices fit into the new abstraction which was my concern in the previous review. He also suggested several simplifications which addressed all of my concerns. Thus, approving the latest changes (I have to review the tests and then I will formally approve).

I left some nitpicks about names and code comments.

internal/typeinfo/member_test.go
```diff
@@ -0,0 +1,153 @@
...
+	output, err := argInfo.OutputMember("M", "foo")
+	c.Assert(err, IsNil)
+
+	m := M{}
+	valOfM := reflect.ValueOf(m)
+	typeToValue := map[reflect.Type]reflect.Value{
+		reflect.TypeOf(m): valOfM,
+	}
+	// Values in maps cannot be set directly. A proxy is set by rows.Scan then
+	// we set it withe OnSuccess function in our map.
```
letFunny:
```suggestion
	// we set it with the OnSuccess function in our map.
```

internal/typeinfo/member_test.go
```diff
@@ -0,0 +1,153 @@
...
+	output, err := argInfo.OutputMember("T", "foo")
+	c.Assert(err, IsNil)
+
+	ptr, scanProxy, err := output.LocateScanTarget(typeToValue)
+	c.Assert(err, IsNil)
+
+	// Check scanProxy has the expected values.
+	c.Assert(scanProxy.original, Equals, valOfT.FieldByName("Foo"))
+	c.Assert(scanProxy.key, Equals, reflect.Value{})
+	c.Assert(scanProxy.scan, FitsTypeOf, reflect.ValueOf((*int)(nil)))
```
letFunny: What are you checking here? The test still pass if you do `c.Assert(scanProxy.scan, FitsTypeOf, reflect.Value{})`.

Go does not have parametrized types so not sure if you can use `FitsTypeOf` to check that the "inner" value of reflect.Value is of type *int.

internal/typeinfo/member_test.go
```diff
@@ -0,0 +1,153 @@
...
+	typeToValue := map[reflect.Type]reflect.Value{
+		reflect.TypeOf(m): valOfM,
+	}
+
+	input, err := argInfo.InputMember("M", "baz")
+	c.Assert(err, IsNil)
+
+	_, err = input.LocateParams(typeToValue)
+	c.Assert(err, ErrorMatches, `map "M" does not contain key "baz"`)
+}
```
letFunny, line 198: This may be redundant but can we add a test for a type that does not exist just to see the error message and run the code path?

internal/typeinfo/member.go
Aflynn50: Clarified

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,227 @@
...
+	}}
+
+	allOutputs, names, err := argInfo.AllStructOutputs("myStruct")
+	c.Assert(err, IsNil)
+	c.Assert(allOutputs, HasLen, len(structFields))
+
+	for i, t := range structFields {
+		expectedStructField := &structField{
+			name:       t.fieldName,
+			structType: reflect.TypeOf(myStruct{}),
```
letFunny: [Nit]: You can define the type above instead of calling reflect for each subtest.

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,227 @@
...
+	c.Assert(err, ErrorMatches, "need struct or map, got string")
+
+	_, err = GenerateArgInfo([]any{0})
+	c.Assert(err, ErrorMatches, "need struct or map, got int")
+
+	_, err = GenerateArgInfo([]any{[10]int{}})
+	c.Assert(err, ErrorMatches, "need struct or map, got array")
+
+	_, err = GenerateArgInfo([]any{t, T{}})
+	c.Assert(err, ErrorMatches, `two types found with name "T": "typeinfo.T" and "typeinfo.T"`)
```
letFunny: Can we refactor this to be a table and run the test against each test case in the table?

internal/typeinfo/typeinfo_test.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (s *typeInfoSuite) TestGenerateArgInfoStructError(c *C) {
```
letFunny: Same, can we make it a table?

package_test.go
```diff
@@ -339,7 +339,7 @@ func (s *PackageSuite) TestIterGetErrors(c *C) {
 		types:   []any{Person{}},
 		inputs:  []any{},
 		outputs: []any{},
-		err:     `cannot get result: type "Person" found in query but not passed to get`,
+		err:     `cannot get result: parameter with type "Person" missing`,
```
letFunny, line 342: Love it! Nice and simple.

letFunny (commented):
A couple of small comments in the tests :)

internal/typeinfo/info.go
Aflynn50: I think it could be nice, except that we have to do this code in the other place we call this function with a tiny difference, and we would probably also want to create a function there that's very similar but slightly different (the map keys are strings here but they are `reflect.Types` in the other callsite so we need to do `.Name()` there). Annoyingly generics can't really help here (we're not calling `String`, but `Name` which makes it awkward)

I've put a comment in both places this code appears to reduce the mental overhead of reading it but I think its fine how it is.

letFunny (approved):
Forgot to approve.

internal/typeinfo/member.go
Aflynn50: Changed.

internal/typeinfo/member.go
Aflynn50: Yea, makes sense, I've moved it up.

internal/typeinfo/member_test.go
Aflynn50: I'm just following the DRY principles ya know :upside_down_face:

internal/typeinfo/member_test.go
Aflynn50: Good point, changed it to compare actual values here with a `.Interface()` on the scan. It should also be `string` not `int`.

internal/typeinfo/member_test.go
Aflynn50: Added.

internal/typeinfo/typeinfo_test.go
Aflynn50: Changed, but its just a test

internal/typeinfo/typeinfo_test.go
Aflynn50: Changed.

internal/typeinfo/typeinfo_test.go
Aflynn50: I think with this one its better how it is, since you can't do type declarations in a table.

PR #126, opened 2023-11-30
Seperate cache operations from SQLair logic

Aflynn50:
Move the cache to a separate file and move operations on it to functions. The cache remains in the `sqlair` rather than being moved to a separate package because it needs access to the types in `sqlair` to set finalizers on new instantiations of them.

The singleton pattern is used in `newStmtCache` to allow only a single copy of the cache to be created.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -1,0 +1,147 @@
+package sqlair
+
+import (
+	"context"
+	"database/sql"
+	"runtime"
+	"sync"
+	"sync/atomic"
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// A SQLair Statement is prepared on a database when a Query method is run on a
+// DB/TX. The prepared statement is then stored in the stmtDBCache and a flag
+// is set in dbStmtCache.
+// A finalizer function is set on the Statement when it is placed in the cache.
+// On garbage collection, the finalizer cycles through the open databases in
+// the cache and closes each matching sql.Stmt. The finalizer then removes the
+// stmtID from stmtDBCache and dbStmtCache.
+// Similarly, a finalizer is set on the SQLair DB which closes all statements
+// prepared on the DB and then the sql.DB itself. It removes the dbID from
+// dbStmtCache and stmtDBCache.
+var stmtIDCount int64
+var dbIDCount int64
+
+type dbID = int64
+type stmtID = int64
+
+// statementCache caches the sql.Stmt values assosiated with each
+// sqlair.Statement. sqlair.Statement values can correspond to multiple
+// sql.Stmt values on different databases. The cache is indexed by the
+// sqlair.Statement ID and the sqlair.DB ID.
+//
+// The mutex must be locked when accessing either the stmtDBCache or the
+// dbStmtCache.
+type statementCache struct {
+	stmtDBCache map[stmtID]map[dbID]*sql.Stmt
+	dbStmtCache map[dbID]map[stmtID]bool
+	mutex       sync.RWMutex
+}
+
+func newStatementCache() *statementCache {
+	//Once something
+	return &statementCache{
+		stmtDBCache: map[stmtID]map[dbID]*sql.Stmt{},
+		dbStmtCache: map[dbID]map[stmtID]bool{},
+	}
+}
+
+// newStatement returns a new sqlair.Statement and allocates it in the cache. A
+// finalizer is set on the sqlair.Statement to remove all sql.Stmt values
+// assosiated with it from the cache and then run Close on the sql.Stmt values.
+// The finalizer is run after the sqlair.Statment is garbage collected.
+func (sc *statementCache) newStatement(te *expr.TypedExpr) *Statement {
+	cacheID := atomic.AddInt64(&stmtIDCount, 1)
+	s := &Statement{te: te, cacheID: cacheID}
+	sc.mutex.Lock()
+	sc.stmtDBCache[cacheID] = map[dbID]*sql.Stmt{}
+	sc.mutex.Unlock()
+	runtime.SetFinalizer(s, sc.getStmtFinalizer(s))
+	return s
+}
+
+// newDB returns a new sqlair.DB and allocates it in the cache. A finalizer is
+// set on the sqlair.DB which removes it from the cache, closes all sql.Stmt
+// values prepared upon it and then closes the DB. The finalizer is run after
+// the sqlair.DB is garbage collected.
+func (sc *statementCache) newDB(sqldb *sql.DB) *DB {
+	cacheID := atomic.AddInt64(&dbIDCount, 1)
+	sc.mutex.Lock()
+	sc.dbStmtCache[cacheID] = map[stmtID]bool{}
+	sc.mutex.Unlock()
+	db := &DB{sqldb: sqldb, cacheID: cacheID}
+	runtime.SetFinalizer(db, sc.getDBFinalizer(db))
+	return db
+}
+
+// prepareSubstrate is an object that queries can be prepared on, e.g. a sql.DB
+// or sql.Conn. It is used in prepareStmt.
+type prepareSubstrate interface {
+	PrepareContext(context.Context, string) (*sql.Stmt, error)
+}
+
+// prepareStmt prepares a Statement on a prepareSubstrate. It first checks in
+// the cache to see if it has already been prepared on the DB.
+// The prepareSubstrate must be assosiated with the same DB that prepareStmt is
+// a method of.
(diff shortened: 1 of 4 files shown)


cache.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// stmtIDCount and dbIDCount are global variables to used to generate unique
+// cache IDs.
+var stmtIDCount int64
+var dbIDCount int64
+
+type dbID = int64
```
manadart: I wouldn't bother creating types for these.

cache.go
```diff
@@ -0,0 +1,155 @@
...
+	"runtime"
+	"sync"
+	"sync/atomic"
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// stmtIDCount and dbIDCount are global variables to used to generate unique
+// cache IDs.
+var stmtIDCount int64
```
manadart: Since you have used a singleton pattern, the singleton being global, these should be inside the cache.
I would also use `uint64` and double the theoretical range.

manadart (commented):
Looks good, but which should we land first - #117, or this?

cache.go
Aflynn50: Removed.

cache.go
Aflynn50: Agreed, changed

Aflynn50:
> Looks good, but which should we land first - #117, or this?

Merging this first is fine

cache.go
```diff
@@ -0,0 +1,152 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// statementCache caches the sql.Stmt objects associated with each
+// sqlair.Statement. A sqlair.Statement can correspond to multiple sql.Stmt
+// objects prepared on different databases. Entries in the cache are therefore
+// indexed by the sqlair.Statement cache ID and the sqlair.DB cache ID.
```
letFunny, line 15: Superb!

cache.go
```diff
@@ -0,0 +1,152 @@
...
+// A finalizer is set on sqlair.Statement objects to close the assosiated
+// sql.Stmt objects. Similarly a finalizer is set on sqlair.DB objects to close
+// all sql.Stmt objects prepared on the DB, close the DB, and remove the DB
+// cache ID from the cache.
+//
+// The mutex must be locked when accessing either the stmtDBCache or the
+// dbStmtCache.
+type statementCache struct {
+	// stmtDBCache stores sql.Stmt objects addressed via the cache ID of the
+	// sqlair.Statement they built from and the sqlair.DB they are prepared on.
```
letFunny, line 26:
```
// stmtDBCache stores sql.Stmt objects addressed via the cache ID of their
// parent sqlair.Statement and the sqlair.DB they are prepared on.

File: cache.go [hunk 5]
```diff
@@ -0,0 +1,152 @@
...
+	stmtDBCache map[uint64]map[uint64]*sql.Stmt
+
+	// dbStmtCache indicates when a sqlair.Statement has been prepared on a particular sqlair.DB.
+	dbStmtCache map[uint64]map[uint64]bool
+
+	// stmtIDCount and dbIDCount used to generate unique cache IDs.
```
letFunny:
```
// stmtIDCount and dbIDCount are monotonically increasing numbers used to generate unique cache IDs.
```

cache.go
```diff
@@ -0,0 +1,152 @@
...
+// prepareStmt prepares a Statement on a prepareSubstrate. It first checks in
+// the cache to see if it has already been prepared on the DB.
+// The prepareSubstrate must be associated with the same DB that prepareStmt is
+// a method of.
+func (sc *statementCache) prepareStmt(ctx context.Context, dbID uint64, ps prepareSubstrate, s *Statement) (*sql.Stmt, error) {
+	var err error
+	sc.mutex.RLock()
+	// The Statement cache ID is only removed from stmtDBCache when the
+	// finalizer is run. The cache ID must be present as we have a reference to
+	// the Statement.
```
letFunny: I dont understand this sentence, the bit about the reference to the Statement.

cache.go
```diff
@@ -0,0 +1,152 @@
...
+			sc.dbStmtCache[dbID][s.cacheID] = true
+		}
+		sc.mutex.Unlock()
+	}
+	return sqlstmt, nil
+}
+
+// getStmtFinalizer returns a finalizer that removes a Statement from the
+// statement caches and closes it.
+func (sc *statementCache) getStmtFinalizer(s *Statement) func(*Statement) {
```
letFunny: Can we rename this function? I think that the name does not add anything now because it is talking about its usage and, said usage already tells us this is a finalizer:
```go
runtime.SetFinalizer(s, sc.getStmtFinalizer(s))
                        ----------------------
                        we already know this is a finalizer based on the function signature
```
Can we instead rename it to be `cleanCacheAndCloseStmtThunk`, if it is too verbose then `closeStmtInCacheThunk`, whatever conveys what is does instead of what it is used for.

sqlair.go
```diff
@@ -26,48 +24,18 @@ type M map[string]any
 var ErrNoRows = sql.ErrNoRows
 var ErrTXDone = sql.ErrTxDone
 
-var stmtIDCount int64
-var dbIDCount int64
```
letFunny, line 30: The benefit of keeping this is that it is much easier to read the map definitions. I would advocate for keeping it if @manadart agrees.
```go
map[uint64]map[uint64]*sql.Stmt
```
vs
```go
map[stmtID]map[dbID]*sql.Stmt
```

sqlair.go
```diff
@@ -167,7 +109,7 @@ func (db *DB) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
 		ctx = context.Background()
 	}
 
-	sqlstmt, err := db.prepareStmt(ctx, db.sqldb, s)
+	sqlstmt, err := stmtCache.prepareStmt(ctx, db.cacheID, db.sqldb, s)
```
letFunny: I find a little bit weird that `prepareStmt` is now a method of the cache. Maybe we can call it `getOrPrepare` which is more explicit about using the cache or falling back to regular prepare.

letFunny (commented):
It will take a while to see how all the new pieces fit together but I think this is a step in the good direction, it is adding more small self-contained functions so the code in the main file feels lighter and more readable.

I just left a few comments, mostly regarding naming.

cache.go
Aflynn50: I've changed this comment to clarify the meaning.

cache.go
Aflynn50: Fair, I've renamed these functions. I changed it to "Func" instead of "Thunk" because I'm not sure it'd be entirely clear to readers of the code what Thunk meant in this context.

sqlair.go
Aflynn50: Haha, he suggested I remove these. They do make the type signature clearer but I don't think its too big a deal. The types always appear next to a variable name that makes this clear.

sqlair.go
Aflynn50: Agreed, I've had a look at this and refactored it to respect the abstractions better. This new way removes the need for the whole PrepareSubstrate thing which I think obfuscated the code a little. Its now more clear that the transaction prepares on a connection, not on the database since all the non-cache related logic is happening in the `Query` function.

sqlair.go
```diff
@@ -467,9 +472,14 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
-	if err != nil {
-		return &Query{ctx: ctx, err: err}
+	sqlstmt, ok := stmtCache.lookupStmt(tx.db, s)
+	var err error
+	if !ok {
+		sqlstmt, err = tx.sqlconn.PrepareContext(ctx, s.te.SQL())
+		if err != nil {
+			return &Query{ctx: ctx, err: err}
+		}
+		stmtCache.storeStmt(tx.db, s, sqlstmt)
```
letFunny, line 482: I like this new abstraction a lot, it is now clear you are looking at the cache then falling back to preparing and then storing it in the cache.

Incidentally the problem with the connections is clearer now, you can see that we are preparing on `sqlconn` and then we are storing the statement based on `tx.db`.

letFunny (approved):
Looking great

cache.go
letFunny: Well thunk seems to have a more defined meaning (https://en.wikipedia.org/wiki/Thunk) but I like `Func` as well.

cache.go
letFunny: It is clear now, thanks

manadart (approved):
I want to do some empirical analysis of performance across some variables, which we will discuss, but this is a good mechanical change to move forward with.

PR #127, opened 2023-12-06
Redistribute Type Bind responsibility 

Aflynn50:
Move the responsibility of generating SQL to the Bind Inputs stage. This will allow SQL to support IN expressions where the SQL that is generated depends on the input arguments.

- Change `TypedBoundExpr` from containing lists of inputs outputs and the SQL to containing a typed versions of the expressions. This means it retains the shape of `ParsedExpr`.
- Rename `TypedExpr` to `TypeBoundExpr`. This removes the confusion with the name of the its constituents: `typedExpression`.
- Move the SQL generation to the bind inputs stage. It is now generated it alongside the query parameters.
- Change `ExprPkgTest` in `expr_test.go` to test the Bind Inputs stage as well.

### DIFF
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -76,8 +76,6 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 			}
 		case *bypass:
 			sqlStr.WriteString(te.chunk)
-		default:
-			return nil, fmt.Errorf("internal error: unknown type %T", te)
 		}
 	}
 
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -1,6 +1,7 @@
 package expr
 
 import (
+	"bytes"
 	"database/sql"
 	"fmt"
 	"reflect"
@@ -10,32 +11,19 @@ import (
 	"github.com/canonical/sqlair/internal/typeinfo"
 )
 
-// TypedExpr represents a SQLair query bound to concrete Go types. It contains
-// all the type information needed by SQLair.
-type TypedExpr struct {
-	outputs []typeinfo.Output
-	inputs  []typeinfo.Input
-	sql     string
-}
-
-// SQL returns the SQL ready for execution.
-func (te *TypedExpr) SQL() string {
-	return te.sql
-}
+// TypeBoundExpr represents a SQLair statement bound to concrete Go types. It
+// contains information used to generate the underlying SQL query and map it to
+// the SQLair query.
+type TypeBoundExpr []typedExpression
 
 // BindInputs takes the SQLair input arguments and returns the PrimedQuery ready
 // for use with the database.
-func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
+func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("invalid input parameter: %s", err)
 		}
 	}()
-
-	inQuery := map[reflect.Type]bool{}
-	for _, input := range te.inputs {
-		inQuery[input.ArgType()] = true
-	}
 
 	typeToValue := map[reflect.Type]reflect.Value{}
 	for _, arg := range args {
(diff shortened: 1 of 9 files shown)


letFunny (approved):
Already reviewed as a subset of the other PR.

internal/expr/bindinputs.go
```diff
@@ -52,33 +40,108 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
 		typeToValue[t] = v
-		if !inQuery[t] {
-			// Check if we have a type with the same name from a different package.
-			for _, input := range te.inputs {
-				if t.Name() == input.ArgType().Name() {
-					return nil, fmt.Errorf("parameter with type %q missing, have type with same name: %q", input.ArgType().String(), t.String())
+	}
+
+	// Generate SQL and query parameters.
+	params := []any{}
```
manadart: I generally leave slices unallocated unless:
- I know in advance what the length will be or;
- Empty means something special versus nil.

So `var params []any`.

internal/expr/bindinputs.go
```diff
@@ -52,33 +40,108 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+			for i, oc := range te.outputColumns {
+				sqlStr.WriteString(oc.sql(outputCount))
+				if i != len(te.outputColumns)-1 {
+					sqlStr.WriteString(", ")
 				}
+				outputCount++
+				outputs = append(outputs, oc.output)
 			}
-			return nil, fmt.Errorf("%s not referenced in query", t.Name())
+		case *bypass:
```
manadart, line 77: Given that `typedExpression` is a completely synthetic interface, just delete it.
- `TypeBoundExpr` becomes `[]any`. 
- Add a default case here that returns an error for types we don't recognise.
- Delete all the marker methods and implementation checks below.

If we come up with a unifying indirection/abstraction, we can reintroduce an interface. As it is, it's just noise.

internal/expr/bindtypes.go
```diff
@@ -44,80 +43,62 @@ func (pe *ParsedExpr) BindTypes(args ...any) (te *TypedExpr, err error) {
...
-	// Check and expand each query expr.
+	// Bind types to each expression.
+	typedExprs := []typedExpression{}
+	outputUsed := map[typeinfo.Output]bool{}
+	var te typedExpression
 	for _, expr := range pe.exprs {
 		switch e := expr.(type) {
 		case *inputExpr:
-			inLoc, err := bindInputTypes(argInfo, e)
+			te, err = bindInputTypes(e, argInfo)
```
manadart, line 53: Here we go.

- Make `typedExpression` generic.
- It has a method `getTypedExpr[T]` or some such.
- So for example, `inputExpr` implements `typedExpression[typedInputExpr]`.
- `bindInputTypes` and `bindOutputTypes` become the bodies of `getTypedExpr`.
- `bypass` just returns itself.

Then it's starting to look nice.

manadart (commented):
I think we have a way to clean this up quite nicely.

internal/expr/bindinputs.go
letFunny: I agree. Just want to point out that the only benefit for having a synthetic interface is that it documents which types are allowed and which are not, and it is in code instead of a comment which might get outdated.

internal/expr/bindtypes.go
Aflynn50: I would like to make the `bindInputTypes` and `bindOutputTypes` functions part of an interface, but I can't find a way to make the generics implementation work nicely.

If `expression` implements this generic interface then there is the problem that I cannot have a list of `typedExpression[T]` where `T` has a different instantiation at each index. This means I need to assert at some point that the expressions do actually implement this interface, and because they need to implement the one with the right instantiation of `T` I would still need to enumerate the types in a switch case statement.

I was thinking I could make `bindTypes` a method of expression, and it can return any so that we can get rid of the `typedExpression` interface.

Aflynn50: I've implemented `bindTypes` as a method on `expression`

letFunny:
I really like the latest changes of moving the `bindTypes` to a method on both input and output expressions. However, now that I see the code I personally disagree with removing the expression types and changing it to `any`. The new signature:
```go
func (b *bypass) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
```
gives me little information about what the return is. I have to either look at the implementation or look at the comments (which can get outdated). Before it would have been:
```go
func (b *bypass) bindTypes(argInfo typeinfo.ArgInfo) (*typedExpression, error) {
```

I my personal opinion, I think the added readability is worth it even if it means having a dummy interface as a substitute for a proper enum type. I would like to know your thoughts on this tradeoff, do you think it is not worth it?

Lastly, this is all based on the latest commit, if we add generics the code might look different enough that it will not be applicable.

internal/expr/bindinputs.go
Aflynn50: Changed

internal/expr/bindinputs.go
```diff
@@ -52,31 +40,86 @@ func (te *TypedExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+				if i != len(te.outputColumns)-1 {
+					sqlStr.WriteString(", ")
+				}
+				outputCount++
+				outputs = append(outputs, oc.output)
+			}
+		case *bypass:
+			sqlStr.WriteString(te.chunk)
+		default:
+			return nil, fmt.Errorf("invalid expression type %t", te)
```
manadart: My mistake. This should be `%T`.

Aflynn50: Fixed, also added the internal error prefix.

manadart (approved):

PR #128, opened 2023-12-14
refactor: remove cache

letFunny:
There are several issues including:
* Overhead of finalizers.
* Limitation in the cache design when the max number of connections is
  1, because of the sql standard library works.
* It is not clear whether preparing statements in SQLite shows an
  improvement in performance.
which require time investment and through testing to qualify performance
gains. Right now, we are going to focus on delivering the missing
features for adoption of SQLair. That is why we have decided to remove
the cache and revisit the conversation in the future when we can commit
with the necessary investment.


sqlair.go
```diff
@@ -183,28 +165,27 @@ func (q *Query) Iter() *Iterator {
...
 	var cols []string
-	var close func() error
-	sqlstmt := q.sqlstmt
-	if q.tx != nil {
-		sqlstmt = q.tx.sqltx.Stmt(q.sqlstmt)
-		close = sqlstmt.Close
-	}
 	if q.pq.HasOutputs() {
-		rows, err = sqlstmt.QueryContext(q.ctx, q.pq.Params()...)
+		if q.tx != nil {
```
Aflynn50: Instead of checking if the query is a transaction here we could save this function as a closure when we create the query object and remove the `tx` field. That way we have all the logic that differentiates transactions and regular queries in one place.

Before the cache was added [there was an interface](https://github.com/canonical/sqlair/pull/81/files#diff-ef99bcd9bee844cd2964fccf754729a6bc3ba08ea4054a5d7a40d82bb7457067L72) `QuerySubstrate` that captured `QueryContext` and `ExectContext`. I think also worked pretty well but if we can save it as a closure we can also get rid of the three fields there that are parameters for Query and Exec.

Aflynn50 (commented):
Looks good

sqlair.go
letFunny: Then we need to create two closures, one for `Query` and one for `Exec` . I remember the `querySubstrate` interface and I thought of doing it for this PR. I thought this way it was simpler and more evident, it is only 3 extra lines and that saves the interface indirection. The same thing happens with closures, you have to find where they were created to see how it is working, in this particular case of few extra lines, I think it is not worth it.

However, if you think it will be clearer by having either closures or an interface we can do that, just wanted to point out the rationale behind my decision.

Aflynn50: Is there any reason we can't just have one closure called `Run` that is `Query` or `Exec` depending on `pq.HasOutputs`? It could return both the rows and results and the code here could action them as it needs to. I would prefer if the code for executing the query was in `Query` itself rather than in `Iter` because those functions are more concerned with how you use the results rather than how the query is executed. To me it feels more natural to keep that in the `Query` function that is different for queries on dbs and transactions.

letFunny: I see. I have tried the idea in the code with the hopes of simplifying the iterator and isolating the logic but I am afraid it is not possible. I cannot remove `pq` nor `tx` from the `Query` object. So, while the closure abstracts some of the logic away, `Iter` still needs to check if it is a transaction or not.

Do you still think it is worth it? I am still reluctant to introduce the closure abstraction if it is not going to simplify the rest.

Aflynn50: I had a go at implementing it and I actually found something interesting. The only use of the `q.tx` stored in the `Query` outside of the above is to check if the `tx` is done. It seems like the check is not actually needed. The custom `sqlair.ErrTXDone` was added so that if you try and run a query on a closed transaction you get that error and not one saying `"cannot prepare a statement a on closed transaction"` (which wouldn't make sense because as far as the user is concerned they are not preparing a statement).

If  the underlying `sqltx` is closed then it will return the same error anyway, and since we set `sqlair.ErrTXDone = sql.ErrTxDone` the user can still check for a sqlair error and find it.

The check in `Query` still has some value since it stops the `Query` being generated when the `tx` is closed so its good that we have it, though its main reason for existence is gone with the cache.

That being said, it means we can remove the `tx` from the cache. We still need `pq` for the `ScanArgs` function, but I think replacing the query with a closure would work pretty nicely.

I can send you that patch for the code I wrote, remind me in the new year!

letFunny: You are right that the error check is not needed because it fails later with the same error. I have changed the code and uploaded it so that we can discuss about something more concrete. The problem was that even if we move the code into a closure we still need to check whether we have outputs or not:
```go
        rows, result, err = q.run()
	if q.pq.HasOutputs() {
		if err == nil { // if err IS nil
			cols, err = rows.Columns()
		}
	}
```
The ideal scenario would be that the closure does everything needed to run the query and the surrounding code does not care about checking the outputs. We could stuff the columns into the closure and return more things which I am not a fan of.

In general I see the advantages of using the closures approach to decouple how you run it but I can see the code becoming more difficult to read as well.

sqlair.go
```diff
@@ -175,36 +167,19 @@ func (q *Query) Iter() *Iterator {
...
-	var rows *sql.Rows
-	var err error
 	var cols []string
-	var close func() error
-	sqlstmt := q.sqlstmt
-	if q.tx != nil {
-		sqlstmt = q.tx.sqltx.Stmt(q.sqlstmt)
-		close = sqlstmt.Close
-	}
+	rows, result, err := q.run()
```
manadart: This is a good change. It makes semantic sense to move this decision from `Iter` to `Query`.

sqlair.go
```diff
@@ -81,14 +76,12 @@ func (db *DB) PlainDB() *sql.DB {
...
-	tx      *TX // tx is only set for queries in transactions.
-	// Persist statement and db so the sql.Stmt is not closed until the Query is dropped.
-	stmt *Statement
-	db   *DB
+	// run executes the Query against the db or the tx.
+	run func() (*sql.Rows, sql.Result, error)
+	ctx context.Context
+	err error
+	pq  *expr.PrimedQuery
+	db  *DB
```
manadart: However, this `db` is now unused, so we can remove it.

sqlair.go
```diff
@@ -389,10 +346,9 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 }
 
 type TX struct {
-	sqltx   *sql.Tx
-	sqlconn *sql.Conn
-	db      *DB
-	done    int32
+	sqltx *sql.Tx
+	db    *DB
```
manadart: And since it's only used here to _pass_ to `Query`, we can remove it here too.

sqlair.go
```diff
@@ -476,14 +422,14 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
 		return &Query{ctx: ctx, err: err}
 	}
 
-	sqlstmt, ok := stmtCache.lookupStmt(tx.db, s)
-	if !ok {
-		sqlstmt, err = tx.sqlconn.PrepareContext(ctx, pq.SQL())
-		if err != nil {
-			return &Query{ctx: ctx, err: err}
+	run := func() (rows *sql.Rows, result sql.Result, err error) {
```
manadart: The amount of knowledge you need from `pq` in both places where this closure is generated suggests that it might be better sourced from `PrimedQuery`.

manadart (commented):
Some good changes, but we can take it further.

sqlair.go
letFunny: You are right! These were in part used to prevent running the finalizers which is no longer happening.

sqlair.go
letFunny: You are right! These were in part used to prevent running the finalizers which is no longer happening.

sqlair.go
letFunny: That would change the public API. Right now you can store the output of query and run it several times. If we instead moved it to `Query` then `.Run()` will return a stale copy of the data and always the same one.

If we want to do that then we can go ahead but then `Run()` would not be the appropriate name any more.

sqlair.go
letFunny: You mean having a function in `PrimedQuery` that takes a TX or a DB and returns the closure?

I put it in `sqlair.go` because one of the project organization ideas was to not use the `sql` package inside the internal packages (there are only some exceptions right now such as when creating the named parameters). Do you think it is better to move it?

manadart: Yes, that's what I mean. I'd move it, particularly if there are already departures from the convention.

letFunny: I have had a go at implementing the change and we would need to revive the `PrepareSubstrate` interface which we just removed in other to be able to pass `TX` or `DB` to the same function (that will return the closure). Additionally, when implementing the cache we are going to want to move the function back out again.

For context, the only exceptions today are:

bindinputs.go:63
```go
namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val.Interface())
```
member.go:10
```go
var scannerInterface = reflect.TypeOf((*sql.Scanner)(nil)).Elem()
```
which are only wrapping sql types.

manadart (approved):
Thanks for the changes. I have some ideas around supply of the `run` method, but what we have here can be iterated on.

PR #129, opened 2023-12-15
bugfix: Fix duplicate output error

Aflynn50:
Fix error where a single map member can be used as an output multiple times in a query.

The `outputUsed` map mapped `typeinfo.Output` to booleans recording if the output had been used. For maps the `mapKey` object that implemented the `typeinfo.Output` has a pointer receiver. It was possible to have two identical `mapKey` structs with different pointers.

Here, the error is fixed by indexing `outputUsed` with a string representation of the `typeinfo.Output` using its `String` method.

### DIFF
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -64,10 +64,10 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 			}
 
 			for _, oc := range toe.outputColumns {
-				if ok := outputUsed[oc.output.Repr()]; ok {
+				if ok := outputUsed[oc.output.Identifier()]; ok {
 					return nil, fmt.Errorf("%s appears more than once in output expressions", oc.output.Desc())
 				}
-				outputUsed[oc.output.Repr()] = true
+				outputUsed[oc.output.Identifier()] = true
 			}
 			te = toe
 		case *bypass:
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -67,10 +67,10 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 			}
 
 			for _, oc := range toe.outputColumns {
-				if ok := outputUsed[oc.output.String()]; ok {
+				if ok := outputUsed[oc.output.IDString()]; ok {
 					return nil, fmt.Errorf("%s appears more than once in output expressions", oc.output.String())
 				}
-				outputUsed[oc.output.String()] = true
+				outputUsed[oc.output.IDString()] = true
 			}
 			te = toe
 		case *bypass:
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -48,7 +48,10 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 
 	// Bind types to each expression.
 	var typedExprs TypeBoundExpr
-	outputUsed := map[typeinfo.Output]bool{}
+	// outputUsed records if the string representation of an output appears
+	// more than once in the query. If the string appears more than once then
+	// there is ambiguity in the query.
+	outputUsed := map[string]bool{}
 	var te any
 	for _, expr := range pe.exprs {
 		switch e := expr.(type) {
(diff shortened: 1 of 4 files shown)


letFunny:
The change itself looks good but I think we need to investigate deeper. Do we know why there are two pointers in the first case? I remember discussing this but I cannot see it in the PR description.

letFunny:
I see that the problem is that we are creating a pointer [here](https://github.com/canonical/sqlair/blob/main/internal/typeinfo/arginfo.go#L118). Why don't we remove the pointer receiver for `mapInfo` which solves the problem directly. I do not think copying the `reflect.Type` field is that bad because it is already an interface.

Aflynn50:
> I see that the problem is that we are creating a pointer [here](https://github.com/canonical/sqlair/blob/main/internal/typeinfo/arginfo.go#L118). Why don't we remove the pointer receiver for `mapInfo` which solves the problem directly. I do not think copying the `reflect.Type` field is that bad because it is already an interface.

We could remove the pointer receiver for `mapInfo` but it would be a bit of a smell because `structField` would still be a pointer and we would have one thing implementing the interface with a pointer and the other without. 

I've added a method `IDString` instead which is a identifier string for the `mapKey`/`structField` that can be used for `outputUsed` in `bindTypes`.

internal/expr/bindtypes.go
```diff
@@ -45,7 +45,10 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 
 	// Bind types to each expression.
 	var typedExprs TypeBoundExpr
-	outputUsed := map[typeinfo.Output]bool{}
+	// outputUsed records if the string representation of an output appears
+	// more than once in the query. If the string appears more than once then
+	// there is ambiguity in the query.
```
letFunny: I think now this comment is a little bit superfluous because it is merely stating that the String representation is unique, which seems to be the case now that we have a specific function.

internal/typeinfo/member.go
```diff
@@ -13,6 +13,7 @@ var scannerInterface = reflect.TypeOf((*sql.Scanner)(nil)).Elem()
 type ValueLocator interface {
 	ArgType() reflect.Type
 	String() string
+	IDString() string
```
letFunny: [Nit]: Can we change the name to `ID`? The `String` bit is already part of the type signature.

letFunny: And can we state in a comment here or in the implementers that the invariant is that this function **must** return a unique representation which lets us differentiate between the different objects?

letFunny (commented):
Thanks! As discussed offline, I agree that this is the correct approach.

internal/typeinfo/member.go
Aflynn50: Name changed to ID, I was 50/50 on which to choose.

I want to be careful around language here, we don't actually want a unique ID for each ValueLocator. In a way that is the problem that caused the bug in the first place since we were differentiating between map keys that looked the same but were different objects. We want to check weather the textual representation of the members looks the same.

I have added a comment to try and capture what the ID represents.

letFunny: I see what you mean. To me ID was the identifier for the underlying concept, not the id for each concrete instantiation to be different from each other. Maybe there is a better name like `repr` in Python or `hash` or something like that.

letFunny: The problem with the name and comment now is that it is hard to know why there is a need for a new function when we have `String` already.

Aflynn50: I've changed the `String` function to `Desc` meaning description and the `ID` function to `Repr` for representation. Hopefully that clears it up a bit.

letFunny (approved):
I was a bit reluctant to change the `String` method which is the standard but I think the new methods make it clearer. If we had `String` there would be some confusion as to why we need `Repr`.

internal/typeinfo/member.go
```diff
@@ -15,7 +15,11 @@ var scannerInterface = reflect.TypeOf((*sql.Scanner)(nil)).Elem()
 // ValueLocator specifies how to locate a value in a SQLair argument type.
 type ValueLocator interface {
 	ArgType() reflect.Type
-	String() string
+	// Desc returns a written description of the ValueLocator for error messages.
+	Desc() string
+	// Repr returns a string representation for the ValueLocator specifying the
+	// type and, if present, member.
+	Repr() string
```
manadart: Call this `Identifier` with a comment that it uniquely identifies the value in the context of a query.
It should appear above `Desc`.

manadart (commented):
Looks fine. I suggest one minor change.

internal/typeinfo/member.go
Aflynn50: Changed.

internal/typeinfo/member.go
```diff
@@ -143,8 +144,9 @@ func (f *structField) Desc() string {
 	return "tag \"" + f.tag + "\" of struct \"" + f.structType.Name() + "\""
```
manadart: This reads much nicer as `fmt.Sprintf("tag %q of struct %q", f.tag, f.structType.Name())`.
Same for the map above.

manadart (commented):
Should have picked this up before; apologies. Just needs tidier string creation.

internal/typeinfo/member.go
Aflynn50: Changed.

manadart (approved):
Thanks for seeing this one through.

PR #130, opened 2024-01-08
Add type support for slice inputs

Aflynn50:
This PR adds support for binding slice types to statements and passing slice values to queries is added using the syntax added in #114.

Slices can now be used in input expressions. For example:
`SELECT ... FROM p WHERE id IN ($S[:])`
A slice can be used anywhere that an input expression can be used in the query, they are not limited to just `IN` clauses.

This PR also adds a type `sqlair.S` of type `[]any` which can be used to avoid having to declare a custom type.


internal/expr/expr_test.go
```diff
@@ -334,6 +338,49 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+}, {
+	summary:        "slice of mixed types",
+	query:          "SELECT name FROM person WHERE id IN ($S[:])",
+	expectedParsed: "[Bypass[SELECT name FROM person WHERE id IN (] Input[S[:]] Bypass[)]]",
+	typeSamples:    []any{sqlair.S{}},
+	inputArgs:      []any{sqlair.S{1, "two", 3.0}},
+	expectedParams: []any{1, "two", 3.0},
+	expectedSQL:    "SELECT name FROM person WHERE id IN (@sqlair_0, @sqlair_1, @sqlair_2)",
+}, {
+	// No error is throw for an empty slice because we do not know where in the
```
letFunny:
```suggestion
	// No error is thrown for an empty slice because we do not know where in the
```

The wording of "we do not know where in the query and how the slice input ..." is a little bit confusing. What about iterating over something like:
```
// No error is throw for an empty slice because we do not want to force the user to handle it as a different case, it is their choice.
```

doc.go
```diff
@@ -1,8 +1,8 @@
 /*
 SQLair is a convenience layer for SQL databases that embeds Go structs and maps directly into SQL queries.
 
-The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query that correspond to Go structs or maps.
-This allows the user to specify the Go structs/maps they want in the SQL query itself whilst allowing the full power of SQL to be utilised.
+The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query correspond to Go values.
+This allows the user to specify Go values (such as structs, maps and slices) to use in the SQL query itself whilst allowing the full power of SQL to be utilised.
```
letFunny: I find the sentence a little hard to read, I dont know if it is because the parentheses break the flow or because of the concrete words. What do you think of the following (I have only removed "to use")?

```
This allows the user to specify Go values (such as structs, maps and slices) in the SQL query itself whilst allowing the full power of SQL to be utilised.
```

doc.go
```diff
@@ -33,12 +33,19 @@ Note that in the SQLair `db` tags (i.e. the column names) appear in the input/ou
 
 # Syntax
 
-SQLair input expressions take the format:
+The SQLair expressions specify Go values to use as query inputs or outputs. The
+input expressions can take values from structs maps and slices whereas the
```
letFunny:
```suggestion
input expressions can take values from structs, maps and slices whereas the
```

Additionally, should this and the next sentence be `or` or `and`?

internal/expr/bindinputs.go
```diff
@@ -29,12 +29,24 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+			}
+			if t.Name() == "" {
+				return nil, fmt.Errorf("cannot use anonymous %s", t.Kind())
+			}
+		case reflect.Struct:
+			if t.Name() == "" {
+				return nil, fmt.Errorf("cannot use anonymous %s", t.Kind())
+			}
+		default:
+			return nil, fmt.Errorf("need supported value, got %s", t.Kind())
```
letFunny: I like the "supported value" terminology a lot. What do you think of using that across all the code instead of using both "supported value" and "valid value"?

internal/expr/bindtypes.go
```diff
@@ -95,7 +95,12 @@ func bindInputTypes(e *inputExpr, argInfo typeinfo.ArgInfo) (te *typedInputExpr,
 			return nil, err
 		}
 	case sliceAccessor:
-		return nil, fmt.Errorf("slice support not implemented")
+		input, err = argInfo.InputSlice(a.typeName)
```
letFunny, line 101: I like the code as is better than as a method. That said, if you wanted to change it to a more OOP pattern you could do it by creating a method on `valueAccessor` that interally calls `InputSlice` or `InputMember` for each of the types (so called double dispatch).

internal/expr/expr_test.go
```diff
@@ -59,8 +63,8 @@ var tests = []struct {
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
-	summary:        "spaces and tabs",
-	query:          "SELECT p.* 	AS 		   &Person.*",
+	summary: "spaces and tabs",
+	query: "SELECT p.* 	AS 		   &Person.*",
```
letFunny: [Nit] Can you run `go fmt` before merging?

internal/expr/expr_test.go
```diff
@@ -650,17 +717,17 @@ func (s *ExprSuite) TestMapError(c *C) {
 		"all output into map star",
 		"SELECT &M.* FROM person WHERE name = 'Fred'",
 		[]any{sqlair.M{}},
-		"cannot prepare statement: output expression: columns must be specified for non-struct type: &M.*",
+		"cannot prepare statement: output expression: cannot use map with asterisk unless columns are specified: &M.*",
```
letFunny, line 747: We sometimes call it `asterisk` and sometimes `star`, I think we should only use one of them.

internal/typeinfo/arginfo.go
```diff
@@ -43,9 +43,9 @@ func GenerateArgInfo(typeSamples []any) (ArgInfo, error) {
 			}
 			argInfo[t.Name()] = info
 		case reflect.Pointer:
-			return nil, fmt.Errorf("need struct or map, got pointer to %s", t.Elem().Kind())
+			return nil, fmt.Errorf("need non-pointer type, got pointer to %s", t.Elem().Kind())
```
letFunny, line 49: Should we make it `supported type` for the error messages to be uniform?

internal/typeinfo/arginfo.go
```diff
@@ -87,7 +87,16 @@ func (argInfo ArgInfo) AllStructOutputs(typeName string) ([]Output, []string, er
...
 	if !ok {
-		return nil, nil, fmt.Errorf("columns must be specified for non-struct type")
+		errStr := ""
+		switch arg.typ().Kind() {
+		case reflect.Map:
+			errStr = "cannot use %s with asterisk unless columns are specified"
+		case reflect.Slice:
+			errStr = "cannot use %s with asterisk"
+		default:
+			errStr = "internal error: unknown invalid arg type %s"
```
letFunny: [Nit]: What about a return instead or creating the error directly using `fmt.Errorf`? It is a little bit weird to do a switch on `arg.typ().Kind()` just to add it later.

internal/typeinfo/arginfo_test.go
```diff
@@ -241,5 +245,27 @@ func (s *typeInfoSuite) TestGenerateArgInfoMapError(c *C) {
 
 	_, _, err = argInfo.AllStructOutputs("myMap")
 	c.Assert(err, Not(IsNil))
-	c.Assert(err.Error(), Equals, "columns must be specified for non-struct type")
+	c.Assert(err.Error(), Equals, `cannot use map with asterisk unless columns are specified`)
+
+	_, err = argInfo.InputSlice("myMap")
+	c.Assert(err, Not(IsNil))
```
letFunny: Just saw this, there is a builtin checker called `NotNil`, can we use that instead in all the tests?

internal/typeinfo/arginfo_test.go
```diff
@@ -241,5 +245,27 @@ func (s *typeInfoSuite) TestGenerateArgInfoMapError(c *C) {
...
+	c.Assert(err.Error(), Equals, `cannot use slice syntax with map`)
+}
+
+func (s *typeInfoSuite) TestGenerateArgInfoSliceError(c *C) {
```
letFunny: The tests are good but I am wondering if we could instead use table tests which will make it easier to add them later. We would have a test for `AllStructOutput` another for `InputMember`, etc. so sort of the opposite organization. Another benefit is that you can see all the test cases for a function at once making it easier to spot if we are missing any combination.

internal/typeinfo/slice.go
```diff
@@ -0,0 +1,28 @@
+package typeinfo
```
letFunny: How is the organization working? I can see a lot of common interfaces in `member.go`  for example `Input` and `Output` and then slices are going on a different file altogether.

internal/typeinfo/slice_test.go
```diff
@@ -0,0 +1,45 @@
...
+	. "gopkg.in/check.v1"
+)
+
+func (*typeInfoSuite) TestLocateParamsSlice(c *C) {
+	type S []any
+
+	argInfo, err := GenerateArgInfo([]any{S{}})
+	c.Assert(err, IsNil)
+
+	s := S{1, 2}
```
letFunny: I think it would not take a lot of refactoring to make this a test array. That way we can add tests easily in the future and we can test boundary conditions like empty slice, mixed types, etc.

internal/typeinfo/slice_test.go
```diff
@@ -0,0 +1,45 @@
...
+	vals, err := input.LocateParams(typeToValue)
+	c.Assert(err, IsNil)
+	c.Assert(vals, HasLen, len(s))
+
+	for i := 0; i < len(s); i++ {
+		c.Assert(vals[i].Interface(), Equals, s[i])
+	}
+}
+
+func (*typeInfoSuite) TestLocateParamsSliceError(c *C) {
```
letFunny: If adding table tests above this can probably be added as a test case.

letFunny (commented):
First pass, code is looking very good! Left comments mostly about naming and minor ones about test organization.

internal/expr/expr_test.go
Aflynn50: Sounds good, how about:
```
// No error is throw for when the user passes an empty slice because we do
// not want to limit the use of slices to only cases we have foreseen.
```

doc.go
Aflynn50: Yea, I think that's better

doc.go
Aflynn50: Added the comma, I've changed the `whereas` to `while` which I think fits a bit better. There are already a few `and`s in the sentence.

internal/expr/bindinputs.go
Aflynn50: Yea, that sounds good. I'd also like to replace this in the future with a list of the types from `bindtypes` since at this stage we do know exactly what we're looking for.

internal/expr/bindtypes.go
Aflynn50: Yea, I would like to do that. Though I think first I need to add the separate value accessor for the asterisk. I had been working on that before Christmas a bit, I plan to pick it back up this week.

internal/expr/expr_test.go
Aflynn50: For some reason `gofmt` is doing this, i.e. removing the tab, [seems to be this issue](https://github.com/golang/go/issues/51910). This is a new issue for us because we only introduced the named fields a few PRs ago.

If this behaves the same on your PC then I think it'd be best just to leave it rather than fight gofmt. The issue is closed but its still broken for me with the latest version of go installed. (Check on your PC though, I'm assuming that nvim is using the snap gofmt but I'm not actually certain.)

internal/expr/expr_test.go
Aflynn50: `asterisk` is less ambiguous for me/what I'm used to. But if there's a good reason to use `star` I'm happy to switch.

internal/typeinfo/arginfo.go
Aflynn50: In this case its a little bit arbitrary that we don't accept pointers (this code is called by from typebind). I phrased it like that so that its as obvious as possible how to fix the problem. Its not really well documented right now that we don't accept pointers to the types either.

internal/typeinfo/arginfo.go
Aflynn50: Fair enough, changed it to just return. I wrote it like this because the line length was annoying me but its not actually that bad.

internal/typeinfo/arginfo_test.go
Aflynn50: Changed to table driven tests here. I agree that its better.

internal/typeinfo/slice.go
Aflynn50: Ah, I didn't think when I created the file. I've changed member.go to valuelocator.go and moved slice.go in there.

internal/typeinfo/slice_test.go
Aflynn50: I've changed it to a table driven test here, there are only two cases right now though.

internal/typeinfo/slice_test.go
Aflynn50: I would prefer to leave this as it is. Generally in the codebase we are keeping the error tests separate from the acceptance tests. I did try this as a table driven test in its own right but this case doesn't really suit it very well.

internal/expr/expr_test.go
letFunny: That works.

internal/expr/bindtypes.go
letFunny: We can keep it as is and then refactor in another PR, in fact that would be the best thing in my opinion.

internal/expr/expr_test.go
letFunny: In my case running `go fmt` actually changed the line, that is why I commented in the first place :P
```
go version go1.20.3 linux/amd64
```

internal/expr/expr_test.go
letFunny: To me both of them look good, let's keep asterisk. But then, change all the places that say `star` in the code.

internal/typeinfo/arginfo.go
letFunny: Good idea then, keep it as is.

internal/typeinfo/slice_test.go
letFunny: Thanks, right now there are only two tests but in the future we can add more easily :)

letFunny: Tests with arrays of more than one type would be nice if that can be included in this PR.

internal/typeinfo/slice_test.go
letFunny: Both ways seem to be standard in Go, I agree with you that it is the convention to separate them in SQLair.

letFunny: And even in this case I can think of a couple of more tests like having a known type and an unknown one, only unknown, etc.

internal/expr/expr_test.go
Aflynn50: Ah yea, will see if I can get that fixed in nvim

internal/expr/expr_test.go
Aflynn50: Ok, will do this by increments when it comes up rather than all in this PR.

internal/typeinfo/slice_test.go
Aflynn50: Added that, `S` is a list of any so its not too much of a special case.

internal/typeinfo/slice_test.go
Aflynn50: I've added that extra tests but its really just testing a map lookup and error message that is at the start of all `LocateParams` methods rather than any functionality related to slices. I don't think there's anything else to really add here, the sanitizing of the inputs is done by `bindinputs` and has already been handled by the time `LocateParams` is called.

internal/expr/expr_test.go
letFunny: Then we have two concepts for the same things, some error messages and code say one thing and some say the other. Then I would say keep it as `star` and change it to `asterisk` in another PR to avoid having both at the same time.

Aflynn50: The only mentions of "star" in the codebase are in test names and variable names, all of the variable names it appears in will disappear when we put methods on `valueAccessor`. The test names I'm really not worried about. It doesn't appear anywhere in the error messages. I think its fine just to leave it for now.

letFunny: Ok let's call it a matter of preference. Because it is not in the error messages, let's leave it as is.

letFunny (approved):
Nice, the support for slices is finally materializing.

doc.go
```diff
@@ -4,8 +4,8 @@
 /*
 SQLair is a convenience layer for SQL databases that embeds Go structs and maps directly into SQL queries.
 
-The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query that correspond to Go structs or maps.
-This allows the user to specify the Go structs/maps they want in the SQL query itself whilst allowing the full power of SQL to be utilised.
+The SQL syntax is expanded with SQLair input and output expressions which indicate the parts of the query correspond to Go values.
```
manadart: Should be "expressions _that_ indicate...", being necessarily a restrictive clause.

internal/expr/bindinputs.go
```diff
@@ -32,12 +32,24 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
 		if v.Kind() == reflect.Invalid || (v.Kind() == reflect.Pointer && v.IsNil()) {
-			return nil, fmt.Errorf("need struct or map, got nil")
+			return nil, fmt.Errorf("need supported value, got nil")
 		}
 		v = reflect.Indirect(v)
 		t := v.Type()
-		if v.Kind() != reflect.Struct && v.Kind() != reflect.Map {
-			return nil, fmt.Errorf("need struct or map, got %s", t.Kind())
+		switch v.Kind() {
+		case reflect.Map, reflect.Slice:
```
manadart: With this method now exceeding 80 LoC, I'd be tempted to extract this first part into a validation method/func.

internal/typeinfo/arginfo.go
```diff
@@ -90,7 +90,14 @@ func (argInfo ArgInfo) AllStructOutputs(typeName string) ([]Output, []string, er
...
 	si, ok := arg.(*structInfo)
 	if !ok {
-		return nil, nil, fmt.Errorf("columns must be specified for non-struct type")
+		switch k := arg.typ().Kind(); k {
+		case reflect.Map:
+			return nil, nil, fmt.Errorf("cannot use %s with asterisk unless columns are specified", k)
+		case reflect.Slice:
+			return nil, nil, fmt.Errorf("cannot use %s with asterisk", k)
+		default:
+			return nil, nil, fmt.Errorf("internal error: unknown invalid arg type %s", k)
```
manadart: "unknown invalid"?

sqlair.go
```diff
@@ -24,6 +24,10 @@ import (
 //	err := q.Get(resultMap) // => sqlair.M{"name": "Fred", "postcode": 10031}
 type M map[string]any
 
+// S is a slice type that, as with other named slice types, can be used with
+// SQLair to pass a slice of input values.
+type S []any
```
manadart, line 29: Was this discussed prior as an addition the public API? I'm not sure we get much from it, give how non--specific it is. At least `M` designates index by `string`.

package_test.go
```diff
@@ -194,6 +197,20 @@ func (s *PackageSuite) TestValidIterGet(c *C) {
...
+	}, {
+		summary:  "simple in",
+		query:    "SELECT * AS &Person.* FROM person WHERE id IN ($S[:])",
+		types:    []any{Person{}, sqlair.S{}},
+		inputs:   []any{sqlair.S{30, 35, 36, 37, 38, 39, 40}},
+		outputs:  [][]any{{&Person{}}, {&Person{}}, {&Person{}}},
+		expected: [][]any{{&Person{30, "Fred", 1000}}, {&Person{40, "Mary", 3500}}, {&Person{35, "James", 4500}}},
+	}, {
+		summary:  "complex in",
+		query:    "SELECT * AS &Person.* FROM person WHERE id IN ($Person.id, $S[:], $Manager.id, $IntSlice[:], $StringSlice[:])",
```
manadart, line 209: I don't think this should be valid syntax.

internal/typeinfo/valuelocator.go
```diff
@@ -156,6 +156,31 @@ func (f *structField) LocateScanTarget(typeToValue map[reflect.Type]reflect.Valu
 	return val.Addr().Interface(), nil, nil
 }
 
+type slice struct {
+	sliceType reflect.Type
+}
+
+func (s *slice) String() string {
```
manadart: Comments on exported methods.

manadart (commented):
Looks good; a few items to clarify.

internal/expr/bindinputs.go
Aflynn50: I intend to refactor this section into `typeinfo` alongside the similar section at the start of `bindtypes`. Would it be alright to bring that in a separate PR and leave it as it as for now?

internal/typeinfo/arginfo.go
Aflynn50: Changed to "invalid arg type".

sqlair.go
Aflynn50: The `S` type is in the slice support spec. I think it adds convenience at least in that the user does not have to design a custom type, since using `[]any` is not possible.

package_test.go
Aflynn50: This is only valid syntax by virtue of us not parsing the IN statement. All the input expressions in this statement are parsed separately independent of each other (i.e. the comma between each one is a `bypass` part). To reject this syntax we would have to actually parse IN statements and be far more context aware than we are with parsing.

I agree that there is no need to encouraged it however.  In general, if there is a valid use case like this that we do not recommend, should it be included in the tests?

internal/typeinfo/valuelocator.go
Aflynn50: Added.

package_test.go
manadart: I think we should catch it early. If we detect slice syntax, there should be only member in the `IN` list.

Aflynn50: I have some reservations about doing this. In the rest of SQLair we avoid parsing any query syntax other than that which is strictly necessary. For example with output expressions we do not even check that they occur after a `SELECT`. This was a conscious decision to avoid as much dependency on the SQL syntax itself as possible and allow us to handle unforeseen usecases and syntax in other dialects. 

When initially discussing this feature with Gustavo it was decided that we would not tie the slice support to `IN` statements and instead allow them to be used anywhere in the query so as not to restrict support to only the use case we could foresee.

I am also not sure that I understand why allowing other values in an `IN` statement beside the slice is a problem, though I do agree that it is bad practice.  There might be cases such as `... WHERE x IN ($S[:], "invalid")` where the slice from the user only captures a subset of values that should be matched against.

manadart: We already place restrictions on allowable syntax at the parse stage, see PL014.

By allowing mixtures of slices and scalar values to be valid, you're obliged to imply a relation between them in SQL generation, because you need to know that they are placeholders in the same clause. 

Let me take a closer look at this.

Aflynn50: When you say that they are placeholders in the same clause are you referring to the parentheses around the input expressions? 

In the implementation as it is SQLair doesn't know that these are in the same clause.
It parses:
`IN($S[:], $T[:])`
as:
`Bypass["IN("], Input["S[:]"], Bypass[", "], Input["T[:]"], Bypass[")"]`
The bypass parts could be anything as far as SQLair is concerned, including the end of the clause. When we generate the SQL to replace the slice input expression we do not generate any parentheses.

letFunny: Just to add to Alastair's comment, when encountering a slice all we generate in the final SQL is
```
@sqlair_0, @sqlair_1, ..., @sqlair_n
```
we do not include parentheses.

internal/expr/bindinputs.go
```diff
@@ -50,10 +50,13 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 				return nil, err
 			}
 			argTypeUsed[te.input.ArgType()] = true
-			for _, val := range vals {
+			for i, val := range vals {
 				namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val.Interface())
 				params = append(params, namedInput)
 				sqlStr.WriteString("@sqlair_" + strconv.Itoa(inputCount))
+				if i < len(vals)-1 {
```
manadart: You could also write this at the top of the loop if `i != 0`, which would save the `len` call each time.

internal/typeinfo/arginfo.go
```diff
@@ -120,8 +127,20 @@ func (argInfo ArgInfo) getMember(typeName string, memberName string) (ValueLocat
 	case *mapInfo:
 		return &mapKey{name: memberName, mapType: arg.mapType}, nil
 	default:
-		return nil, fmt.Errorf("internal error: arg type %T does not have members", arg)
+		return nil, fmt.Errorf("cannot get named member of %s", arg.typ().Kind())
+	}
+}
+
+func (argInfo ArgInfo) InputSlice(typeName string) (Input, error) {
```
manadart, line 135: Exported method needs comment.

internal/typeinfo/valuelocator.go
```diff
@@ -172,6 +174,44 @@ func (f *structField) LocateScanTarget(typeToValue map[reflect.Type]reflect.Valu
...
+	params := []reflect.Value{}
+	for i := 0; i < sv.Len(); i++ {
+		params = append(params, sv.Index(i))
+	}
+	return params, nil
+}
+
 // locateValue locates the value corresponding to the given type in the
 // typeToValueMap and returns an error message if it cannot be found.
 func locateValue(typeToValue map[reflect.Type]reflect.Value, typ reflect.Type) (reflect.Value, error) {
```
manadart: You've started something here, and left it unfinished. It isn't used.

manadart (commented):
This looks good. I like the test coverage. 

Just a couple of small things and it's good to go.

internal/expr/bindinputs.go
Aflynn50: Moved

internal/typeinfo/valuelocator.go
Aflynn50: Ah yea, I started writing locateValue to lookup values in the `typeToValue` map but decided it was a unnecessary overhead and it could just be done inline, must have forgotten to remove it.

manadart (approved):
Nice job.

PR #131, opened 2024-01-12
Add parsing support for input expressions in INSERT statements

Aflynn50:
This PR adds parsing support for two new forms of input expression. These will allow ergonomic type mapping with INSERT statements. The new forms match the existing forms of output expression. More information can be found in  [Spec PL014](https://docs.google.com/document/d/1DMKPS85msymrdbA5-EoP4Sf6b9blSAjtj367EuEmP7Y/edit#heading=h.jytz1gr6mmar).

The new forms of input expression are:

#### 1. Insert all Go values on the right hand side into the table.
```
INSERT INTO table (*) VALUES (<accessorList>)
```
`<accessorList>` is a comma separated list of Go object accessors. 

The object accessors are the form:
`$Struct.tag` - This will cause the value in the field with the “db” tag `tag` of Struct to be inserted into the column with the name `tag`.
`$Map.key` will cause the value at location `key` in Map to be inserted into the column `key`.
`$Struct.*` will cause all values in fields with “db” tags in Struct to be inserted into the column named by the tag.

This syntax allows columns from multiple Go objects to be inserted without separately enumerating the columns. It is expected to be used in the majority of situations.

#### 2. Insert specific columns from Go value.
The syntax makes it easy to insert a small subset of columns from types containing many.

```
INSERT INTO table (<colList>) VALUES (<accessorList>)
```
`<colList>` is a comma separated list of column names and <accessorList> is a comma separated list of Go object accessors.

The different object accessors having the following meanings:
`$Struct.tag` - The tag must match a column `<colList>`. The value in the struct will be inserted into the column tag.
`$Map.key` - The key must match a column `<colList>`. The value at key in the map will be inserted into the column key.
`$Struct.*` - Any values with tags matching a column in `<colList>` will be inserted.
`$Map.*` - columns in `<colList>` that are not captured by another object assessor are drawn from the map. It must be unambiguous which columns are to be found in Map  when the query is prepared or an error will be returned (e.g. a query with two $Map.* accessors is ambiguous).

### DIFF
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,34 +118,34 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
-// insertIntoAsteriskExpr is a form of input expression for insert statements e.g.
+// insertToAsteriskExpr is a form of input expression for insert statements e.g.
 // "(*) VALUES ($Type1.member, $Type2.*)".
-type insertIntoAsteriskExpr struct {
+type insertToAsteriskExpr struct {
 	sources []memberAccessor
 	raw     string
 }
 
-func (e *insertIntoAsteriskExpr) String() string {
+func (e *insertToAsteriskExpr) String() string {
 	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
 }
 
-func (e *insertIntoAsteriskExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *insertToAsteriskExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
-// insertIntoColumnsExpr is a form of input expression for insert statements e.g.
+// insertToColumnsExpr is a form of input expression for insert statements e.g.
 // "(col1, col2, col3) VALUES ($Type.*)".
-type insertIntoColumnsExpr struct {
+type insertToColumnsExpr struct {
 	columns  []columnAccessor
 	typeName string
 	raw      string
 }
 
-func (e *insertIntoColumnsExpr) String() string {
+func (e *insertToColumnsExpr) String() string {
 	return fmt.Sprintf("ColumnInsert[%v [%v.*]]", e.columns, e.typeName)
 }
 
-func (e *insertIntoColumnsExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *insertToColumnsExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,6 +118,8 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
+// insertInput is a form of input expression for insert statements e.g.
+// "(*) VALUES ($Type1.member, $Type2.*)".
 type insertInputExpr struct {
 	raw           string
 	targetColumns []columnAccessor
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,39 +118,39 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
-// asteriskInputExpr is an input expression occurring within an INSERT
+// asteriskInsertExpr is an input expression occurring within an INSERT
 // statement that consists of an asterisk on the left and explicit type accessors
 // on the right. This means that SQLair generates the columns.
 // e.g. "(*) VALUES ($Type1.col1, $Type2.*)".
-type asteriskInputExpr struct {
+type asteriskInsertExpr struct {
 	sources []memberAccessor
 	raw     string
 }
 
 // String returns a text representation for debugging and testing purposes.
-func (e *asteriskInputExpr) String() string {
-	return fmt.Sprintf("AsteriskInput[[*] %v]", e.sources)
+func (e *asteriskInsertExpr) String() string {
+	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
 }
 
-func (e *asteriskInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *asteriskInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
-// columnsInputExpr is an input expression occurring within an INSERT statement
+// columnsInsertExpr is an input expression occurring within an INSERT statement
 // that consists of explicit columns on the left and type accessors on the right.
 // e.g. "(col1, col2, col3) VALUES ($Type.*, $Type2.col1)".
-type columnsInputExpr struct {
+type columnsInsertExpr struct {
 	columns []columnAccessor
 	sources []memberAccessor
 	raw     string
 }
 
 // String returns a text representation for debugging and testing purposes.
-func (e *columnsInputExpr) String() string {
-	return fmt.Sprintf("ColumnInput[%v %v]", e.columns, e.sources)
+func (e *columnsInsertExpr) String() string {
+	return fmt.Sprintf("ColumnInsert[%v %v]", e.columns, e.sources)
 }
 
-func (e *columnsInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -115,7 +115,21 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	if err != nil {
 		return nil, fmt.Errorf("input expression: %s: %s", err, e.raw)
 	}
-	return &typedInputExpr{input}, nil
+	return &typedInputExpr{input: input}, nil
+}
+
+type insertInputExpr struct {
+	raw           string
+	targetColumns []columnAccessor
+	sourceTypes   []memberAccessor
+}
+
+func (ie *insertInputExpr) String() string {
+	return fmt.Sprintf("Input[%+v %+v]", ie.targetColumns, ie.sourceTypes)
+}
+
+func (ie *insertInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
 // sliceInputExpr is an input expression of the form "$S[:]" that represents a
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -139,14 +139,14 @@ func (e *asteriskInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 // asterisk type and explicit columns.
 // e.g. "(col1, col2, col3) VALUES ($Type.*)".
 type columnsInputExpr struct {
-	columns  []columnAccessor
-	typeName string
-	raw      string
+	columns []columnAccessor
+	sources []memberAccessor
+	raw     string
 }
 
 // String returns a text representation for debugging and testing purposes.
 func (e *columnsInputExpr) String() string {
-	return fmt.Sprintf("ColumnInput[%v [%v.*]]", e.columns, e.typeName)
+	return fmt.Sprintf("ColumnInput[%v %v]", e.columns, e.sources)
 }
 
 func (e *columnsInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,9 +118,10 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
-// asteriskInputExpr is an input expression in an INSERT statement with an
-// asterisk on the left meaning that SQLair generates the columns.
-// e.g. "(*) VALUES ($Type1.member, $Type2.*)".
+// asteriskInputExpr is an input expression occurring within an INSERT
+// statement that consists of an asterisk on the left and explicit type accessors
+// on the right. This means that SQLair generates the columns.
+// e.g. "(*) VALUES ($Type1.col1, $Type2.*)".
 type asteriskInputExpr struct {
 	sources []memberAccessor
 	raw     string
@@ -135,9 +136,9 @@ func (e *asteriskInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
-// columnsInputExpr is an input expression in an INSERT statement with a single
-// asterisk type and explicit columns.
-// e.g. "(col1, col2, col3) VALUES ($Type.*)".
+// columnsInputExpr is an input expression occurring within an INSERT statement
+// that consists of explicit columns on the left and type accessors on the right.
+// e.g. "(col1, col2, col3) VALUES ($Type.*, $Type2.col1)".
 type columnsInputExpr struct {
 	columns []columnAccessor
 	sources []memberAccessor
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,34 +118,38 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
-// insertToAsteriskExpr is a form of input expression for insert statements e.g.
-// "(*) VALUES ($Type1.member, $Type2.*)".
-type insertToAsteriskExpr struct {
+// asteriskInputExpr is an input expression in an INSERT statement with an
+// asterisk on the left meaning that SQLair generates the columns.
+// e.g. "(*) VALUES ($Type1.member, $Type2.*)".
+type asteriskInputExpr struct {
 	sources []memberAccessor
 	raw     string
 }
 
-func (e *insertToAsteriskExpr) String() string {
-	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
+// String returns a text representation for debugging and testing purposes.
+func (e *asteriskInputExpr) String() string {
+	return fmt.Sprintf("AsteriskInput[[*] %v]", e.sources)
 }
 
-func (e *insertToAsteriskExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *asteriskInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
-// insertToColumnsExpr is a form of input expression for insert statements e.g.
-// "(col1, col2, col3) VALUES ($Type.*)".
-type insertToColumnsExpr struct {
+// columnsInputExpr is an input expression in an INSERT statement with a single
+// asterisk type and explicit columns.
+// e.g. "(col1, col2, col3) VALUES ($Type.*)".
+type columnsInputExpr struct {
 	columns  []columnAccessor
 	typeName string
 	raw      string
 }
 
-func (e *insertToColumnsExpr) String() string {
-	return fmt.Sprintf("ColumnInsert[%v [%v.*]]", e.columns, e.typeName)
+// String returns a text representation for debugging and testing purposes.
+func (e *columnsInputExpr) String() string {
+	return fmt.Sprintf("ColumnInput[%v [%v.*]]", e.columns, e.typeName)
 }
 
-func (e *insertToColumnsExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *columnsInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -118,19 +118,34 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return &typedInputExpr{input: input}, nil
 }
 
-// insertInput is a form of input expression for insert statements e.g.
+// insertIntoAsteriskExpr is a form of input expression for insert statements e.g.
 // "(*) VALUES ($Type1.member, $Type2.*)".
-type insertInputExpr struct {
-	raw           string
-	targetColumns []columnAccessor
-	sourceTypes   []memberAccessor
+type insertIntoAsteriskExpr struct {
+	sources []memberAccessor
+	raw     string
 }
 
-func (ie *insertInputExpr) String() string {
-	return fmt.Sprintf("Input[%+v %+v]", ie.targetColumns, ie.sourceTypes)
+func (e *insertIntoAsteriskExpr) String() string {
+	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
 }
 
-func (ie *insertInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+func (e *insertIntoAsteriskExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+	return nil, fmt.Errorf("insert input expression not implemented")
+}
+
+// insertIntoColumnsExpr is a form of input expression for insert statements e.g.
+// "(col1, col2, col3) VALUES ($Type.*)".
+type insertIntoColumnsExpr struct {
+	columns  []columnAccessor
+	typeName string
+	raw      string
+}
+
+func (e *insertIntoColumnsExpr) String() string {
+	return fmt.Sprintf("ColumnInsert[%v [%v.*]]", e.columns, e.typeName)
+}
+
+func (e *insertIntoColumnsExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	return nil, fmt.Errorf("insert input expression not implemented")
 }
 
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -449,6 +449,52 @@ func (s *ExprSuite) TestExprPkg(c *C) {
 	}
 }
 
+func (s *ExprSuite) TestInsertInputParser(c *C) {
+	tests := []struct {
+		summary        string
+		query          string
+		expectedParsed string
+	}{{
+		summary:        "insert asterisk",
+		query:          "INSERT INTO person (*) VALUES ($Address.street, $Person.*, $M.team)",
+		expectedParsed: "[Bypass[INSERT INTO person ] Input[[*] [Address.street Person.* M.team]]]",
+	}, {
+		summary:        "insert specified columns",
+		query:          "INSERT INTO person (id, street) VALUES ($Address.*)",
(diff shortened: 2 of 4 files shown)


internal/expr/bindtypes.go
```diff
@@ -115,7 +115,42 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
 	if err != nil {
 		return nil, fmt.Errorf("input expression: %s: %s", err, e.raw)
 	}
-	return &typedInputExpr{input}, nil
+	return &typedInputExpr{input: input}, nil
+}
+
+// asteriskInputExpr is an input expression in an INSERT statement with an
```
letFunny: [Nit] the words "an input expression in an INSERT" dont seem precise to me. We can have several input expressions, (`$Type.member`, `$Type2.*`). What about `asteriskInputExpr is an input expression that consists of an INSERT statement with an asterisk on the left, meaning ...`

internal/expr/bindtypes.go
```diff
@@ -115,7 +115,42 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
...
+// String returns a text representation for debugging and testing purposes.
+func (e *asteriskInputExpr) String() string {
+	return fmt.Sprintf("AsteriskInput[[*] %v]", e.sources)
+}
+
+func (e *asteriskInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
+	return nil, fmt.Errorf("insert input expression not implemented")
+}
+
+// columnsInputExpr is an input expression in an INSERT statement with a single
```
letFunny: Same nit plus another one. Can we keep the other consistent so first mention columns then types, `with explicit columns and a single type` (left then right)?

internal/expr/expr_test.go
```diff
@@ -449,6 +449,60 @@ func (s *ExprSuite) TestExprPkg(c *C) {
...
 
+func (s *ExprSuite) TestInsertInputParser(c *C) {
+	tests := []struct {
+		summary        string
+		query          string
+		expectedParsed string
+	}{{
+		summary:        "insert asterisk",
+		query:          "INSERT INTO person (*) VALUES ($Address.street, $Person.*, $M.team)",
+		expectedParsed: "[Bypass[INSERT INTO person ] AsteriskInput[[*] [Address.street Person.* M.team]]]",
```
letFunny: This is more of an open question not even a suggestion, what do you think of having a Bypass for `VALUES`?
```
[Bypass[INSERT INTO person ] Bypass[VALUES ] AsteriskInput[[*] [Address.street Person.* M.team]]]
```

internal/expr/expr_test.go
```diff
@@ -449,6 +449,60 @@ func (s *ExprSuite) TestExprPkg(c *C) {
...
+		summary:        "insert asterisk (no space)",
+		query:          "INSERT INTO person(*) VALUES ($Person.*)ON CONFLICT DO NOTHING",
+		expectedParsed: "[Bypass[INSERT INTO person] AsteriskInput[[*] [Person.*]] Bypass[ON CONFLICT DO NOTHING]]",
+	}, {
+		summary:        "insert with returning clause",
+		query:          "INSERT INTO address(*) VALUES($Address.*) RETURNING (&Address.*)",
+		expectedParsed: "[Bypass[INSERT INTO address] AsteriskInput[[*] [Address.*]] Bypass[ RETURNING (] Output[[] [Address.*]] Bypass[)]]",
+	}, {
+		summary:        "insert rename columns with standalone inputs",
+		query:          `INSERT INTO person (id, street) VALUES ($Person.address_id, "random string", rand(), $Address.street)`,
```
letFunny: Is this valid SQL? I understand that it does not matter that much but it might throw off someone reading the tests, can you add two more columns?

internal/expr/expr_test.go
```diff
@@ -519,7 +573,7 @@ comment */ WHERE x = $Address.&d`,
 		err:   `cannot parse expression: column 22: missing closing parentheses`,
 	}, {
 		query: "SELECT (name, id) WHERE id = $Person.*",
-		err:   `cannot parse expression: column 30: asterisk not allowed in input expression "$Person.*"`,
+		err:   `cannot parse expression: column 30: invalid asterisk input placement "$Person.*"`,
```
letFunny: Love the error message!

internal/expr/expr_test.go
```diff
@@ -570,6 +624,24 @@ of three lines' AND id = $Person.*`,
...
+		query: "INSERT INTO person * VALUES $Address.*",
+		err:   `cannot parse expression: column 29: invalid asterisk input placement "$Address.*"`,
+	}, {
+		query: "INSERT INTO person * VALUES ($Address.*)",
+		err:   `cannot parse expression: column 30: invalid asterisk input placement "$Address.*"`,
+	}, {
+		query: "INSERT INTO person VALUES ($Address.*)",
+		err:   `cannot parse expression: column 28: invalid asterisk input placement "$Address.*"`,
+	}, {
+		query: "INSERT INTO person VALUES ($Address.*)",
```
letFunny, line 644: Dupe

internal/expr/parser.go
```diff
@@ -280,14 +280,19 @@ loop:
...
 		case ' ', '\t', '\n', '\r', '=', ',', '(', '[', '>', '<', '+', '-', '*', '/', '|', '%':
-			p.advanceByte()
 			break loop
 		}
-		p.advanceByte()
+
+		// An opening parentheses might be the start of an expression.
+		if p.peekByte('(') {
+			break loop
+		}
```
letFunny: I know we discussed this at lengths but I find it very confusing and prone to breaking that we are matching for `(` above and then again here using `peek`. It might seem like a minor thing but when you try to change the parser (e.g. create a sub-parser) it makes things really hard to reason about.

internal/expr/parser.go
```diff
@@ -663,6 +673,15 @@ func (p *Parser) parseTargetTypes() (types []memberAccessor, parentheses bool, o
 	return nil, false, false, nil
 }
 
+// parseInputMemberAccessor parses an accessor preceded by a doller
```
letFunny:
```suggestion
// parseInputMemberAccessor parses an accessor preceded by '$'
```

internal/expr/parser.go
```diff
@@ -716,30 +735,139 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+	return &memberInputExpr{ma: ma, raw: p.input[cp.pos:p.pos]}, true, nil
+}
+
+// parseAsteriskInputExpr parses an INSERT statement input expression where
+// SQLair generates the columns.
+// It is of the form "(*) VALUES ($Type.*, $Type.member,...)".
+func (p *Parser) parseAsteriskInputExpr() (*asteriskInputExpr, bool, error) {
+	cp := p.save()
+
+	parenCol := p.pos + 1
```
letFunny: This will not work because the column could be on a separate line. You have to do cp.colNum() or p.colNum() and increment it when you are sure there was a parentheses.

internal/expr/parser.go
```diff
@@ -716,30 +735,139 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+	if !p.skipString("(*)") {
```
letFunny: I think it would be good to allow `( * )` with whitespace, it seems like an artificial limitation.

internal/expr/parser.go
```diff
@@ -716,30 +735,139 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+	p.skipBlanks()
+
+	parenCol := p.colNum()
+	// Ignore errors and leave them to be handled by
+	// parseMemberInputExpr later.
+	sources, ok, _ := parseList(p, (*Parser).parseInputMemberAccessor)
+	if !ok {
+		// Check for types with missing parentheses.
+		if _, ok, _ := p.parseTypeAndMember(); ok {
+			return nil, false, errorAt(fmt.Errorf(`missing parentheses around types after "VALUES"`), p.lineNum, parenCol, p.input)
```
letFunny: Good idea! Do we want to `cp.restore()` here, I think that we need some general advice, like do not restore if you are going to emit an error but do in all other cases.

letFunny (commented):
Did a first pass, thanks for the PR. I have some open questions and some comments, will do another thorough review next revision.

internal/expr/bindtypes.go
Aflynn50: Changed, I used "occurring within an INSERT..." since we don't actually parse the "INSERT" so the expressions doesn't really consist of it.

internal/expr/bindtypes.go
Aflynn50: Changed, also corrected the comment, it still said we could only have one type which is now not true!

internal/expr/expr_test.go
Aflynn50: It would make generating the SQL a lot more confusing, it would no longer be linear since the `Bypass[VALUES ]` would come in the middle of `AsteriskInput`. We could change the way things are printed for the parser tests so that the `VALUES` is actually printed, i.e. `[Bypass[INSERT INTO person ] AsteriskInput[[*] VALUES [Address.street Person.* M.team]]]`

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: The policy is to restore every-time, even when emitting an error, just cuz the error could be ignored by the caller.

internal/expr/expr_test.go
letFunny: Yeah I like your suggestion a lot more. Either that or keep it like it is implemented now, the VALUES might be clearer when reading tests as you point out.

Aflynn50: I'll just leave it as it is for now, since the outputs have never put in the `AS`.

internal/expr/parser.go
Aflynn50: I've updated this function, hopefully its a bit clearer now.

internal/expr/parser.go
```diff
@@ -281,13 +281,27 @@ loop:
...
+			if p.pos >= len(p.input) || isInitialNameByte(p.input[p.pos]) {
+				break loop
+			}
+			continue
 		}
+
 		p.advanceByte()
+		if p.pos == len(p.input) {
+			break loop
+		}
```
letFunny: [Nit]: This is not needed right? The condition in for loop is already checking this.

letFunny (commented):
Need to do a full pass, only a nit

internal/expr/expr_test.go
```diff
@@ -626,10 +626,10 @@ of three lines' AND id = $Person.*`,
 		err:   `cannot parse expression: column 8: cannot read function call "count(*)" into asterisk`,
 	}, {
 		query: "INSERT INTO person (*) VALUES $Address.*",
-		err:   `cannot parse expression: column 20: missing parentheses around types after "VALUES"`,
+		err:   `cannot parse expression: column 21: missing parentheses around types after "VALUES"`,
```
letFunny: The column now is pointing to the asterisk, shouldn't we point to the start of the expression or even better to the missing parentheses?
```
INSERT INTO person (*) VALUES $Address.*
                    ^
```

internal/expr/parser.go
```diff
@@ -64,6 +63,8 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
 			p.add(in)
 			continue
 		}
+
+		p.advanceByte()
```
letFunny, line 69: What do you think of adding a comment like:
```
// We have not found an expression, we will start searching again in the next character. This prevents `advanceToNextExpression` to enter an infinite loop.
p.advanceByte()
```

internal/expr/parser.go
```diff
@@ -40,8 +40,7 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
 	p.init(input)
 
 	for {
-		// Advance the parser to the start of the next expression.
-		if err := p.advance(); err != nil {
+		if err := p.advanceToNextExpression(); err != nil {
```
letFunny, line 43: I think we should probably have tests for advance. If not full tests, one that checks that it goes on an infinite loop. So if we change it in the future we know what is failing.

internal/expr/parser.go
```diff
@@ -280,19 +280,24 @@ loop:
...
+		switch p.input[p.pos] {
+		// These bytes may be the start of an expression.
+		case '(', '*', '$', '&':
 			break loop
+		// An expression can also start with an initial name byte, e.g. a
+		// expression starting with a column name or a SQL function.
+		// Rather than testing for every initial name byte (we would stop at
+		// every letter of every word), we look for bytes that may precede the
+		// start of an expression and then check if the next byte is an initial
+		// name byte.
```
letFunny, line 299: Great!!!! Finally

letFunny (commented):
Did a pass on the new changes. Thanks, it is looking very good!

internal/expr/expr_test.go
Aflynn50: Ah yea, put the position save in the wrong place.

internal/expr/parser.go
Aflynn50: Sure, added.

internal/expr/parser.go
Aflynn50: Added some tests, I've not done a super comprehensive job, I'm just thinking that bits can be added to the table of tests when you encounter something weird, as often happens.

internal/expr/parser.go
```diff
@@ -779,11 +781,11 @@ func (p *Parser) parseMemberInputExpr() (*memberInputExpr, bool, error) {
 func (p *Parser) parseAsteriskInputExpr() (*asteriskInputExpr, bool, error) {
 	cp := p.save()
 
+	parenCol := p.colNum()
+	parenLine := p.lineNum
```
letFunny: I think this is redundant now and you could do:
```
cp.lineNum
cp.colNum()
```
without having to store it explicitly.

internal/expr/parser.go
letFunny: No worries that was expected. What about adding a test for staying in the same position? That is an invariant of the function that if we break I would want to see a test failing.
```go
}, {
// advance... does not consume characters if the current one could be the start of an expression.
		input:   `col1`,
		stopPos: []int{0},
}
```

Aflynn50: Cool, added the cases for it stopping at the start of the input if its possibly an expression.

internal/expr/parser.go
Aflynn50: Good point

internal/expr/expr_test.go
```diff
@@ -449,6 +449,64 @@ func (s *ExprSuite) TestExprPkg(c *C) {
...
+	}, {
+		summary:        "insert with standalone input expressions",
+		query:          `INSERT INTO person VALUES ($Person.name, "random string", $Person.id)`,
+		expectedParsed: `[Bypass[INSERT INTO person VALUES (] Input[Person.name] Bypass[, "random string", ] Input[Person.id] Bypass[)]]`,
+	}}
+	for i, t := range tests {
+		parser := expr.NewParser()
+		expr, err := parser.Parse(t.query)
+		c.Assert(err, IsNil, Commentf("test %d failed (Parse):\nsummary: %s\nquery: %s\nexpected: %s\nerr: %s\n", i, t.summary, t.query, t.expectedParsed, err))
+		c.Assert(expr.String(), Equals, t.expectedParsed, Commentf("test %d failed (Parse):\nsummary: %s\nquery: %s\nexpected: %s\nactual:   %s\n", i, t.summary, t.query, t.expectedParsed, expr.String()))
```
letFunny, line 506: [Not even Nit]: Lately I have seen tests in Chisel using `c.Log(test.summary)` and keeping the asserts bare, that seems to work because the logs are only printed when the test fails. Seems like an interesting pattern.

internal/expr/expr_test.go
```diff
@@ -519,7 +577,7 @@ comment */ WHERE x = $Address.&d`,
 		err:   `cannot parse expression: column 22: missing closing parentheses`,
 	}, {
 		query: "SELECT (name, id) WHERE id = $Person.*",
-		err:   `cannot parse expression: column 30: asterisk not allowed in input expression "$Person.*"`,
+		err:   `cannot parse expression: column 30: invalid asterisk input placement "$Person.*"`,
```
letFunny: Now that I re-read it I still love it but I am not sure whether `asterisk input` is a well defined thing or whether we should go for `invalid asterisk placement` or `invalid asterisk placement in input`. Thoughts?

internal/expr/parser.go
```diff
@@ -281,11 +283,21 @@ loop:
...
 			break loop
+		// An expression can also start with an initial name byte, e.g. a
+		// expression starting with a column name or a SQL function.
+		// Rather than testing for every initial name byte (we would stop at
+		// every letter of every word), we look for bytes that may precede the
+		// start of an expression and then check if the next byte is an initial
+		// name byte.
+		case ' ', '\t', '\n', '\r', '=', ',', '[', '>', '<', '+', '-', '/', '|', '%':
+			p.advanceByte()
+			if p.pos >= len(p.input) || isInitialNameByte(p.input[p.pos]) {
```
letFunny: I think that for clarity and for maintenance `p.pos >= len(p.input)` should be a separate clause with an early return. I am afraid that somebody adds code after the loop and it breaks only in the specific instance where the SQL ends with one of these characters. I mean:
```go
if p.pos >= len(p.input) {
     return nil
}
if isInitial... {
    break loop
}

File: internal/expr/parser.go [hunk 21]
```diff
@@ -624,7 +639,9 @@ func parseList[T any](p *Parser, parseFn func(p *Parser) (T, bool, error)) ([]T,
 
 		nextItem = p.skipByte(',')
 	}
-	return nil, false, errorAt(fmt.Errorf("missing closing parentheses"), openingParenLine, openingParenCol, p.input)
+	err := errorAt(fmt.Errorf("missing closing parentheses"), openingParenLine, openingParenCol, p.input)
```
letFunny: I see that openingParenLine and openingParenCol are redundant given that we have `cp`. We could remove them in this PR :P

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+// parseSliceInputExpr parses an input expression of the form "$Type[:]".
+func (p *Parser) parseSliceInputExpr() (*sliceInputExpr, bool, error) {
 	cp := p.save()
 	if !p.skipByte('$') {
 		return nil, false, nil
 	}
 
-	// Case 1: Slice range, "Type[:]".
 	if st, ok, err := p.parseSliceAccessor(); err != nil {
+		cp.restore()
```
letFunny, line 779: [Idea]: What do you think of restoring the `cp` using a defer clause checking the `err` value? It is more implicit but this way we don't have to remember to write it everywhere.

I am not saying to do it in this PR just asking for future changes.

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
-			return nil, false, errorAt(fmt.Errorf("asterisk not allowed in input expression %q", "$"+ma.String()), cp.lineNum, cp.colNum(), p.input)
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseMemberInputExpr parses an input expression of the form "$Type.member".
+func (p *Parser) parseMemberInputExpr() (*memberInputExpr, bool, error) {
+	cp := p.save()
+	ma, ok, err := p.parseInputMemberAccessor()
+	if !ok {
```
letFunny, line 793: This is relying on the fact that `ok = false` when `err != nil`. I think in Go the convention is to check the `err` value, then the `ok` value. There are some functions in the standard library that do not return the zero value when returning an error for example.

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+	if !p.skipByte('*') {
+		cp.restore()
+		return nil, false, nil
+	}
+	p.skipBlanks()
+	if !p.skipByte(')') {
+		cp.restore()
+		return nil, false, nil
+	}
+
```
letFunny: I would remove this newline, the code after it is part of the same parsing.

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
 		return nil, false, err
 	}
 
-	cp.restore()
+	return &asteriskInputExpr{sources: sources, raw: p.input[cp.pos:p.pos]}, true, nil
+}
+
+// parseColumnsInputExpr parses an INSERT statement input expression where the
+// user specifies the columns to insert from a single type.
+// It is of the form "(col1, col2,...) VALUES ($Type.*)".
```
letFunny, line 847: Should we say `(tag1, tag2, ...) VALUES ($Type.*)`? I feel like in some meetings the naming of columns and tags has lead to some misunderstandings. In that regard, `tag` is more explicit.

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+func (p *Parser) parseColumnsInputExpr() (*columnsInputExpr, bool, error) {
+	cp := p.save()
+
+	columns, paren, ok := p.parseColumns()
+	if !(ok && paren) {
+		cp.restore()
+		return nil, false, nil
+	}
+
```
letFunny: I would remove this newline (same reasons as above).

internal/expr/parser.go
```diff
@@ -716,30 +742,150 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+		if _, ok, _ := p.parseTypeAndMember(); ok {
+			cp.restore()
+			return nil, false, errorAt(fmt.Errorf(`missing parentheses around types after "VALUES"`), parenLine, parenCol, p.input)
+		}
+		cp.restore()
+		return nil, false, nil
+	}
+	if starCountTypes(sources) == 0 {
+		// If there are no asterisk accessors leave the accessors to be handled
+		// by parseMemberInputExpr later.
```
letFunny: The problem with this comment is that it is documenting this function in relationship to another one in an order that is imposed higher in the abstraction. I think we should stick to saying that this function should not handle this case.

internal/expr/parser_helper_test.go
```diff
@@ -243,3 +243,65 @@ func (s parseSuite) TestParseSliceRange(c *C) {
...
+
+	for _, t := range tests {
+		p.init(t.input)
+		currentStopPos := 0
+		for p.pos < len(p.input) {
+			err := p.advanceToNextExpression()
+			c.Assert(err, IsNil)
+			if p.pos >= len(p.input) {
+				break
+			}
```
letFunny: Maybe I am misunderstanding it but I think this line is somewhat bypassing the tests. If I changed the implementation so that `&col` did not stop I think the tests will still succeed. I think we need to assert that `currentStopPos == len(t.stopPos)` here.

internal/expr/parser_helper_test.go
```diff
@@ -243,3 +243,65 @@ func (s parseSuite) TestParseSliceRange(c *C) {
...
+
+			if len(t.stopPos) <= currentStopPos {
```
letFunny: [Nit]: I think the convention is to write `currentStopPos >= len(t.stopPos)`, I would assume most people are used to reading if from left to right.

letFunny (commented):
The PR is looking really good, congratulations! As it usually happens, the better the PR, the more nitpicks and comments about naming stuff :).

internal/expr/expr_test.go
Aflynn50: That's seems like a better pattern, we should change it to do that at some point, though I won't do it in this PR

internal/expr/expr_test.go
Aflynn50: Changed to the second of the two suggestions, that seems better.

internal/expr/parser.go
Aflynn50: Fair enough, changed.

internal/expr/parser.go

internal/expr/parser.go
Aflynn50: Yea, that would be better generally I think, there are a lot of places where it is repeated a ton.

internal/expr/parser.go
Aflynn50: Right, but here we don't actually check the err value, if `ok=true` then its one of our invariant that `err=nil`. If we did check the err here then we would just be doing `if err != nil {return err}; if !ok {return nil} which has the same effect.

Can change it though if it makes you uncomfortable.

internal/expr/parser.go
Aflynn50: But the type can be a map, so they're not always tags.

internal/expr/parser.go
Aflynn50: Changed the comment to reflect that.

internal/expr/parser_helper_test.go
Aflynn50: Good point, added in the assertion.

internal/expr/parser.go
letFunny: Right, the term for maps would be keys.

internal/expr/bindtypes.go
```diff
@@ -115,7 +115,43 @@ func (e *memberInputExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
...
 	}
-	return &typedInputExpr{input}, nil
+	return &typedInputExpr{input: input}, nil
+}
+
+// asteriskInputExpr is an input expression occurring within an INSERT
+// statement that consists of an asterisk on the left and explicit type accessors
+// on the right. This means that SQLair generates the columns.
+// e.g. "(*) VALUES ($Type1.col1, $Type2.*)".
+type asteriskInputExpr struct {
```
manadart: Are these better named to expressly indicate part of an insert? `asteriskInsertExpr` for example.

internal/expr/parser.go
```diff
@@ -716,30 +748,148 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+		cp.restore()
+		return nil, false, errorAt(fmt.Errorf("invalid asterisk placement in input %q", "$"+ma.String()), cp.lineNum, cp.colNum(), p.input)
+	}
+	return &memberInputExpr{ma: ma, raw: p.input[cp.pos:p.pos]}, true, nil
+}
+
+// parseAsteriskInputExpr parses an INSERT statement input expression where
+// SQLair generates the columns.
+// It is of the form "(*) VALUES ($Type.*, $Type.member,...)".
+func (p *Parser) parseAsteriskInputExpr() (*asteriskInputExpr, bool, error) {
```
manadart: Which would rename these too.

internal/expr/parser.go
```diff
@@ -716,30 +748,148 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+		cp.restore()
+		return nil, false, nil
+	}
+
+	return &columnsInputExpr{columns: columns, sources: sources, raw: p.input[cp.pos:p.pos]}, true, nil
+}
+
+// parseInputExpr parses all forms of input expressions, that is, expressions
+// containing a "$".
+func (p *Parser) parseInputExpr() (expression, bool, error) {
```
manadart: I would tend to locate this above the methods it calls.

manadart: You could also do a loop over over the methods as first class funcs and shorten this.

manadart (approved):
Looks good; just semantic suggestions.

internal/expr/parser.go
letFunny: What happens if there is an error in the parse function, we are checking `ok` and that might be set because the invariants only mandate `err != nil`. That is what I meant. We would conitnue with the function as normal instead of returning.

internal/expr/parser.go
```diff
@@ -716,30 +748,148 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
...
+		if _, ok, _ := p.parseTypeAndMember(); ok {
+			cp.restore()
+			return nil, false, errorAt(fmt.Errorf(`missing parentheses around types after "VALUES"`), parenLine, parenCol, p.input)
+		}
+		cp.restore()
+		return nil, false, nil
+	}
+	if starCountTypes(sources) == 0 {
+		// If there are no asterisk accessors leave the accessors then leave
+		// the accessors to be handled elsewhere.
```
letFunny: Typo?

letFunny (approved):
Thanks!

internal/expr/parser.go
Aflynn50: Moved and changed.

internal/expr/parser.go
Aflynn50: Renamed.

internal/expr/bindtypes.go
Aflynn50: Renamed.

internal/expr/parser.go
Aflynn50: Fixed.

PR #132, opened 2024-01-12
Add bindTypes method to expression

Aflynn50:
This PR adds a `bindTypes` method to expression. The functions `bindInputExpr` and `bindOutputExpr` are refactored into this new method. This PR also separates the `inputExpr` into two separate expressions, `memberInputExpr` and `sliceInputExpr`.

The `valueAccessor` interface is removed as it is no longer a useful abstraction with the addition of the separate input expression types above.

This PR is motivated by the upcoming support for input expressions in INSERT statements which will require a new kind in input expression. Rather than overload the current `inputExpr` it makes sense to split it into three types. `insertInputExpr`, `sliceInputExpr` and `memberInputExpr`. These types can then each have their own `bindTypes` function which will handle the type samples from the user appropriately depending on the type of expression.

Going forward, it would also be natural to split the `outputExpr` into three separate types of expression. Currently `bindOutputTypes` has three separate cases (defined in comments rather than by separate methods) for the different forms output expressions. By codifying these as three different types it would make the code far more maintainable.

### DIFF
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -49,31 +49,19 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 	// Bind types to each expression.
 	var typedExprs TypeBoundExpr
 	outputUsed := map[string]bool{}
-	var te any
 	for _, expr := range pe.exprs {
-		switch e := expr.(type) {
-		case *inputExpr:
-			te, err = bindInputTypes(e, argInfo)
-			if err != nil {
-				return nil, err
-			}
-		case *outputExpr:
-			toe, err := bindOutputTypes(e, argInfo)
-			if err != nil {
-				return nil, err
-			}
+		te, err := expr.bindTypes(argInfo)
+		if err != nil {
+			return nil, err
+		}
 
+		if toe, ok := te.(*typedOutputExpr); ok {
 			for _, oc := range toe.outputColumns {
 				if ok := outputUsed[oc.output.Identifier()]; ok {
 					return nil, fmt.Errorf("%s appears more than once in output expressions", oc.output.Desc())
 				}
 				outputUsed[oc.output.Identifier()] = true
 			}
-			te = toe
-		case *bypass:
-			te = e
-		default:
-			return nil, fmt.Errorf("internal error: unknown query expr type %T", expr)
 		}
 		typedExprs = append(typedExprs, te)
 	}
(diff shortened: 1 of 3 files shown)


internal/expr/parser.go
```diff
@@ -525,26 +475,26 @@ func (p *Parser) parseTargetType() (memberAccessor, bool, error) {
 }
 
 // parseSliceAccessor parses a slice range composed of the form "[:]".
-func (p *Parser) parseSliceAccessor() (sa sliceAccessor, ok bool, err error) {
+func (p *Parser) parseSliceAccessor() (string, bool, error) {
```
letFunny: Can we make this return `sliceAccessor`? I think it makes the method clearer, you dont have to state that the string you return is the type name which is then to be wrapped by `sliceAccessor`

internal/expr/bindtypes.go
```diff
@@ -49,64 +49,93 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
...
+	// bindTypes binds the types to the expression to generate either a
+	// *typedInputExpr or *typedOutputExpr.
+	bindTypes(typeinfo.ArgInfo) (any, error)
+}
+
+// memberInputExpr is an input expression of the form "$Type.member" which
+// represents a query parameter contained in a member of a type.
+type memberInputExpr struct {
+	raw string
+	ma  memberAccessor
```
letFunny, line 103: Maybe we can go one step further because it would be nice to flatten this and remove the `memberInputExpr` that is only a wrapper. Could we implement  `bindTypes` on `memberAccessor` directly?

The challenging bit would be the separation between input and output expressions and making that clear.

internal/expr/bindtypes.go
```diff
@@ -49,64 +49,93 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
...
+	raw           string
+	sliceTypeName string
+}
+
+// String returns a text representation for debugging and testing purposes.
+func (e *sliceInputExpr) String() string {
+	return fmt.Sprintf("Input[%s[:]]", e.sliceTypeName)
+}
+
+// bindTypes generates a *typedInputExpr containing type information about the
```
letFunny, line 133: In my opinion, all of these comments about the output type should be in the types themselves.

letFunny (commented):
Looking very good! I left some comments. Some are not actionable, I just want to have the discussion onto why things were done this way.

internal/expr/parser.go
Aflynn50: `sliceAccessor` is removed in this PR. I could  change this to return a `sliceInputExpr` instead since that directly contains the slice type name as a string. However, to do this I think it would also make sense to parse the `$` (or the function name would not make sense). Doing this would mean that we couldn't directly use this in the `parseOutputExpr` to error check.

internal/expr/bindtypes.go
Aflynn50: We could, but I think this would get a bit confusing when the INSERT expressions are added. In their current implementation the insert input expression type has a slice of `memberAccessor`. We could add `memberName` and `typeName` directly to `memberInputExpr` rather than having the `memberAccessor` as a field but I think when the INSERT PR is added this will make more sense.

Of course if we change how it is implemented in the INSERT PR then it makes sense to change it here.

internal/expr/bindtypes.go
Aflynn50: In my opinion they should really be implicit in the type, but since we need to implement the interface that has an `any` its necessary to specify it. I would rather not put them in the function because they define the result of the function, not how it works.

It would make sense to remove the output type from the function name if we replace the `any` in the `bindTypes` function in the interface with an interface such as `typedExpr`. I did try doing that but it was hard to make a compelling case to myself that it added anything more than it did in PR #127 when we decided not to add it.

letFunny: I was advocating for your second paragraph. If you tried it and it does not fit then let's leave it like in this PR.

internal/expr/bindtypes.go
letFunny: I see, then we can go ahead with this PR and then discuss it further in the INSERT one. But taking into account what you just mentioned I think the wrapper probably unavoidable.

internal/expr/parser.go
Aflynn50: I will try and define what a `sliceAccessor` is better in the comment above (i.e. `Slice[:]` without a `$` or `&`), but for the reasons above, I think the function name and content makes more sense as it is.

internal/expr/bindtypes.go
Aflynn50: Yea, I felt that Joe's comment from before still stands so there is no justification to change it.

internal/expr/parser.go
letFunny: EDIT: the previous message did not make sense, I read yours wrong. Let's continue the discussion below.

---

Basically what you are saying is that it does not make sense to return an input expression and I agree. I also dont like the fact that we are returning a string instead of something more descriptive (like we do in other functions where we return *Accessor).

So what about changing the function name and adding typeName as the name of the returned string?

Aflynn50: Have changed it to that.

internal/expr/bindtypes.go
```diff
@@ -49,69 +49,97 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
...
+	raw           string
+}
+
+// String returns a text representation for debugging and testing purposes.
+func (e *outputExpr) String() string {
+	return fmt.Sprintf("Output[%+v %+v]", e.sourceColumns, e.targetTypes)
+}
+
+// bindTypes binds the output expression to concrete types. It then checks the
+// expression valid with respect to its bound types and returns a
```
manadart: I presume you meant to type "expression is valid".

internal/expr/bindtypes.go
```diff
@@ -49,69 +49,97 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
...
-			toe, err := bindOutputTypes(e, argInfo)
-			if err != nil {
-				return nil, err
-			}
+		te, err := expr.bindTypes(argInfo)
+		if err != nil {
+			return nil, err
+		}
 
+		if toe, ok := te.(*typedOutputExpr); ok {
```
manadart, line 58: This is OK for now, but we should keep an eye out for a way to poke it into the abstraction you've created.

internal/expr/parser.go
```diff
@@ -159,44 +131,22 @@ func (sfc sqlFunctionCall) String() string {
...
-
-func (p *outputExpr) String() string {
-	return fmt.Sprintf("Output[%+v %+v]", p.sourceColumns, p.targetTypes)
-}
-
-func (p *outputExpr) expr() {}
-
 // bypass represents part of the expression that we want to pass to the backend
 // database verbatim.
 type bypass struct {
```
manadart: You relocated the other types, but left this one in parser. Is that deliberate?

manadart (approved):
Couple of nits, but this looks good. We can arrange to land it on Monday.

internal/expr/parser.go
Aflynn50: On reflect it should be with the others. Moved.

internal/expr/bindtypes.go
Aflynn50: Agreed, this chunk sticks out like a sore thumb.

PR #133, opened 2024-01-15
bugfix: ignore SQL function calls in output expressions 

letFunny:
The parser now works when using SQL functions instead of column names in
output expressions.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -122,27 +122,25 @@ type columnAccessor interface {
 	columnName() string
 }
 
-// staticColumn stores a SQL column name and optionally its table name.
-type staticColumn struct {
+// basicColumn stores a SQL column name and optionally its table name.
+type basicColumn struct {
 	table, column string
 }
 
-func (sc staticColumn) columnName() string {
+func (sc basicColumn) columnName() string {
 	return sc.column
 }
 
-func (sc staticColumn) tableName() string {
+func (sc basicColumn) tableName() string {
 	return sc.table
 }
 
-func (sc staticColumn) String() string {
+func (sc basicColumn) String() string {
 	if sc.table == "" {
 		return sc.column
 	}
 	return sc.table + "." + sc.column
 }
-
-var _ columnAccessor = staticColumn{}
 
 // sqlFunctionCall stores a function call that is used in place of a column.
 type sqlFunctionCall struct {
@@ -160,8 +158,6 @@ func (sfc sqlFunctionCall) tableName() string {
 func (sfc sqlFunctionCall) String() string {
 	return sfc.raw
 }
-
-var _ columnAccessor = sqlFunctionCall{}
 
 // inputExpr represents a named parameter that will be sent to the database
 // while performing the query.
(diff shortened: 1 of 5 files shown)


internal/expr/parser.go
```diff
@@ -677,6 +764,13 @@ func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
 				if !parenCols && parenTypes {
 					return nil, false, errorAt(fmt.Errorf(`unexpected parentheses around types after "AS"`), p.lineNum, parenCol, p.input)
 				}
+				if starCountTypes(targetTypes) > 0 {
```
Aflynn50, line 767: This block of code is pretty messy right now, but if we are going to switch to separate expressions (and separate parsing functions for them) in the future this will go away.

internal/expr/parser.go
```diff
@@ -116,18 +116,53 @@ func (sa sliceAccessor) getTypeName() string {
...
-type columnAccessor struct {
-	tableName, columnName string
+type columnAccessor interface {
+	String() string
+	tableName() string
+	columnName() string
+}
+
+// staticColumn stores a SQL column name and optionally its table name.
+type staticColumn struct {
```
Aflynn50: Not sure about the name `staticColumn`, what is static about it? Imo something like `simpleColumn` or `basicColumn` would be better.

internal/expr/parser.go
```diff
@@ -116,18 +116,53 @@ func (sa sliceAccessor) getTypeName() string {
...
+
+func (sfc sqlFunctionCall) tableName() string {
+	return ""
+}
+
+func (sfc sqlFunctionCall) String() string {
+	return sfc.raw
 }
 
+var _ columnAccessor = sqlFunctionCall{}
```
Aflynn50: At some point Joe advised that these should go in the tests rather than the main code, this can go in `parser_helper_test.go`

internal/expr/parser.go
```diff
@@ -486,24 +521,76 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+			parenCount++
+			continue
+		} else if p.skipByte(')') {
+			parenCount--
+			continue
+		}
+		p.advanceByte()
+	}
+
+	if parenCount > 0 {
```
Aflynn50, line 519: We restore the state with a checkpoint before returning an error in `skipStringLiteral`. We should do it here as well to keep it consistent. It can be useful for error messages in function calling this one.

internal/expr/parser.go
```diff
@@ -486,24 +521,76 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
-// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// skipEnclosedParentheses starts from a opening parenthesis '(' and skips until
+// its closing ')', taking into account comments and string literals in between.
+func (p *Parser) skipEnclosedParentheses() (bool, error) {
```
Aflynn50, line 490: This function should be grouped with the other `skip` functions further up the file.

internal/expr/parser.go
```diff
@@ -486,24 +521,76 @@ func (p *Parser) parseIdentifier() (string, bool) {
 	return "", false
 }
 
-// parseColumn parses a column made up of name bytes, optionally dot-prefixed by
+// skipEnclosedParentheses starts from a opening parenthesis '(' and skips until
+// its closing ')', taking into account comments and string literals in between.
```
Aflynn50: Maybe point out that it also takes into account nested parenthesis inside the main pair.

internal/expr/parser.go
```diff
@@ -486,24 +521,76 @@ func (p *Parser) parseIdentifier() (string, bool) {
...
+		p.advanceByte()
+	}
+
+	if parenCount > 0 {
+		return false, errorAt(fmt.Errorf(`missing closing parenthesis`), initialLine, initialCol, p.input)
+	}
+	return true, nil
+}
+
+// parseColumnAccessor parses a column made up of name bytes, optionally dot-prefixed by
```
Aflynn50: This comment should be updated.

Aflynn50 (commented):
Looks very good, only minor comments.

internal/expr/parser.go
letFunny: I tried to convey static as in a regular column name, I agree with you, basicColumn is better.

internal/expr/parser.go
letFunny: Moved.

internal/expr/parser.go
letFunny: Done, both `cp.restore` added when returning.

internal/expr/parser.go
letFunny: Yep, it is partially because it is doing error handling and validation. We could move it out by I think it will have the same messy-ness / complexity.

Aflynn50 (approved):
Looking good, might be good to have a black box test in package_test.go to check that the functions behave as expected, but basically ready to go.

manadart (approved):
Nice job.

PR #134, opened 2024-01-16
Move input and output argument validation to typeinfo package

Aflynn50:
This PR moves the sections of `BindInputs` and `ScanArgs` that validate the arguments from the user to the `typeinfo` package. The new functions return the `typeToValue` map that is used by `BindInputs` and `ScanArgs`.

The error messages returned by the two validation functions are made consistent and the two functions are changed to match in shape even though they are subtly different.

A test is also added for nil maps as arguments, since this was not previously checked.

### DIFF
--- a/internal/typeinfo/validate.go
+++ b/internal/typeinfo/validate.go
@@ -1,0 +1,64 @@
+package typeinfo
+
+import (
+	"fmt"
+	"reflect"
+)
+
+func ValidateInputs(args []any) (map[reflect.Type]reflect.Value, error) {
+	typeToValue := map[reflect.Type]reflect.Value{}
+	for _, arg := range args {
+		v := reflect.ValueOf(arg)
+		if isNil(v) {
+			return nil, fmt.Errorf("need struct or map, got nil")
+		}
+		v = reflect.Indirect(v)
+		k := v.Kind()
+		if k != reflect.Struct && k != reflect.Map {
+			return nil, fmt.Errorf("need struct or map, got %s", k)
+		}
+		t := v.Type()
+		if _, ok := typeToValue[t]; ok {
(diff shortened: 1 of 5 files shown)


internal/typeinfo/validate.go
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under Apache 2.0, see LICENCE file for details.
+
+package typeinfo
+
+import (
+	"fmt"
+	"reflect"
+)
+
+func ValidateInputs(args []any) (map[reflect.Type]reflect.Value, error) {
```
manadart: Comments on these methods please.

I'd consider a type alias for `map[reflect.Type]reflect.Value`

manadart (commented):
I really like this change.

internal/typeinfo/validate.go
Aflynn50: Added alias and comments.

internal/typeinfo/validate.go
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under Apache 2.0, see LICENCE file for details.
+
+package typeinfo
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type TypeToValue map[reflect.Type]reflect.Value
```
manadart: This is a type definition, rather than a type alias.
An alias would be `type TypeToValue = map[reflect.Type]reflect.Value` (with an equals sign).

I tend to prefer using an alias if you don't need to further extend the type because:
- You can assign one to the other directly without casting.
- You can compare them.

Aflynn50: Ah yea, I'm so used to only using type definitions because SQLair doesn't work with aliases! Changed.

manadart (approved):

PR #135, opened 2024-01-23
Basic usage examples

Aflynn50:
This PR adds example for `Get`, `GetAll` and `Iter`. These are the three methods on `Query` that can be used to run queries on the database.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,211 @@
+package sqlair_test
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/canonical/sqlair"
+)
+
+func ExampleQuery_Get() {
+	type Employee struct {
+		ID   int    `db:"employee_id"`
+		Name string `db:"name"`
+		Team string `db:"team_name"`
+	}
+
+	db, err := employeeDB()
+	if err != nil {
+		return
+	}
+
+	stmt, err := sqlair.Prepare("SELECT &Employee.* FROM employees", Employee{})
+	if err != nil {
+		return
+	}
+
+	var e Employee
+	err = db.Query(context.Background(), stmt).Get(&e)
+	if err != nil {
+		return
+	}
+
+	fmt.Printf("Employee: %+v", e)
+
+	// Output: Employee: {ID:1 Name:Alastair Team:Juju}
+}
+
+func ExampleQuery_Get_withInput() {
+	type Employee struct {
+		ID   int    `db:"employee_id"`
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -0,0 +1,170 @@
...
+	err = iter.Close()
+	if err != nil {
+		return
+	}
+
+	fmt.Printf("Employees: %+v", es)
+	// Output: Employees: [{ID:1 Name:Alastair Team:Engineering} {ID:2 Name:Alberto Team:Engineering}]
+}
+
+func employeeDB() (*sql.DB, error) {
```
letFunny: Can we simplify this to return a `*sqlair.DB`?

example_test.go
```diff
@@ -0,0 +1,170 @@
...
+	// Output: Employee: {ID:1 Name:Alastair Team:Engineering}
+}
+
+func ExampleQuery_Get_withInput() {
+	type Employee struct {
+		ID   int    `db:"employee_id"`
+		Name string `db:"name"`
+		Team string `db:"team_name"`
+	}
+
```
letFunny, line 45: [Nit] Remove newline.

example_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/canonical/sqlair"
+)
+
+func ExampleQuery_Get() {
+	type Employee struct {
+		ID   int    `db:"employee_id"`
+		Name string `db:"name"`
+		Team string `db:"team_name"`
+	}
```
letFunny, line 16: What do you think of moving the employee type outside so that we can reuse across the examples?

example_test.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = db.Exec(`CREATE TABLE employees (employee_id integer, team_name string, name text);`)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = db.Exec(`INSERT INTO employees (employee_id, name, team_name) VALUES (?, ?, ?);`, 1, "Alastair", "Engineering")
```
letFunny: We have to remember to change this to the new syntax once it is in.

letFunny (approved):
Looking great, only a few nitpicks. We can iterate once it is landed to keep improving the examples and adding new syntax (e.g. inserts).

example_test.go
Aflynn50: These are the examples that show up on the website. If you put it outside it leads to it being missed out from the examples in some cases.

PR #136, opened 2024-01-31
Add support for nil slices

Aflynn50:
The decision was made to allow the empty slice as a valid slice input in SQLair. When the slice is empty no input placeholders are generated for the slice (e.g. `... IN ($S[:])` => `... IN ()`). 

In Go, the nil slice is often used interchangeably with the empty slice (since append works on both) however, when a nil slice is passed to SQLair, it returns the error `"need supported value, got nil"`.

The behavior for `nil` slices should match that for empty slices to give a consistent user experience.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -401,7 +401,8 @@ AND z = @sqlair_0 -- The line with $Person.id on it
 	typeSamples:    []any{sqlair.S{}},
 	inputArgs:      []any{(sqlair.S)(nil)},
 	expectedParams: []any{},
-	expectedSQL:    "SELECT name FROM person WHERE id IN ()",
+	// This is valid in SQLite (though not in MySQL).
+	expectedSQL: "SELECT name FROM person WHERE id IN ()",
 }}
 
 func (s *ExprSuite) TestExprPkg(c *C) {
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -803,17 +803,17 @@ func (s *ExprSuite) TestBindInputsError(c *C) {
 		query:       "SELECT street FROM t WHERE x = $Address.street, y = $Person.name",
 		typeSamples: []any{Address{}, Person{}},
 		inputArgs:   []any{nil, Person{Fullname: "Monty Bingles"}},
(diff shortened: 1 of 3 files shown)


internal/expr/expr_test.go
```diff
@@ -390,6 +390,17 @@ AND z = @sqlair_0 -- The line with $Person.id on it
 	typeSamples:    []any{sqlair.S{}},
 	inputArgs:      []any{sqlair.S{}},
 	expectedParams: []any{},
+	// This is valid in SQLite (though not in MySQL).
```
letFunny, line 393: Do you need to also add this comment to the new testcase?

internal/typeinfo/validate.go
```diff
@@ -72,7 +72,7 @@ func isNil(v reflect.Value) bool {
 	switch v.Kind() {
 	case reflect.Invalid:
 		return true
-	case reflect.Pointer, reflect.Map, reflect.Slice:
+	case reflect.Pointer, reflect.Map:
```
letFunny: It is a little bit weird now that a function that checks whether an element is nil does not care about slices. In my opinion we should change the function name or we should add a explicit check for slices.

letFunny (commented):
Nice change! Left some comments about naming as usual

letFunny (approved):

internal/typeinfo/validate.go
Aflynn50: Changed name to `isInvalidNil`

internal/typeinfo/validate.go
```diff
@@ -45,7 +45,7 @@ func ValidateOutputs(args []any) (TypeToValue, error) {
 	typeToValue := TypeToValue{}
 	for _, arg := range args {
 		v := reflect.ValueOf(arg)
-		if isNil(v) {
+		if isInvalidNil(v) {
```
manadart: Instead of keeping the boolean signature, make it return an error. That way, you can return a message based on the case you detect, then just return it if it's not `nil`. Call it `ValidateValue`.

Aflynn50: Changed type of function and made error messages more useful.

manadart (approved):
Nice; thanks.

PR #137, opened 2024-02-05
simplify parser

letFunny:
This commits helps reason about the parser and in adding recursive parsing latter. The main idea is the separation of concern between expression parsing, adding bypass parts and the advance() function; all three were interleaved before.

### DIFF
--- a/internal/expr/parser.go
+++ b/internal/expr/parser.go
@@ -37,20 +37,9 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
 		}
 	}()
 
-	p.init(input)
+	p.init(input, 0)
 
-	for {
-		// Advance the parser to the start of the next expression.
-		if err := p.advance(); err != nil {
-			return nil, err
-		}
-
-		p.currentExprStart = p.pos
-
-		if p.pos == len(p.input) {
-			break
-		}
-
+	for p.pos < len(p.input) {
 		if out, ok, err := p.parseOutputExpr(); err != nil {
 			return nil, err
 		} else if ok {
@@ -64,10 +53,16 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
 			p.add(in)
 			continue
 		}
-	}
 
+		// Advance the parser to the start of the next expression.
+		if err := p.advanceUntil(nil); err != nil {
+			return nil, err
+		}
+	}
 	// Add any remaining unparsed string input to the parser.
-	p.add(nil)
+	if p.pos != p.prevExprEnd {
+		p.exprs = append(p.exprs, &bypass{p.input[p.prevExprEnd:p.pos]})
+	}
 	return &ParsedExpr{exprs: p.exprs}, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/expr/parser.go
```diff
@@ -37,20 +37,9 @@ func (p *Parser) Parse(input string) (pe *ParsedExpr, err error) {
...
-		if err := p.advance(); err != nil {
-			return nil, err
-		}
-
-		p.currentExprStart = p.pos
-
-		if p.pos == len(p.input) {
-			break
-		}
+	p.init(input, 0)
```
Aflynn50, line 40: I'm not sure about adding the extra parameter to `init` without a use case in this PR. This is a pretty easy thing to do when needed since it doesn't require much change. Its one way to do the function parsing but its still possible we change our minds on how to do it.

internal/expr/parser.go
```diff
@@ -734,21 +733,21 @@ func (p *Parser) parseTargetTypes() (types []memberAccessor, parentheses bool, o
 // parseOutputExpr requires that the ampersand before the identifiers must
 // be followed by a name byte.
 func (p *Parser) parseOutputExpr() (*outputExpr, bool, error) {
-	start := p.pos
+	cp := p.save()
+	p.currentExprStart = p.pos
```
Aflynn50, line 737: Having `currentExprStart` set here and used in `add` feels a bit weird to me. One of the abstractions I don't really want to break with the parser is that we do all the setting of the `p.` variables in the functions directly related to the operation of the parser i.e. `Parse`, `advance` and `add` and not elsewhere. 

I think even just abstracting this to a separate function rather than setting it directly here would be good. Then that function can sit beside `add` in the code to make things a bit more readable. This also has the benefit of forcing some documentation.

Aflynn50 (commented):
I think these changes may be useful, but in my opinion I think it would make more sense to bring them with the code that uses them. That way we can get the context for them and decide if they are necessary or if there are other options that would also work.

We're adding a lot of extra features to the parser here without actually using (or testing) them.

internal/expr/parser.go
letFunny: I agree `add` could be refactored but I think setting the `currentExprStart` in `advance` is very difficult to reason about. It makes much more sense to set it when you are finding an expression and only if you have found one. Else you get jumps for now reason every time you call advance which makes other ways of parsing (e.g. recursive) very hacky.

Conceptually I will like to set the expression start only when there is an expression (what we are achieving with the checkpoint there). What we can do is change the `add` function and separate it into one that adds the bypass and is documented as such. At the same time I dont want to add the part inside the `parseOutputExpr` because you might want to parse and not add it (e.g. to collect a list and discard it if necessary).

internal/expr/parser.go
letFunny: The only reason it is here is because I dont want anyone to have to debug it when the need arises later. It is not trivial to see that you need to set all three variables and keep the input for error reporting purposes (in fact we made the same mistake). Think about adding scaffolding so that us or other people do not have to re-do the investigation later.

letFunny:
@Aflynn50 Agree on the tests, I was trying to publish this as a draft PR to discuss it and then add the tests once we agreed on what to include.

PR #138, opened 2024-02-07
Return ErrNoRows when GetAll finds no results

Aflynn50:
If `GetAll` finds no results in the database it is currently a no-op. In the interests of safety it would be better if this function returned the `ErrNoRows` error. If no results is not an error case the user can explicitly match against and ignore the error.

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -658,19 +658,19 @@ func (s *PackageSuite) TestValidGetAll(c *C) {
 		slices:   []any{&[]*Manager{}, &[]*Person{}, &[]*Address{}},
 		expected: []any{&[]*Manager{{30, "Fred", 1000}}, &[]*Person{{30, "Fred", 1000}}, &[]*Address{{1000, "Happy Land", "Main Street"}}},
 	}, {
-		summary:  "nothing returned",
-		query:    "SELECT &Person.* FROM person WHERE id = $Person.id",
-		types:    []any{Person{}},
-		inputs:   []any{Person{ID: 1243321}},
-		slices:   []any{&[]*Person{}},
-		expected: []any{&[]*Person{}},
-	}, {
 		summary:  "select into maps",
 		query:    "SELECT &M.name, &CustomMap.id FROM person WHERE name = 'Mark'",
 		types:    []any{sqlair.M{}, CustomMap{}},
 		inputs:   []any{},
 		slices:   []any{&[]sqlair.M{}, &[]CustomMap{}},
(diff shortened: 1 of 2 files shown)


sqlair.go
```diff
@@ -300,8 +302,10 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 		sliceVals = append(sliceVals, sliceVal)
 	}
 
+	noRows := true
```
letFunny: [Nit] I think normally you would define the boolean as starting in false (the default value) so that you do not have to negate the name. What about `returnsRows`? Then the check below is: `if !returnsRows && q.pq.HasOutputs`

package_test.go
```diff
@@ -657,20 +657,20 @@ func (s *PackageSuite) TestValidGetAll(c *C) {
...
-		expected: []any{&[]*Person{}},
 	}, {
 		summary:  "select into maps",
 		query:    "SELECT &M.name, &CustomMap.id FROM person WHERE name = 'Mark'",
 		types:    []any{sqlair.M{}, CustomMap{}},
 		inputs:   []any{},
 		slices:   []any{&[]sqlair.M{}, &[]CustomMap{}},
 		expected: []any{&[]sqlair.M{{"name": "Mark"}}, &[]CustomMap{{"id": int64(20)}}},
+	}, {
+		summary:  "run insert with GetAll",
```
letFunny: I think we should name the test something like `GetAll returns no error when there are no outputs`, that way we won't be tempted to refactor it if we add tests for inserts later.

letFunny (commented):
Thanks for the PR, I have some minor comments.

package_test.go
```diff
@@ -766,6 +766,13 @@ func (s *PackageSuite) TestGetAllErrors(c *C) {
 		inputs:  []any{},
 		slices:  []any{&[]Person{}},
 		err:     `cannot populate slice: output variables provided but not referenced in query`,
+	}, {
+		summary: "nothing returned",
+		query:   "SELECT &Person.* FROM person WHERE id = $Person.id",
+		types:   []any{Person{}},
+		inputs:  []any{Person{ID: 1243321}},
+		slices:  []any{&[]*Person{}},
+		err:     "cannot populate slice: sql: no rows in result set",
```
letFunny, line 779: Sorry I missed this on a previous review but should we assert that the error is `ErrNoRows`? If we have the same message but not the exact error we could break the public API. The rest is perfect!

Aflynn50: Added this test further down the file next to the existing one for `Get`.

letFunny (approved):

manadart (approved):

PR #139, opened 2024-03-07
feat: insert expr support

letFunny:
This PR completes the support for the new insert expression syntax.

### DIFF
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -59,12 +59,43 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 				sqlStr.WriteString("@sqlair_" + strconv.Itoa(inputCount))
 				inputCount++
 			}
-		case *typedOutputExpr:
-			for i, oc := range te.outputColumns {
-				sqlStr.WriteString(oc.sql(outputCount))
-				if i != len(te.outputColumns)-1 {
+		case *typedInsertExpr:
+			// Write out the columns.
+			sqlStr.WriteString("(")
+			for i, col := range te.columns {
+				if i != 0 {
 					sqlStr.WriteString(", ")
 				}
+				sqlStr.WriteString(col)
+			}
+			sqlStr.WriteString(") VALUES (")
+			// Write out the values.
+			for i, input := range te.inputs {
+				if i != 0 {
+					sqlStr.WriteString(", ")
+				}
+				vals, err := input.LocateParams(typeToValue)
+				if err != nil {
+					return nil, err
+				}
+				argTypeUsed[input.ArgType()] = true
+				for j, val := range vals {
+					if j != 0 {
+						sqlStr.WriteString(", ")
+					}
+					namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val.Interface())
+					params = append(params, namedInput)
+					sqlStr.WriteString("@sqlair_" + strconv.Itoa(inputCount))
+					inputCount++
+				}
+			}
+			sqlStr.WriteString(")")
+		case *typedOutputExpr:
+			for i, oc := range te.outputColumns {
+				if i != 0 {
+					sqlStr.WriteString(", ")
+				}
+				sqlStr.WriteString(oc.sql(outputCount))
 				outputCount++
 				outputs = append(outputs, oc.output)
 			}
@@ -108,6 +139,13 @@ func newOutputColumn(tableName string, columnName string, output typeinfo.Output
 // typedInputExpr stores information about a Go value to use as a query input.
 type typedInputExpr struct {
 	input typeinfo.Input
+}
+
+// typedInsertExpr stores information about the Go values to use as inputs inside
+// an INSERT statement.
+type typedInsertExpr struct {
+	columns []string
+	inputs  []typeinfo.Input
 }
 
 // typedOutputExpr contains the columns to fetch from the database and
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -136,7 +136,8 @@ func newOutputColumn(tableName string, columnName string, output typeinfo.Output
 	return outputColumn{column: tableName + "." + columnName, output: output}
 }
 
-// typedInputExpr stores information about a Go value to use as a query input.
+// typedInputExpr stores information about a Go value to use as a standalone query
+// input.
 type typedInputExpr struct {
 	input typeinfo.Input
 }
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -133,7 +133,7 @@ func (e *asteriskInsertExpr) String() string {
 	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
 }
 
-// bindTypes generates a *typedInputExpr containing type information about the
+// bindTypes generates a *typedInsertExpr containing type information about the
 // asteriskInsertExpr.
 func (e *asteriskInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err error) {
 	defer func() {
@@ -178,7 +178,7 @@ func (e *columnsInsertExpr) String() string {
 	return fmt.Sprintf("ColumnInsert[%v %v]", e.columns, e.sources)
 }
 
-// bindTypes generates a *typedInputExpr containing type information about the
+// bindTypes generates a *typedInsertExpr containing type information about the
 // columnsInsertExpr. It checks that all the listed columns are provided by the
 // supplied types. If a map with an asterisk is passed, the spare columns are
 // taken from that map.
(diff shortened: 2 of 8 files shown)


internal/expr/bindinputs.go
```diff
@@ -59,12 +59,43 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 				sqlStr.WriteString("@sqlair_" + strconv.Itoa(inputCount))
 				inputCount++
 			}
+		case *typedInsertExpr:
```
Aflynn50, line 62: In previous PR's we've talked about moving this switch case to be methods on the `typedInputExpr`. I think we decided against it because it would have been a lot of boilerplate for not much gain. Though now, in my opinion, its probably worth doing, since this function has grown so much.

The annoying thing about putting these things in methods is that they change the input/output count as well as adding the params. Also that we are actually doing something completely different for the outputs and the inputs alongside generating the SQL (generating named params vs extracting the `arginfo.Output`s).

IMO this change should be done in a subsequent PR, since frankly, the code is a lot simpler to read/review in this form and finding the least worst way to split this into functions may take some iterations.

internal/expr/bindinputs.go
```diff
@@ -110,6 +141,13 @@ type typedInputExpr struct {
 	input typeinfo.Input
```
Aflynn50, line 142: Maybe worth updating the comment on this to somehow imply that doesn't include the input expressions in INSERT statements.

internal/expr/bindtypes.go
```diff
@@ -132,8 +133,35 @@ func (e *asteriskInsertExpr) String() string {
 	return fmt.Sprintf("AsteriskInsert[[*] %v]", e.sources)
 }
 
-func (e *asteriskInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
-	return nil, fmt.Errorf("insert input expression not implemented")
+// bindTypes generates a *typedInputExpr containing type information about the
```
Aflynn50:
```suggestion
// bindTypes generates a *typedInsertExpr containing type information about the
```

internal/expr/bindtypes.go
```diff
@@ -150,8 +178,80 @@ func (e *columnsInsertExpr) String() string {
 	return fmt.Sprintf("ColumnInsert[%v %v]", e.columns, e.sources)
 }
 
-func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (any, error) {
-	return nil, fmt.Errorf("insert input expression not implemented")
+// bindTypes generates a *typedInputExpr containing type information about the
```
Aflynn50:
```suggestion
// bindTypes generates a *typedInsertExpr containing type information about the
```

internal/expr/bindtypes.go
```diff
@@ -150,8 +178,80 @@ func (e *columnsInsertExpr) String() string {
...
+func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("input expression: %s: %s", err, e.raw)
+		}
+	}()
+
+	// 1. We go over all the supplied types and collect all the columns that
+	// they provide.
+	var remainingMap *string
```
Aflynn50: Maybe put this below the colToInput, having it in the middle of the two comments confused me.

internal/expr/bindtypes.go
```diff
@@ -150,8 +178,80 @@ func (e *columnsInsertExpr) String() string {
...
+// supplied types. If a map with an asterisk is passed, the spare columns are
+// taken from that map.
+func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("input expression: %s: %s", err, e.raw)
+		}
+	}()
+
+	// 1. We go over all the supplied types and collect all the columns that
```
Aflynn50: Could maybe change this comment to:
```
Work out all the columns available on the right hand side of the insert expression. For each valueAccessor on the right, save all the columns names can be found in the type it specifies
```

internal/expr/bindtypes.go
Aflynn50: Could maybe have a seperate comment about the remaining map.

internal/expr/bindtypes.go
```diff
@@ -150,8 +178,80 @@ func (e *columnsInsertExpr) String() string {
...
+		} else {
+			inp, err := argInfo.InputMember(source.typeName, source.memberName)
+			if err != nil {
+				return nil, err
+			}
+			colToInput[source.memberName] = []typeinfo.Input{inp}
+		}
+	}
+
+	// 2. We go over all the columns listed in SQL and match them against the
```
Aflynn50: Could change it to:
```
	// 2. We go over all the columns listed on the left of the insert expression and match them against the
	// columns provided by the types on the right. If a map with an asterisk is present, we use it to supply the
	// spare columns. If it is not present, a spare column is an error.
```

package_test.go
```diff
@@ -819,7 +819,23 @@ func (s *PackageSuite) TestRun(c *C) {
 	var jimCheck = Person{}
 	err = db.Query(nil, selectStmt, &jim).Get(&jimCheck)
 	c.Assert(err, IsNil)
-	c.Assert(jimCheck, Equals, jim)
+
```
Aflynn50, line 823: It would be good to dogfood the new syntax in the `personAndAddressDB` function as well.

Aflynn50 (commented):
Excellent work

internal/expr/bindinputs.go
letFunny: I changed it to:
```
// typedInputExpr stores information about a Go value to use as a standalone query
// input.
```

internal/expr/bindtypes.go
letFunny: Agree, I think yours is more explicit

internal/expr/bindtypes.go
letFunny: Agree

letFunny: Agree

internal/expr/bindtypes.go
letFunny: Agreed and changed

package_test.go
letFunny: Good idea! Done in the last commit!

internal/expr/bindinputs.go
letFunny: > The annoying thing about putting these things in methods is that they change the input/output count as well as adding the params. Also that we are actually doing something completely different for the outputs and the inputs alongside generating the SQL (generating named params vs extracting the arginfo.Outputs).

Another annoying thing here is that even though the code is very similar the types are not the same, so we'll have to create an interface or something to refactor it. I agree with you that the function is too big but I am not sure whether the remedy would be worse than the cure.

letFunny: > IMO this change should be done in a subsequent PR, since frankly, the code is a lot simpler to read/review in this form and finding the least worst way to split this into functions may take some iterations.

I agree with you and we should explore it further to see if it is worth it. It is best in another PR because it would probably change a lot of code and the priority is getting inserts in.

Aflynn50 (approved):

internal/expr/bindinputs.go
manadart, line 62: _Code Complete 6.3 Design and Implementation Issues_:
"Prefer polymorphism to extensive type checking".

Of course we don't have polymorphism in Go, but I feel like we could do something as mentioned.
Happy for it to come later.

internal/typeinfo/arginfo_test.go
```diff
@@ -61,10 +61,15 @@ func (s *typeInfoSuite) TestArgInfoStruct(c *C) {
...
 	}}
 
-	allOutputs, names, err := argInfo.AllStructOutputs("myStruct")
+	allOutputs, outputNames, err := argInfo.AllStructOutputs("myStruct")
 	c.Assert(err, IsNil)
 	c.Assert(allOutputs, HasLen, len(structFields))
 
+	allInputs, inputNames, err := argInfo.AllStructInputs("myStruct")
+	c.Assert(err, IsNil)
+	c.Assert(allInputs, HasLen, len(structFields))
```
manadart: After you have asserted a nil error, which you want to fail immediately to prevent later check panics, prefer to use `c.Check`, which does not jump out of the test immediately. That way you get more feedback for failure cases.

manadart (approved):
Nice job. Consider the check fixes, and we're good to go.

internal/typeinfo/arginfo_test.go
Aflynn50: Changed.

internal/expr/bindinputs.go
Aflynn50: Yea, we were thinking the same and have planned to do it in a subsequent PR. This chunk of code is particularly annoying because it really does quite different things for the different structs.

I'll have a go at separating it out.

PR #140, opened 2024-03-11
Add support for parsing unicode chars in column names and tags

Aflynn50:
The SQLair parser does not currently recognise unicode characters in column names. It also doesn't support a way to select numbers into outputs, or to select column names with quotes into outputs. Adding these features will increase the reliability and polish of SQLair.

The parser is changed to check runes rather than bytes, it also allows for numbers and quoted names to be parsed as column names. The validation for tag names is also changed to include the above so that these columns can be parsed directly into structs.

You will now be able to write SQLair queries such as:
```
SELECT 世界 AS &S.* FROM t
SELECT 1 AS &M.doesExist FROM t
SELECT "string or column name!" AS $M.quoted FROM t
```

Also tidied up use of go check in some of the tests, and added tests for lower case structs.

P.S. sorry for the poor commit history, this will all get squashed away.

### DIFF
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -53,7 +53,7 @@ type NumberLiteralColumn struct {
 
 type M map[string]any
 
-type lowerCaseMap map[string]any
+type unexportedMap map[string]any
 
 type IntMap map[string]int
 
@@ -84,8 +84,8 @@ var tests = []struct {
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
-	summary: "spaces and tabs",
-	query: "SELECT p.* 	AS 		   &Person.*",
+	summary:        "spaces and tabs",
+	query:          "SELECT p.* 	AS 		   &Person.*",
 	expectedParsed: "[Bypass[SELECT ] Output[[p.*] [Person.*]]]",
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
@@ -535,9 +535,9 @@ AND z = @sqlair_0 -- The line with $Person.id on it
 	expectedSQL:    "INSERT INTO arr VALUES (ARRAY[[1,2],[@sqlair_0,4]], ARRAY[[5,6],[@sqlair_1,8]]);",
 }, {
 	summary:        "lower case struct",
-	query:          "SELECT &lowerCaseMap.x FROM person",
-	expectedParsed: "[Bypass[SELECT ] Output[[] [lowerCaseMap.x]] Bypass[ FROM person]]",
-	typeSamples:    []any{lowerCaseMap{}},
+	query:          "SELECT &unexportedMap.x FROM person",
+	expectedParsed: "[Bypass[SELECT ] Output[[] [unexportedMap.x]] Bypass[ FROM person]]",
+	typeSamples:    []any{unexportedMap{}},
 	expectedSQL:    "SELECT x AS _sqlair_0 FROM person",
 }, {
 	summary:        "unicode asterisk",
--- a/internal/expr/expr_test.go
+++ b/internal/expr/expr_test.go
@@ -39,11 +39,24 @@ type HardMaths struct {
 	Coef int `db:"coef"`
 }
 
+type QuotedColumnNames struct {
+	Ex int `db:"\"!!!\""`
+	Qu int `db:"'???'"`
+}
+
 type M map[string]any
+
+type lowerCaseMap map[string]any
 
 type IntMap map[string]int
 
 type StringMap map[string]string
+
+type Unicode我Struct struct {
+	X人 int    `db:"საფოსტო"`
+	X我 int    `db:"住所"`
+	X这 string `db:"鑑別"`
+}
 
 var tests = []struct {
 	summary          string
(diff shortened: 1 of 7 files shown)


internal/expr/parser.go
```diff
@@ -343,34 +365,33 @@ func (p *Parser) skipStringLiteral() (bool, error) {
...
 }
 
-// skipByteFind looks for a byte that matches the one passed as parameter and
+// skipCharFind looks for a char that matches the one passed as parameter and
 // then advances the parser to jump over it. In that case returns true. If the
-// end of the string is reached and no matching byte was found, it returns
+// end of the string is reached and no matching char was found, it returns
 // false and it does not change the parser.
-func (p *Parser) skipByteFind(b byte) bool {
+func (p *Parser) skipCharFind(c rune) bool {
```
TheSignPainter98, line 389: As Go's `rune` type already handles what a char is (not to be confused with C's 1-byte chars), I think it could be nice just to call these runes everywhere to avoid a duplicate concept

internal/expr/parser.go
```diff
@@ -400,36 +421,39 @@ func (p *Parser) skipBlanks() bool {
 func (p *Parser) skipString(s string) bool {
 	if p.pos+len(s) <= len(p.input) &&
 		strings.EqualFold(p.input[p.pos:p.pos+len(s)], s) {
+		// Manually advance the parser to the end of the string.
```
TheSignPainter98: Is this comment necessary? I think this function is short enough that its doc-comment is sufficient :P

internal/expr/parser.go
```diff
@@ -400,36 +421,39 @@ func (p *Parser) skipBlanks() bool {
...
+func isNameChar(c rune) bool {
+	return unicode.IsLetter(c) || unicode.IsDigit(c) || c == '_'
 }
 
-// isNameInitialByte returns true if the given byte can appear at the start of a
+// isNameInitialChar returns true if the given char can appear at the start of a
 // name. It returns false otherwise.
-func isInitialNameByte(c byte) bool {
-	return 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' || c == '_'
+func isInitialNameChar(c rune) bool {
```
TheSignPainter98, line 446: The function name and doc-comment name are slightly out of alignment

internal/expr/expr_test.go
```diff
@@ -66,8 +84,8 @@ var tests = []struct {
 	typeSamples:    []any{Person{}},
 	expectedSQL:    "SELECT p.address_id AS _sqlair_0, p.id AS _sqlair_1, p.name AS _sqlair_2",
 }, {
-	summary:        "spaces and tabs",
-	query:          "SELECT p.* 	AS 		   &Person.*",
+	summary: "spaces and tabs",
+	query: "SELECT p.* 	AS 		   &Person.*",
```
TheSignPainter98: Not sure what happened here but I think this looks unformatted?

internal/expr/expr_test.go
```diff
@@ -41,8 +41,20 @@ type HardMaths struct {
...
+}
+
+type NumberLiteralColumn struct {
+	Zero int `db:"0"`
+	One  int `db:"1"`
+}
+
 type M map[string]any
 
+type lowerCaseMap map[string]any
```
TheSignPainter98: I don't quite follow what's lowercase here, is this testing unexported types? In this case I think `unexportedMap` would be more descriptive

nit: I think the 'C' in the name should be lowercase.

TheSignPainter98 (commented):
Very nice! Just some nitpicks :)

internal/expr/parser.go
Aflynn50: Fair, makes sense, though I chose char for this because it doesn't tie it to either of the implementations. In most cases the code is still unchanged and is operating with the same literals as it was with bytes (i.e. 'b'). If you have a load of `skipRune` funcs everywhere it might imply that there is also a `skipByte` whereas `skipChar` is more clearly the default option.

internal/expr/parser.go
Aflynn50: This `skipString` func is funky. We're using that `EqualFold` function because its case insensitive, this is really the only point in the parser where we are not matching the chars ourselves. This means its also the only point where we manually advance the parser outside of `nextChar`.

Tbh I should put some more docs in this func to explain that its a bit strange.

internal/expr/parser.go
Aflynn50: Nice spot.

internal/expr/expr_test.go
Aflynn50: This is my number one reason for switch to Goland. No idea what is wrong with my vim setup, I've tried changing gofmt versions, but it doesn't seem to be using the one on my path.

internal/expr/expr_test.go
Aflynn50: Good point, will change it to that.

internal/expr/parser.go
```diff
@@ -398,38 +419,43 @@ func (p *Parser) skipBlanks() bool {
 // In that case returns true, false otherwise.
 // This function is case insensitive.
 func (p *Parser) skipString(s string) bool {
+	// EqualFold is used here because it is case insensitive.
 	if p.pos+len(s) <= len(p.input) &&
 		strings.EqualFold(p.input[p.pos:p.pos+len(s)], s) {
+		// EqualFold does not advnace the parser, so we must manually advance
```
manadart: Typo with "advance".

internal/expr/parser.go
```diff
@@ -486,48 +512,76 @@ func (p *Parser) skipEnclosedParentheses() (bool, error) {
...
 
 	// asterisk cannot be followed by anything.
-	if p.skipByte('*') {
+	if p.skipChar('*') {
 		return basicColumn{column: "*"}, true, nil
 	}
 
-	id, ok := p.parseIdentifier()
+	// Parse a column or a table name.
+	id, ok, err := p.parseColumnName()
```
manadart: Either the comment isn't right, or this method is poorly named.

internal/typeinfo/arginfo.go
```diff
@@ -312,9 +309,38 @@ func parseTag(tag string) (string, bool, error) {
...
+			return "", false, fmt.Errorf("missing quotes at end of 'db' tag: %q", name)
+		}
+		// No need to validate chars in quotes.
+		return name, omitEmpty, nil
+	}
+
+	char, size := utf8.DecodeRuneInString(name)
+	nextPos := size
+	var checker func(rune) bool
+	if unicode.IsDigit(char) {
```
manadart: This could be a case statement.

internal/typeinfo/arginfo.go
```diff
@@ -312,9 +309,38 @@ func parseTag(tag string) (string, bool, error) {
...
+		// If it starts with a digit, check the tag is a number.
+		checker = func(char rune) bool {
```
manadart: `checker = unicode.IsDigit`.

manadart (approved):
Looks OK, with a few semantic things to check.

Prior reviewer made a suggestion regarding making consistent reference to runes. I know we can't name something `rune`, but the comments are not consistent between char/rune.

internal/expr/parser.go
Aflynn50: Changed the name back to `parseIdentifer` and reinstate `parseIdentifierAsterisk`. Identifier matches the SQL column and table names as well as what we accept for db tags, since this is the same thing. A generic name like Identifier makes more sense.

PR #141, opened 2024-03-11
Add embedded struct support

Aflynn50:
Embedded structs in Go are not automatically promoted to the containing struct. No explicit support for them has been added to SQLair before so they are not currently supported.

This PR adds support for arbitrary levels of embedded structs. All fields in the embedded structs will be promoted to the top level.

You will now be able to do:
```
type Embedded struct {
  F int `db:"embeddedColName"`
}

type S struct {
  Embedded
}

stmt, err := sqlair.Prepare("SELECT &S.embeddedColName FROM t", S{})
```

Also:
- Add check for repeated tags.
- Add some tests for scanner/valuer interfaces

### DIFF
--- a/internal/typeinfo/arginfo.go
+++ b/internal/typeinfo/arginfo.go
@@ -345,8 +345,11 @@ func getStructFields(structType reflect.Type) ([]*structField, error) {
 		tag := field.Tag.Get("db")
 
 		// If Anonymous is true, the field is embedded.
-		// If the embedded struct is tagged, treat it as a regular field.
 		if field.Anonymous && tag == "" {
+			// If the embedded struct is tagged then we do not look inside it and
+			// we pass it straight to the driver. This means it must implement the
+			// Valuer or Scanner interface (for inputs/outputs respectively) or the
+			// driver will reject it with a panic.
 			if !field.IsExported() {
 				continue
 			}
--- a/internal/typeinfo/arginfo.go
+++ b/internal/typeinfo/arginfo.go
@@ -245,29 +245,19 @@ func getArgInfo(t reflect.Type) (arg, error) {
 		}
 		tags := []string{}
 
-		for i := 0; i < t.NumField(); i++ {
-			f := t.Field(i)
-			// Fields without a "db" tag are outside of SQLAir's remit.
-			tag := f.Tag.Get("db")
-			if tag == "" {
-				continue
-			}
-			if !f.IsExported() {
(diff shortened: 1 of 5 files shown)


internal/typeinfo/arginfo.go
```diff
@@ -345,6 +335,65 @@ func parseTag(tag string) (string, bool, error) {
...
+			}
+			// Promote the embedded struct fields into the current parent struct
+			// scope, making sure to update the Index list for navigation back
+			// to the original nested location.
+			nestedFields, err := getStructFields(fieldType)
+			if err != nil {
+				return nil, err
+			}
+			for _, nestedField := range nestedFields {
+				nestedField.index = append([]int{i}, nestedField.index...)
```
manadart, line 372: I didn't know that you could relocate these back in a composite struct via this index-as-slice...

internal/typeinfo/arginfo_test.go
```diff
@@ -137,6 +137,76 @@ func (s *typeInfoSuite) TestArgInfoMap(c *C) {
...
+	type Embedded2 struct {
+		F2 int `db:"col2"`
+		Embedded3
+	}
+	type Embedded1 struct {
+		F1 int `db:"col1"`
+	}
+	type Embeddings struct {
+		EmbeddedString
+		StructWithTag `db:"col4"`
```
manadart: What happens with such cases?
- I assume select x.* just causes the SQL to run but fail on _get_ if the struct will not hold the return.
- Attempting to insert would fail during parameter binding for incompatible values.

manadart (approved):
I think this is OK. I've got a little twitch about _stucts_ that are tagged.

internal/typeinfo/arginfo_test.go
Aflynn50: If the struct is tagged then it will be passed straight to the driver, this means it must implement the Valuer/Scanner for inputs/outputs respectively.

A comment has been added in the `getStructFields` function to explain this behavior.

internal/typeinfo/arginfo.go
Aflynn50: Thats the difference between the illustratively named `.Field` and `.FieldByIndex` functions on `reflect.Value`

internal/typeinfo/arginfo_test.go
Aflynn50: Also added a tagged field in the struct in this test case to verify that it is ignored by SQLair

PR #142, opened 2024-03-18
Allow omitempty flag to be used with INSERT statements

Aflynn50:
When inserting a into a table with unique primary keys the user needs to specify the primary key in their object every time. If it is left to its default zero value it will most likely break the unique constraint. In some tables, the primary key can is auto when a new row is inserted. In this case the user may want to insert a full struct using the asterisk notation but omitting the primary key.

To automatically omit a primary key, the user can set a flag in the `db` tag after the column name for the field e.g. 
```
type Person struct {
	ID        int    `db:"id, omitempty"`
	Name      string `db:"name"`
	PostCode  int    `db:"address_id"`
}
```
This will indicate to SQLair that if this value is zero when the parameter is passed for execution, it should not be inserted into the database.

As specified in the [INSERT statement support specification](https://docs.google.com/document/d/1DMKPS85msymrdbA5-EoP4Sf6b9blSAjtj367EuEmP7Y/edit), the omit empty flag should only work for struct fields being inserted via an asterisk. If the field is inserted explicitly, e.g. `INSERT INTO t (id) VALUES ($Person.id)` then we should throw an error. Similarly, if a tag with `omitempty` set is used in an input expression elsewhere, an error should also be thrown e.g. `SELECT &Person.* FROM person WHERE id = $Person.id`.

### DIFF
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -49,10 +49,10 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 			if err != nil {
 				return nil, err
 			}
-
 			if omit {
 				return nil, omitEmptyInputError(te.input.Desc())
 			}
+
 			argTypeUsed[te.input.ArgType()] = true
 			sqlBuilder.writeInput(inputCount, len(vals))
 			for _, val := range vals {
@@ -142,6 +142,7 @@ type insertColumn struct {
 	explicit bool
 }
 
+// newInsertColumn builds an insert column.
 func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertColumn {
 	return insertColumn{
 		input:    input,
@@ -162,6 +163,8 @@ type typedOutputExpr struct {
 	outputColumns []outputColumn
 }
 
+// sqlBuilder is used to generate SQL string piece by piece using the struct
+// methods.
 type sqlBuilder struct {
 	buf bytes.Buffer
 }
@@ -211,7 +214,7 @@ func (b *sqlBuilder) write(sql string) {
 	b.buf.WriteString(sql)
 }
 
-// write writes the SQL to the sqlBuilder.
+// getSQL returns the generated SQL string
 func (b *sqlBuilder) getSQL() string {
 	return b.buf.String()
 }
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -41,66 +41,61 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 	argTypeUsed := map[reflect.Type]bool{}
 	inputCount := 0
 	outputCount := 0
-	sqlStr := bytes.Buffer{}
+	var sqlBuilder sqlBuilder
 	for _, te := range *tbe {
 		switch te := te.(type) {
 		case *typedInputExpr:
-			vals, err := te.input.LocateParams(typeToValue)
+			vals, omit, err := te.input.LocateParams(typeToValue)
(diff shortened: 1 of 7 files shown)


internal/expr/bindinputs.go
```diff
@@ -41,66 +41,61 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
 					return nil, err
 				}
-				argTypeUsed[input.ArgType()] = true
-				for j, val := range vals {
-					if j != 0 {
-						sqlStr.WriteString(", ")
+				if len(vals) > 1 {
+					// Only slices return multiple values and they are not
+					// allowed in insert expressions.
+					return nil, fmt.Errorf("internal error: unexpected values")
```
manadart: The error might be more descriptive if the text is along the lines of the comment above.

internal/expr/bindinputs.go
```diff
@@ -142,11 +132,20 @@ type typedInputExpr struct {
...
+	// query. If the column is inserted via an asterisk type, it is false.
+	explicit bool
+}
+
 // typedInsertExpr stores information about the Go values to use as inputs inside
 // an INSERT statement.
 type typedInsertExpr struct {
-	columns []string
-	inputs  []typeinfo.Input
+	insertColumns []insertColumn
```
manadart, line 157: I like the way you're thinking here.

internal/expr/bindinputs.go
```diff
@@ -155,6 +154,61 @@ type typedOutputExpr struct {
 	outputColumns []outputColumn
 }
 
+type sqlBuilder struct {
```
manadart, line 168: This encapsulation is a nice addition.

internal/expr/bindinputs.go
```diff
@@ -155,6 +154,61 @@ type typedOutputExpr struct {
...
+	buf bytes.Buffer
+}
+
+// writeInsert writes the SQL for INSERT statements to the sqlBuilder.
+func (b *sqlBuilder) writeInsert(inputCount int, columns []string) {
+	// Write out the columns.
+	b.buf.WriteString("(")
+	for i, col := range columns {
+		if i != 0 {
```
manadart: This repeated block could be a method accepting `i` and the string to write (`col` in this case).

internal/expr/bindtypes.go
```diff
@@ -142,26 +142,33 @@ func (e *asteriskInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err e
...
 		if source.memberName == "*" {
-			inps, tags, err := argInfo.AllStructInputs(source.typeName)
+			inputs, tags, err := argInfo.AllStructInputs(source.typeName)
 			if err != nil {
 				return nil, err
 			}
-			columns = append(columns, tags...)
-			inputs = append(inputs, inps...)
+			for i, input := range inputs {
+				cols = append(cols, insertColumn{
```
manadart: Worth a constructor that makes this `insertColumn`, then these are nice one-line functions.

manadart (approved):
Nice job on this one. A little house-keeping and we're good.

internal/expr/bindinputs.go
Aflynn50: Changed.

internal/expr/bindinputs.go
Aflynn50: Added a `sqlBuilder.writeCommaSeperatedList` which accepts a writer function for printing each list element to the buffer.

internal/expr/bindtypes.go
Aflynn50: Added the constructor. This now also mirrors the `newOutputColumn` constructor.

PR #143, opened 2024-03-18
Bugfix: Expose SQL ErrNoRows from GetAll by removing unhelpful wrapper

Aflynn50:
The `sql.ErrNoRows` errors were being wrapped in a string by `GetAll` and the tests for this were broken. This fixes the bug by removing the somewhat unhelpful wrapping text and fixing the tests.


manadart (approved):
I think you can use `%w`, but "cannot..." is redundant with an error anyway.

PR #144, opened 2024-03-21
Use vars for database entries and add INSERT tests

Aflynn50:
The tests currently check query results against expected literals. With INSERT statement support we can now easily insert variables into the database. We can simply check the results against these variables instead to make the tests more standard. This PR also generally cleans up the tests.

- Add variables for rows in the database
- Clear up DB creation function
- Add test for INSERT statements
- Change the order of the tests to group tests of core functions together vs tests of types of query
- Remove redundant Juju lease tests


cderici (approved):
LGTM 👍

PR #145, opened 2024-04-03
Add bulk insert support

Aflynn50:
SQLair currently supports insert expressions, but not bulk inserts. To insert multiple rows into a database with SQLair, users currently have to execute the same statement multiple times using a for loop or write a query that inserts a fixed number of rows by explicitly writing out each row with separate input expressions for every column.

Additionally, users may want to bulk insert where network latency needs to be minimised such as in Juju where the distributed database requires multiple trips.

Bulk inserts are supported using the same syntax currently used for INSERTS. If one or more of the types passed to `Query` is a slice instead of a simple value, then it becomes a bulk insert with any other values in the INSERT remaining constant.

For example:
```
type S struct {
    v string `db:"col1"`
}

type T struct {
   v string `db:"col4"`
}

s := sqlair.MustPrepare(
    "INSERT INTO t (col1, col2, col3, col4) VALUES ($S.*, $M.col2, "v3", $T.col4)",
    S{},
    sqlair.M{},
    T{},
)

db.Query(
    s,
    []S{{v: "v1.1"}, {v: "v1.2"}, {v: "v1.3"}},
    sqlair.M{"col2": "v2"},
    []T{{v: "v4.1"}, {v: "v4.2"}, {v: "v4.3"}},
)
```
The query sent to the DB from the above example will be:
```
INSERT INTO  t (col1, col2, col3, col4) VALUES (@sqlair_0, @sqlair_3, "v3", @sqalir_4), (@sqlair_1, @sqlair_3, "v3", @sqalir_5), (@sqlair_2, @sqlair_3, "v3", @sqalir_6)
```
Where the placeholders are:
Name | Value
---|---
sqlair_0 | v1.1
sqlair_1 | v1.2
sqlair_2 | v1.3
sqlair_3 | v2
sqlair_4 | v4.1
sqlair_5 | v4.2
sqlair_6 | v4.3

Note that SQLair will check that all the slices passed to Query for bulk INSERTS are of the same cardinality, else throw out an error.

### DIFF
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -175,6 +175,8 @@ func (ic insertColumn) bindInputs(tv typeinfo.TypeToValue, ia *inputAssigner) (*
 	}
 	var firstInputNum int
 	if !params.Omit {
+		// Reserve input numbers for all values that are getting inserted in the
+		// boundColumn.
 		firstInputNum = ia.assignInputs(len(params.Vals))
 	}
 	bc := &boundInsertColumn{
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -36,7 +36,7 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 	// Generate SQL and query parameters.
 	var namedInputs []any
 	var outputs []typeinfo.Output
-	// argType is used to check that all the query parameters are
+	// argTypeUsed is used to check that all the query parameters are
 	// referenced in the query.
 	argTypeUsed := map[reflect.Type]bool{}
 	inputCount := 0
@@ -59,12 +59,13 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 			argTypeUsed[params.ArgTypeUsed] = true
 			sqlBuilder.writeInput(inputCount, len(params.Vals))
 			for _, val := range params.Vals {
-				namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val.Interface())
+				namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val)
 				namedInputs = append(namedInputs, namedInput)
 				inputCount++
 			}
 		case *typedInsertExpr:
 			var boundColumns []*boundColumn
+			var columnNames []string
 			bulk := false
 			numRows := 1
 			// firstBulkColumn stores the type name of the first column used in
@@ -78,21 +79,16 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 				bc.markArgUsed(argTypeUsed)
 
 				if bc.bulk {
-					if bulk {
-						if len(bc.vals) != numRows {
-							return nil, mismatchedBulkLengthsError(firstBulkColumn, numRows, bc.inputName, len(bc.vals))
-						}
-					} else {
+					if !bulk {
+						// First bulk row.
 						bulk = true
 						firstBulkColumn = bc.inputName
 						numRows = len(bc.vals)
+					} else if len(bc.vals) != numRows {
+						return nil, mismatchedBulkLengthsError(firstBulkColumn, numRows, bc.inputName, len(bc.vals))
 					}
 				}
 				boundColumns = append(boundColumns, bc)
-			}
-
-			var columnNames []string
-			for _, bc := range boundColumns {
 				if !bc.omit {
 					columnNames = append(columnNames, bc.column)
 				}
@@ -100,16 +96,16 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 
 			var rowsSQL [][]string
 			for rowNum := 0; rowNum < numRows; rowNum++ {
-				rowSQL := []string{}
+				var rowSQL []string
 				for _, bc := range boundColumns {
 					if !bc.omit {
-						valueSQL, value, literal, err := bc.parameter(rowNum, inputCount)
+						valueSQL, namedInput, literal, err := bc.parameter(rowNum, inputCount)
 						if err != nil {
 							return nil, err
 						}
 						rowSQL = append(rowSQL, valueSQL)
 						if !literal {
-							namedInputs = append(namedInputs, value)
+							namedInputs = append(namedInputs, namedInput)
 							inputCount++
 						}
 					}
@@ -163,15 +159,17 @@ type typedInputExpr struct {
 	input typeinfo.Input
 }
 
+// typedColumn represents a column and input locator in an insert statement.
 type typedColumn interface {
+	// bindInputs binds a concrete value to a typedColumn to generate a
+	// boundColumn.
 	bindInputs(typeinfo.TypeToValue) (*boundColumn, error)
 }
 
-// insertColumn stores information about a single column of a row in an INSERT
+// insertColumn stores information about a single column of a row in an insert
 // statement.
 type insertColumn struct {
-	input typeinfo.Input
-	// literal is "" if it is not a literal value.
+	input  typeinfo.Input
 	column string
 	// explicit is true if the column is explicity inserted in the SQLair
 	// query. If the column is inserted via an asterisk type, it is false.
@@ -187,6 +185,8 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
 	}
 }
 
+// bindInputs generates and verifies the query parameters corresponding to the
+// insertColumn and returns them as a boundColumn.
 func (ic insertColumn) bindInputs(tv typeinfo.TypeToValue) (*boundColumn, error) {
 	params, err := ic.input.LocateParams(tv)
 	if err != nil {
@@ -197,8 +197,6 @@ func (ic insertColumn) bindInputs(tv typeinfo.TypeToValue) (*boundColumn, error)
 		// slices are not allowed in insert expressions.
 		return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
 	}
-	// We check for params.Omit after params.Bulk because we still
-	// want to do a bulk insert even if all bulk inputs are omitted.
 	if params.Omit {
 		if ic.explicit {
 			return nil, omitEmptyInputError(ic.input.Desc())
@@ -216,15 +214,19 @@ func (ic insertColumn) bindInputs(tv typeinfo.TypeToValue) (*boundColumn, error)
 	return bc, nil
 }
 
+// literalColumn represents a column in an insert statement populated with a
+// literal value.
 type literalColumn struct {
 	column  string
 	literal string
 }
 
+// bindInputs creates a boundColumn from a literalColumn. It is part of the
+// typedColumn interface.
 func (lc literalColumn) bindInputs(_ typeinfo.TypeToValue) (*boundColumn, error) {
 	bc := &boundColumn{
 		column:    lc.column,
-		vals:      []reflect.Value{},
+		vals:      []any{},
 		omit:      false,
 		bulk:      false,
 		literal:   lc.literal,
@@ -242,10 +244,12 @@ func newLiteralColumn(column, literal string) literalColumn {
 	}
 }
 
+// boundColumn represents a column in an insert statement along with the values
+// to be inserted into it.
 type boundColumn struct {
 	// vals contains the query parameters. There may be multiple for a bulk
 	// insert, or none for a literal.
-	vals []reflect.Value
+	vals []any
 	// omit indicates if the column and values should be ommited.
 	omit bool
 	// bulk is true if the list of values should be inserted in a bulk insert
@@ -262,27 +266,28 @@ type boundColumn struct {
 	column string
 }
 
+// markArgUsed marks the argument type passed to SQL as used in the argUsed map.
 func (bc *boundColumn) markArgUsed(argUsed map[reflect.Type]bool) {
 	if bc.argType != nil {
 		argUsed[bc.argType] = true
 	}
 }
 
+// parameter returns the value to be inserted into the boundColumn in the given
+// row. The inputCount it used to generate the parameter name.
 func (bc *boundColumn) parameter(row, inputCount int) (string, any, bool, error) {
 	if bc.literal != "" {
 		return bc.literal, nil, true, nil
 	}
 	name := "sqlair_" + strconv.Itoa(inputCount)
 	if !bc.bulk {
-		val := bc.vals[0].Interface()
-		return "@" + name, sql.Named(name, val), false, nil
+		return "@" + name, sql.Named(name, bc.vals[0]), false, nil
 	}
 
 	if row >= len(bc.vals) {
 		return "", nil, false, fmt.Errorf("internal error: no bulk insert value for row %d, only have %d values", row, len(bc.vals))
 	}
-	val := bc.vals[row].Interface()
-	return "@" + name, sql.Named(name, val), false, nil
+	return "@" + name, sql.Named(name, bc.vals[row]), false, nil
 }
 
 // typedInsertExpr stores information about the Go values to use as inputs inside
--- a/internal/expr/bindinputs.go
+++ b/internal/expr/bindinputs.go
@@ -36,7 +36,7 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 	// Generate SQL and query parameters.
 	var namedInputs []any
 	var outputs []typeinfo.Output
-	// argTypeUsed is used to check that all the query parameters are
+	// argType is used to check that all the query parameters are
 	// referenced in the query.
 	argTypeUsed := map[reflect.Type]bool{}
 	inputCount := 0
@@ -64,61 +64,59 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
 				inputCount++
 			}
 		case *typedInsertExpr:
-			var columnNames []string
-			var paramsToInsert []*typeinfo.Params
+			var boundColumns []*boundColumn
 			bulk := false
-			// bulkRows is the number of rows in the bulk insert. If it is not a
-			// bulk insert, it remains 1.
-			bulkRows := 1
-			// firstBulkColumn stores the type name of the first inputs that are
-			// part of a bulk insert. This is used for error messages.
+			numRows := 1
+			// firstBulkColumn stores the type name of the first column used in
+			// a bulk insert. This is used for error messages.
 			var firstBulkColumn string
 			for _, ic := range te.insertColumns {
-				params, err := ic.input.LocateParams(typeToValue)
+				bc, err := ic.bindInputs(typeToValue)
 				if err != nil {
 					return nil, err
 				}
-				argTypeUsed[params.ArgTypeUsed] = true
+				bc.markArgUsed(argTypeUsed)
 
-				if params.Bulk {
+				if bc.bulk {
 					if bulk {
-						if len(params.Vals) != bulkRows {
-							return nil, mismatchedBulkLengthsError(firstBulkColumn, bulkRows, ic.input.ArgType(), len(params.Vals))
+						if len(bc.vals) != numRows {
+							return nil, mismatchedBulkLengthsError(firstBulkColumn, numRows, bc.inputName, len(bc.vals))
 						}
 					} else {
 						bulk = true
-						firstBulkColumn = ic.input.ArgType().Name()
-						bulkRows = len(params.Vals)
+						firstBulkColumn = bc.inputName
+						numRows = len(bc.vals)
 					}
-				} else if len(params.Vals) > 1 {
-					// Only slices and bulk inserts return multiple values and
-					// slices are not allowed in insert expressions.
-					return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
 				}
-
-				// We check for params.Omit after params.Bulk because we still
-				// want to do a bulk insert even if all bulk inputs are omitted.
-				if params.Omit {
-					if ic.explicit {
-						return nil, omitEmptyInputError(ic.input.Desc())
-					}
-					continue
-				}
-				columnNames = append(columnNames, ic.column)
-				paramsToInsert = append(paramsToInsert, params)
+				boundColumns = append(boundColumns, bc)
 			}
 
-			sqlBuilder.writeInsert(inputCount, bulkRows, columnNames)
-			for i := 0; i < bulkRows; i++ {
-				for _, params := range paramsToInsert {
-					index := i
-					if !params.Bulk {
-						index = 0
-					}
-					namedInputs = append(namedInputs, sql.Named("sqlair_"+strconv.Itoa(inputCount), params.Vals[index].Interface()))
-					inputCount++
+			var columnNames []string
+			for _, bc := range boundColumns {
+				if !bc.omit {
+					columnNames = append(columnNames, bc.column)
 				}
 			}
+
+			var rowsSQL [][]string
+			for rowNum := 0; rowNum < numRows; rowNum++ {
+				rowSQL := []string{}
+				for _, bc := range boundColumns {
+					if !bc.omit {
+						valueSQL, value, literal, err := bc.parameter(rowNum, inputCount)
+						if err != nil {
+							return nil, err
+						}
+						rowSQL = append(rowSQL, valueSQL)
+						if !literal {
+							namedInputs = append(namedInputs, value)
+							inputCount++
+						}
+					}
+				}
+				rowsSQL = append(rowsSQL, rowSQL)
+			}
+			sqlBuilder.writeInsert(columnNames, rowsSQL)
 		case *typedOutputExpr:
 			var columns []string
 			for _, oc := range te.outputColumns {
@@ -165,10 +163,15 @@ type typedInputExpr struct {
 	input typeinfo.Input
 }
 
+type typedColumn interface {
+	bindInputs(typeinfo.TypeToValue) (*boundColumn, error)
+}
+
 // insertColumn stores information about a single column of a row in an INSERT
 // statement.
 type insertColumn struct {
-	input  typeinfo.Input
+	input typeinfo.Input
+	// literal is "" if it is not a literal value.
 	column string
 	// explicit is true if the column is explicity inserted in the SQLair
 	// query. If the column is inserted via an asterisk type, it is false.
@@ -184,10 +187,108 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
 	}
 }
 
+func (ic insertColumn) bindInputs(tv typeinfo.TypeToValue) (*boundColumn, error) {
+	params, err := ic.input.LocateParams(tv)
+	if err != nil {
+		return nil, err
+	}
+	if !params.Bulk && len(params.Vals) > 1 {
+		// Only slices and bulk inserts return multiple values and
+		// slices are not allowed in insert expressions.
+		return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
+	}
+	// We check for params.Omit after params.Bulk because we still
+	// want to do a bulk insert even if all bulk inputs are omitted.
+	if params.Omit {
+		if ic.explicit {
+			return nil, omitEmptyInputError(ic.input.Desc())
+		}
+	}
+	bc := &boundColumn{
+		column:    ic.column,
+		vals:      params.Vals,
+		omit:      params.Omit,
+		bulk:      params.Bulk,
+		literal:   "",
+		inputName: ic.input.ArgType().Name(),
+		argType:   params.ArgTypeUsed,
+	}
+	return bc, nil
+}
+
+type literalColumn struct {
+	column  string
+	literal string
+}
+
+func (lc literalColumn) bindInputs(_ typeinfo.TypeToValue) (*boundColumn, error) {
+	bc := &boundColumn{
+		column:    lc.column,
+		vals:      []reflect.Value{},
+		omit:      false,
+		bulk:      false,
+		literal:   lc.literal,
+		inputName: "",
+		argType:   nil,
+	}
+	return bc, nil
+}
+
+// newLiteralColumn builds a literal column.
+func newLiteralColumn(column, literal string) literalColumn {
+	return literalColumn{
+		column:  column,
+		literal: literal,
+	}
+}
+
+type boundColumn struct {
+	// vals contains the query parameters. There may be multiple for a bulk
+	// insert, or none for a literal.
+	vals []reflect.Value
+	// omit indicates if the column and values should be ommited.
+	omit bool
+	// bulk is true if the list of values should be inserted in a bulk insert
+	// expression.
+	bulk bool
+	// argType is the type of the argument that was used to generate the
+	// params.
+	argType reflect.Type
+	// inputName is the type name of the input parameter.
+	inputName string
+	// literal set if the value to insert is a literal.
+	literal string
+	// column is the column name.
+	column string
+}
+
+func (bc *boundColumn) markArgUsed(argUsed map[reflect.Type]bool) {
+	if bc.argType != nil {
+		argUsed[bc.argType] = true
+	}
+}
+
+func (bc *boundColumn) parameter(row, inputCount int) (string, any, bool, error) {
+	if bc.literal != "" {
+		return bc.literal, nil, true, nil
+	}
+	name := "sqlair_" + strconv.Itoa(inputCount)
+	if !bc.bulk {
+		val := bc.vals[0].Interface()
+		return "@" + name, sql.Named(name, val), false, nil
+	}
+
+	if row >= len(bc.vals) {
+		return "", nil, false, fmt.Errorf("internal error: no bulk insert value for row %d, only have %d values", row, len(bc.vals))
+	}
+	val := bc.vals[row].Interface()
+	return "@" + name, sql.Named(name, val), false, nil
+}
+
 // typedInsertExpr stores information about the Go values to use as inputs inside
 // an INSERT statement.
 type typedInsertExpr struct {
-	insertColumns []insertColumn
+	insertColumns []typedColumn
 }
 
 // typedOutputExpr contains the columns to fetch from the database and
@@ -203,7 +304,7 @@ type sqlBuilder struct {
 }
 
 // writeInsert writes the SQL for INSERT statements to the sqlBuilder.
-func (b *sqlBuilder) writeInsert(inputCount, bulkRows int, columns []string) {
+func (b *sqlBuilder) writeInsert(columns []string, rows [][]string) {
 	// Write out the columns.
 	b.buf.WriteString("(")
 	b.writeCommaSeperatedList(columns, func(_ int, column string) string {
@@ -211,13 +312,13 @@ func (b *sqlBuilder) writeInsert(inputCount, bulkRows int, columns []string) {
 	})
 	b.buf.WriteString(") VALUES ")
 	// Write out the values.
-	for i := 0; i < bulkRows; i++ {
+	for i, row := range rows {
 		if i != 0 {
 			b.buf.WriteString(", ")
 		}
 		b.buf.WriteString("(")
-		b.writeCommaSeperatedList(columns, func(j int, _ string) string {
-			return "@sqlair_" + strconv.Itoa(inputCount+(i*len(columns))+j)
+		b.writeCommaSeperatedList(row, func(_ int, value string) string {
+			return value
 		})
 		b.buf.WriteString(")")
 	}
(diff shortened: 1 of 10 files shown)


internal/typeinfo/valuelocator.go
```diff
@@ -134,16 +185,52 @@ func (f *structField) ArgType() reflect.Type {
...
+			// The slice has the correct type so there is no need to check the
+			// type of each element.
+			val := s.FieldByIndex(f.index)
+			if f.omitEmpty {
+				// If the omitemtpy flag is present, we expect either all rows to
+				// have a zero value, or all have a none zero value. If we have a
+				// mix then throw an error.
+				if i == 0 && val.IsZero() {
+					omit = true
+				} else if val.IsZero() && !omit || !val.IsZero() && omit {
```
letFunny: `val.IsZero() != omit`

internal/expr/parser.go
```diff
@@ -922,6 +930,26 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+func (p *Parser) parseColumnsInsertValues() ([]memberAccessor, bool, error) {
+	cp := p.save()
 	parenCol := p.colNum()
 	parenLine := p.lineNum
```
letFunny: These are redundant now.

internal/expr/bindinputs.go
```diff
@@ -45,47 +45,78 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+					if bulk {
+						if len(bc.vals) != numRows {
+							return nil, mismatchedBulkLengthsError(firstBulkColumn, numRows, bc.inputName, len(bc.vals))
+						}
+					} else {
+						bulk = true
+						firstBulkColumn = bc.inputName
+						numRows = len(bc.vals)
+					}
+				}
```
letFunny: Draft suggestion and questions:
```go
// do we need another error for when first column is NOT bulk but second one is? Right now bulk will be set on the second one?
	if bc.bulk && !bulk {
		// first row.
		bulk = true
		firstBulkColumn = bc.inputName
		numRows = len(bc.vals)
		continue
	}

	if bulk {
		if bc.bulk && len(bc.vals) != numRows {
			return nil, mismatchedBulkLengthsError(firstBulkColumn, numRows, bc.inputName, len(bc.vals))
		}
		if !bc.bulk {
			// error because all columns have to either be bulk or not?
		}
	}
```

internal/expr/parser.go
```diff
@@ -435,6 +417,32 @@ func (p *Parser) skipString(s string) bool {
...
+			continue
+		}
+
+		if p.char == ',' || p.char == ')' {
+			return true, nil
+		}
+		p.advanceChar()
+	}
+	return false, nil
+}
```
letFunny, line 444: This looks like a specific case of the parseUntil function we discussed months ago, maybe it will be helpful now.

internal/expr/parser.go
```diff
@@ -942,6 +970,57 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
+			// them.
+			if !inputParsed {
+				return nil, false, nil
+			}
+			return vs, true, nil
+		}
+
+		nextItem = p.skipChar(',')
+		p.skipBlanks()
+		itemStart = p.pos
```
letFunny: This can be moved to the top of the loop to remove L983 and L986 above and make it more uniform.

internal/expr/parser.go
```diff
@@ -942,6 +970,57 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
+		if p.skipChar(')') {
+			// If we only parsed literals, and not SQLair inputs then ignore
+			// them.
+			if !inputParsed {
+				return nil, false, nil
+			}
+			return vs, true, nil
+		}
+
+		nextItem = p.skipChar(',')
```
letFunny: I think normally you would do:
```go
if !p.skipChar(',') {
    break
}
```
and a for loop without condition on top.

letFunny (commented):
Skimmed through it, looks good. Waiting for the proper review.

internal/expr/bindinputs.go
Aflynn50: We have any some columns bulk and some not, the ones that are not are just duplicated

Aflynn50: I'll swap the order of the if/else, makes sense to have the first one first

internal/expr/parser.go
Aflynn50: True, I think we should add that if we add another similar function to this in the future. The `advanceToNextExpression` function where you originally suggested using a `parseUntil` function isn't quite such a good fit for it now, though feel free to bring a PR if you think it could be useful once this is in

letFunny: Only problem with that is that I doubt we are going to remember to refactor it if we need to create another one in the future. If we added it now it wouldn't add more than 5 lines (passing in the closure to get ',' or ')' and return the error).

SimonRichardson:
Out of curiosity more than anything. From the PR description, the SQL sent to the driver is:

```
INSERT INTO  t (col1, col2, col3, col4) VALUES (@sqlair_0, @sqlair_1, "v3", @sqalir_2), (@sqlair_3, @sqlair_4, "v3", @sqalir_5), (@sqlair_6, @sqlair_7, "v3", @sqalir_8)
```

Yet, `col2`, will always be a map for a constant value, so the query _could_ be turned into?

```
INSERT INTO  t (col1, col2, col3, col4) VALUES (@sqlair_0, @sqlair_1, "v3", @sqalir_2), (@sqlair_3, @sqlair_1, "v3", @sqalir_4), (@sqlair_5, @sqlair_1, "v3", @sqalir_6)
```

SimonRichardson (commented):
I'd love to see a full end to end test added to [package_test.go](https://github.com/canonical/sqlair/blob/main/package_test.go) that baked in the bulk insert changes.

I think the more common approach that I regularly want to use bulk insert in sqlair for is the following:

```
type S struct {
    a string `db:"col1"`
    b string `db:"col2"`
    c string `db:"col3"`
    d string `db:"col4"`
}

s := sqlair.MustPrepare(
    "INSERT INTO t (col1, col2, col3, col4) VALUES ($S.*)",
    S{},
)

db.Query(
    s,
    []S{
        {a: "a1", b: "b1", c: "c1", d: "d1"}, 
        {a: "a2", b: "b2", c: "c2", d: "d2"},
    },
)
```

Name | Value
--- | ---
sqlair_0 | "a1"
sqlair_1 | "b1"
sqlair_2 | "c1"
sqlair_3 | "d1"
sqlair_4 | "a2"
sqlair_5 | "b2"
sqlair_6 | "c2"
sqlair_7 | "d2"

internal/expr/parser.go
```diff
@@ -922,16 +930,36 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
+		return nil, false, err
+	} else if ok {
+		return &basicInsertExpr{columns: columns, sources: sources, raw: p.input[cp.pos:p.pos]}, true, nil
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseColumnsInsertValues parses insert expressions that are value on the
```
manadart: "that are _values_ ..."?

internal/expr/parser.go
```diff
@@ -942,6 +970,59 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
 		cp.restore()
 		return nil, false, nil
 	}
+	return sources, true, nil
+}
+
+// parseBasicInsertValues parses insert expressions that are value on the
```
manadart: Same here.

internal/typeinfo/valuelocator.go
```diff
@@ -32,9 +32,32 @@ type Input interface {
...
-	LocateParams(typeToValue TypeToValue) (vals []reflect.Value, omit bool, err error)
+	LocateParams(typeToValue TypeToValue) (params *Params, err error)
+}
+
+// Params contains query parameters and metadata generated from an input value.
+type Params struct {
+	// Vals contains the query parameters.
+	Vals []any
+	// Omit indicates if the params have an omitempty flag set and should be
+	// ommited from input.
```
manadart: "omitted".

internal/typeinfo/valuelocator.go
```diff
@@ -32,9 +32,32 @@ type Input interface {
...
 	// argument that are to be used in query parameters. An error is returned
 	// if typeToValue does not contain the input argument.
-	// The omit flag indicates if the params have an omitempty flag set and
-	// should be ommited from input.
-	LocateParams(typeToValue TypeToValue) (vals []reflect.Value, omit bool, err error)
+	LocateParams(typeToValue TypeToValue) (params *Params, err error)
+}
+
+// Params contains query parameters and metadata generated from an input value.
+type Params struct {
```
manadart, line 39: Good extraction.

internal/expr/bindinputs.go
```diff
@@ -45,47 +45,74 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
 				if err != nil {
 					return nil, err
 				}
-				if len(vals) > 1 {
-					// Only slices return multiple values and they are not
-					// allowed in insert expressions.
-					return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
+				bc.markArgUsed(argTypeUsed)
+
+				if bc.bulk {
```
manadart: These additions take the length of this method over 100 lines. We should seek ways to make it more terse.

manadart (commented):
Some comments from a partial review.

internal/expr/bindinputs.go
```diff
@@ -45,47 +45,74 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
+				namedInput := sql.Named("sqlair_"+strconv.Itoa(inputCount), val)
+				namedInputs = append(namedInputs, namedInput)
 				inputCount++
 			}
 		case *typedInsertExpr:
-			var columns []string
+			var boundColumns []*boundColumn
+			var columnNames []string
+			bulk := false
+			numRows := 1
```
letFunny: I think we should have a comment here explaining what / how this works. Maybe combine it with the comment below:
```
// bulk detects whether we want to do a bulk insert (numRows >= 1). If that is the case, firstBulkColumn stores...
```
What do you think? Because right now I feel that you need to read the whole thing to get an idea of how it works. Or maybe this is alleviated completely by the method refactor manadart@ suggested.

internal/expr/bindinputs.go
```diff
@@ -151,10 +185,115 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
...
+	// column is the column name.
+	column string
+}
+
+// markArgUsed marks the argument type passed to SQL as used in the argUsed map.
+func (bc *boundColumn) markArgUsed(argUsed map[reflect.Type]bool) {
+	if bc.argType != nil {
+		argUsed[bc.argType] = true
+	}
+}
```
letFunny: This method might be a little overkill seeing it is only used once.

internal/expr/bindinputs.go
```diff
@@ -45,47 +45,74 @@ func (tbe *TypeBoundExpr) BindInputs(args ...any) (pq *PrimedQuery, err error) {
...
-					return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
+				bc.markArgUsed(argTypeUsed)
+
+				if bc.bulk {
+					if !bulk {
+						// First bulk row.
+						bulk = true
+						firstBulkColumn = bc.inputName
+						numRows = len(bc.vals)
+					} else if len(bc.vals) != numRows {
```
letFunny: [Nit]: If this condition is relying on bulk == true, maybe it is better to put it explicitly or even as a separate if conditional, else you have to keep track of previous conditionals in your head.

internal/expr/bindinputs.go
letFunny: This first part of getting the columns and the inputs seems like a good fit.

internal/expr/bindinputs.go
```diff
@@ -132,7 +159,14 @@ type typedInputExpr struct {
 	input typeinfo.Input
 }
 
-// insertColumn stores information about a single column of a row in an INSERT
+// typedColumn represents a column and input locator in an insert statement.
```
letFunny, line 77: Can you add a `TODO` here or in the `columnAccessor` stating that we need to do a next to unify the column types?

internal/expr/bindinputs.go
```diff
@@ -151,10 +185,115 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
...
+	if err != nil {
+		return nil, err
+	}
+	if !params.Bulk && len(params.Vals) > 1 {
+		// Only slices and bulk inserts return multiple values and
+		// slices are not allowed in insert expressions.
+		return nil, fmt.Errorf("internal error: types in insert expressions cannot return multiple values")
+	}
+	if params.Omit {
+		if ic.explicit {
```
letFunny: [Nit]: combine with &&

internal/expr/bindinputs.go
```diff
@@ -151,10 +185,115 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
...
+	omit bool
+	// bulk is true if the list of values should be inserted in a bulk insert
+	// expression.
+	bulk bool
+	// argType is the type of the argument that was used to generate the
+	// params.
+	argType reflect.Type
+	// inputName is the type name of the input parameter.
+	inputName string
+	// literal set if the value to insert is a literal.
```
letFunny:
```suggestion
	// literal is set if the value to insert is a literal.
```

internal/expr/bindinputs.go
```diff
@@ -151,10 +185,115 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
...
+func newLiteralColumn(column, literal string) literalColumn {
+	return literalColumn{
+		column:  column,
+		literal: literal,
+	}
+}
+
+// boundColumn represents a column in an insert statement along with the values
+// to be inserted into it.
+type boundColumn struct {
```
letFunny: Nice change! it unifies all the omit and bulk logic into one type.

internal/expr/bindinputs.go
```diff
@@ -151,10 +185,115 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
...
+// parameter returns the value to be inserted into the boundColumn in the given
+// row. The inputCount it used to generate the parameter name.
+func (bc *boundColumn) parameter(row, inputCount int) (string, any, bool, error) {
+	if bc.literal != "" {
+		return bc.literal, nil, true, nil
+	}
+	name := "sqlair_" + strconv.Itoa(inputCount)
+	if !bc.bulk {
+		return "@" + name, sql.Named(name, bc.vals[0]), false, nil
+	}
```
letFunny: Do we need this extra check? or can we interpret regular inserts as bulk or row == 1?

internal/expr/bindinputs.go
```diff
@@ -170,18 +309,24 @@ type sqlBuilder struct {
 }
 
 // writeInsert writes the SQL for INSERT statements to the sqlBuilder.
-func (b *sqlBuilder) writeInsert(inputCount int, columns []string) {
+func (b *sqlBuilder) writeInsert(columns []string, rows [][]string) {
 	// Write out the columns.
 	b.buf.WriteString("(")
 	b.writeCommaSeperatedList(columns, func(_ int, column string) string {
```
letFunny: There is a typo in `Seperated`.

internal/expr/bindinputs.go
```diff
@@ -239,3 +384,20 @@ func markerIndex(s string) (int, bool) {
 func omitEmptyInputError(valueDesc string) error {
 	return fmt.Errorf("%s has zero value and has the omitempty flag but the value is explicitly input", valueDesc)
 }
+
+func mismatchedBulkLengthsError(firstBulkColumn string, expectedNumRows int, badTypeName string, badRowLength int) error {
+	return fmt.Errorf("different slices sizes in bulk insert: slice of %q has length %d but slice of %q has length %d",
```
letFunny: [Nit]: `slices sizes` is a little bit weird to read. What about `expected slices of same size in bulk insert: slice ....`?

internal/expr/bindtypes.go
```diff
@@ -259,6 +259,44 @@ func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err er
 	return &typedInsertExpr{insertColumns: cols}, nil
 }
 
+// basicInsertExpr is an input expression occurring within an INSERT statement
+// that consists of columns on the left and type accessors or literals on the
+// right. Unlike the columnInsertExpr, the values on the right are independent
+// of the columns on the left and are matched by position rather than by name.
+// e.g. (col1, col2, col3) VALUES ($M.key, "literal value", $T.value).
```
letFunny, line 271: 10/10 explanation.

internal/expr/bindtypes.go
```diff
@@ -259,6 +259,44 @@ func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err er
...
+// bindTypes generates a *typedInsertExpr containing type information about the
+// values to be inserted in the basicInsertExpr.
+func (e *basicInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("input expression: %s: %s", err, e.raw)
+		}
+	}()
+	if len(e.columns) != len(e.sources) {
+		return nil, fmt.Errorf("mismatched number of columns and values (%d!=%d)", len(e.columns), len(e.sources))
```
letFunny: [Nit]: I think normally we use a colon before the inequality in the error messages, something like `mismatched number of columns and values: %d != %d`.

internal/expr/bindtypes.go
```diff
@@ -378,6 +416,50 @@ func (e *outputExpr) bindTypes(argInfo typeinfo.ArgInfo) (te any, err error) {
 	return toe, nil
 }
 
+// valueAccessor defines an accessor that can be used to generate a typedColumn
+// with the given column name.
+type valueAccessor interface {
+	typedColumn(argInfo typeinfo.ArgInfo, columnName string) (typedColumn, error)
```
letFunny, line 429: If I am reading this correctly, `typedColumn` is only used for insert columns? (context: memberAccessor below returns a `newInsertColumn`.

internal/expr/bindtypes.go
```diff
@@ -259,6 +259,44 @@ func (e *columnsInsertExpr) bindTypes(argInfo typeinfo.ArgInfo) (tie any, err er
...
+// e.g. (col1, col2, col3) VALUES ($M.key, "literal value", $T.value).
+type basicInsertExpr struct {
+	columns []columnAccessor
+	sources []valueAccessor
+	raw     string
+}
+
+// String returns a text representation for debugging and testing purposes.
+func (e *basicInsertExpr) String() string {
+	return fmt.Sprintf("RenamingInsert[%v %v]", e.columns, e.sources)
```
letFunny: Maybe it would be useful to use commas to separate the sources, if you have an expression like `1 + 2 + 3` or even two words it might look like two separate sources instead of one.

internal/expr/expr_test.go
```diff
@@ -655,6 +668,70 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+}, {
+	summary:        "bulk insert with explicit columns and constants",
+	query:          `INSERT INTO person (name, mapKey, street) VALUES ($Person.*, $Address.*, $M.*)`,
+	expectedParsed: `[Bypass[INSERT INTO person ] ColumnInsert[[name mapKey street] [Person.* Address.* M.*]]]`,
+	typeSamples:    []any{Person{}, Address{}, M{}},
+	inputArgs:      []any{[]Person{{Fullname: "Al"}, {Fullname: "Albert"}}, M{"mapKey": "const map"}, Address{Street: "const street"}},
+	expectedParams: []any{"Al", "const map", "const street", "Albert", "const map", "const street"},
+	expectedSQL:    `INSERT INTO person (name, mapKey, street) VALUES (@sqlair_0, @sqlair_1, @sqlair_2), (@sqlair_3, @sqlair_4, @sqlair_5)`,
+}, {
+	summary:        "bulk insert with omit empty struct not inserted at all",
```
letFunny, line 681: Can you add a test case for a slice of `OmitEmptyID` where they are not omitted? And another one with mixed values, zero and non-zero?

internal/expr/expr_test.go
```diff
@@ -655,6 +668,70 @@ AND z = @sqlair_0 -- The line with $Person.id on it
...
+}, {
+	summary:        "bulk insert with slices of length 1",
+	query:          `INSERT INTO person (*) VALUES ($Person.id, $M.key, $Address.street)`,
+	expectedParsed: `[Bypass[INSERT INTO person ] AsteriskInsert[[*] [Person.id M.key Address.street]]]`,
+	typeSamples:    []any{Person{}, M{}, Address{}},
+	inputArgs:      []any{[]Person{{ID: 1}}, M{"key": "val"}, []Address{{Street: "Church Road"}}},
+	expectedParams: []any{1, "val", "Church Road"},
+	expectedSQL:    `INSERT INTO person (id, key, street) VALUES (@sqlair_0, @sqlair_1, @sqlair_2)`,
+}, {
+	summary:        "bulk insert with explicit columns and literal",
```
letFunny, line 721: This test is the same as the one below, right?

internal/expr/parser.go
```diff
@@ -922,16 +930,36 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
-	parenCol := p.colNum()
-	parenLine := p.lineNum
+	if sources, ok, err := p.parseColumnsInsertValues(); err != nil {
+		cp.restore()
+		return nil, false, err
+	} else if ok {
+		return &columnsInsertExpr{columns: columns, sources: sources, raw: p.input[cp.pos:p.pos]}, true, nil
+	}
+
+	if sources, ok, err := p.parseBasicInsertValues(); err != nil {
```
letFunny, line 937: It is a little bit weird that `parseColumnsInsertExpr` has a `parseColumnsInsertValues` and a `parseBasicInsertValues`, it seems like the first one matches the intent and the second one does not belong here naming-wise.

internal/expr/parser.go
```diff
@@ -922,16 +930,36 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
...
+	} else if ok {
+		return &basicInsertExpr{columns: columns, sources: sources, raw: p.input[cp.pos:p.pos]}, true, nil
+	}
+
+	cp.restore()
+	return nil, false, nil
+}
+
+// parseColumnsInsertValues parses insert expressions that are value on the
+// right hand side of a columns insert expression.
```
letFunny: Can you state what the expression is instead of `columns insert expression`. I find that confusing as the outer parse function is named the same and there is no clear indication what the difference is.

internal/expr/parser.go
```diff
@@ -942,6 +970,59 @@ func (p *Parser) parseColumnsInsertExpr() (expression, bool, error) {
 		cp.restore()
 		return nil, false, nil
 	}
+	return sources, true, nil
+}
+
+// parseBasicInsertValues parses insert expressions that are value on the
+// right hand side of a basic insert expression.
```
letFunny: Ditto for `basic insert expression`.

internal/expr/parser_helper_test.go
```diff
@@ -306,3 +306,49 @@ func (s parseSuite) TestAdvanceToNextExpr(c *C) {
...
+	for _, t := range tests {
+		p.init(t.input)
+		vs, ok, err := p.parseBasicInsertValues()
+		c.Assert(err, IsNil)
+		c.Assert(ok, Equals, true)
+		c.Assert(vs, DeepEquals, t.expected)
+	}
+}
+
+func (s parseSuite) TestParseRenamingInsertValuesFalse(c *C) {
```
letFunny, line 348: Should we also cover the corner cases like `()`, `("literalNoSpaces")`? These proved troublesome in the past.

internal/typeinfo/valuelocator.go
```diff
@@ -65,20 +88,49 @@ func (mk *mapKey) ArgType() reflect.Type {
...
+	var argType reflect.Type
+	var vals []any
+	if m, ok := typeToValue[mk.mapType]; ok {
+		v := m.MapIndex(reflect.ValueOf(mk.name))
+		if v.Kind() == reflect.Invalid {
+			return nil, fmt.Errorf("map %q does not contain key %q", mk.mapType.Name(), mk.name)
+		}
+		argType = m.Type()
+		vals = append(vals, v.Interface())
+	} else if ms, ok := locateBulkType(typeToValue, mk.mapType); ok {
```
letFunny: [Nit]:
> In general prefer early returns and separate conditionals

there is no need for `bulk` if you return early.

internal/typeinfo/valuelocator.go
```diff
@@ -131,19 +183,55 @@ func (f *structField) ArgType() reflect.Type {
...
-	}
-	val := s.FieldByIndex(f.index)
-	if val.IsZero() && f.omitEmpty {
-		omitempty = true
+// LocateParams locates the struct (or slice of structs for a bulk insert) that
+// contains the field in the TypeToValue map. It returns Params containing the
+// value of this field.
+func (f *structField) LocateParams(typeToValue TypeToValue) (*Params, error) {
+	omit := false
+	bulk := false
```
letFunny: [Nit]: Same nit about early returns and flags, that would avoid one level of conditionals and make it more explicit.

internal/typeinfo/valuelocator.go
```diff
@@ -131,19 +183,55 @@ func (f *structField) ArgType() reflect.Type {
...
+			// type of each element.
+			val := s.FieldByIndex(f.index)
+			if f.omitEmpty {
+				// If the omitemtpy flag is present, we expect either all rows to
+				// have a zero value, or all have a none zero value. If we have a
+				// mix then throw an error.
+				if i == 0 && val.IsZero() {
+					omit = true
+				} else if val.IsZero() != omit {
+					return nil, fmt.Errorf("got mix of zero and none zero values in %s which has the omitempty flag set, in a bulk insert, values must be all zero or all none zero", f.Desc())
```
letFunny, line 224: This test case is missing.

letFunny (commented):
One test file to go, I have reviewed the rest in depth. Changes look very good and the architecture and organization looks great. Thanks!

I have added some comments mainly about nitpicks, test cases and small improvements to readability. I don't want to block the PR so I will approve once I review the last file.

internal/typeinfo/valuelocator_test.go
```diff
@@ -127,162 +127,266 @@ func (s *typeInfoSuite) TestLocateScanTargetError(c *C) {
...
+		},
+		err: `got nil map in slice of "M" at position 1`,
+	}, {
+		summary:    "map bulk insert nil pointer to map in slice",
+		typeSample: M{},
+		arg:        []*M{{"foo": "bar"}, nil},
+		input: func(ai ArgInfo) (Input, error) {
+			return ai.InputMember("M", "foo")
+		},
+		err: `got nil pointer in slice of "M" at position 1`,
```
letFunny: Should we say `index` instead of position? Index to me starts at 0 clearly.

internal/typeinfo/valuelocator_test.go
```diff
@@ -127,162 +127,266 @@ func (s *typeInfoSuite) TestLocateScanTargetError(c *C) {
...
-}
-
-func (*typeInfoSuite) TestLocateParamsSlice(c *C) {
-	type S []any
-	type T []int
-
-	argInfo, err := GenerateArgInfo([]any{S{}, T{}})
-	c.Assert(err, IsNil)
-
+func (s *typeInfoSuite) TestLocateParams(c *C) {
```
letFunny, line 140: Nice refactor!

letFunny (approved):
Thanks

internal/expr/bindinputs.go
Aflynn50: I've added a TODO to the `parseInsertColumns` to say that it needs to return the correct column type. I'm not sure we want to unify the parser column types and the "typed" column types.

internal/expr/bindinputs.go
Aflynn50: Fair, I've inlined it

internal/expr/bindinputs.go
Aflynn50: Changed, I've made this whole method a switch case on len(bc.vals) instead

internal/expr/bindinputs.go
Aflynn50: Good spot, changed

internal/expr/bindtypes.go
Aflynn50: True, its a bit confusing, but this is only for our tests and will make the diff even worse for this PR.

internal/expr/bindtypes.go
Aflynn50: Changed, and added tests for this error

internal/expr/bindtypes.go
Aflynn50: And literals

internal/expr/expr_test.go
Aflynn50: Added

internal/expr/bindtypes.go
letFunny: literals in insert expressions, right?

internal/expr/expr_test.go
Aflynn50: Yea pretty much, the second is a more complex case of the first, though I'm inclined to keep both, because these also serve as examples to people looking at the library. 

I've removed the third insert value from the first one so that it at least tests a literal at the end.

internal/expr/parser.go
Aflynn50: I'd rather leave it as it is, this way the name of the function documents exactly what it does.

internal/expr/parser.go
Aflynn50: Yea, I've just redone this whole section, there was a load of redundancy and confusion there

internal/expr/parser_helper_test.go
Aflynn50: Added

internal/typeinfo/valuelocator_test.go
Aflynn50: Changed

internal/expr/bindtypes.go
Aflynn50: Yea, right now the valueAccessor is implemented by `memberAccessor` and `literal` the typedColumn interface is implemented by `literalColumn` and `insertColumn`.

Its not the best abstraction, but I was struggling to find a nice way to do literals.

Aflynn50:
> Out of curiosity more than anything. From the PR description, the SQL sent to the driver is:
> 
> ```
> INSERT INTO  t (col1, col2, col3, col4) VALUES (@sqlair_0, @sqlair_1, "v3", @sqalir_2), (@sqlair_3, @sqlair_4, "v3", @sqalir_5), (@sqlair_6, @sqlair_7, "v3", @sqalir_8)
> ```
> 
> Yet, `col2`, will always be a map for a constant value, so the query _could_ be turned into?
> 
> ```
> INSERT INTO  t (col1, col2, col3, col4) VALUES (@sqlair_0, @sqlair_1, "v3", @sqalir_2), (@sqlair_3, @sqlair_1, "v3", @sqalir_4), (@sqlair_5, @sqlair_1, "v3", @sqalir_6)
> ```

Thats a good point, no reason not to do this. I've changed the code and updated the PR description to reflect this.

Aflynn50:
> I'd love to see a full end to end test added to [package_test.go](https://github.com/canonical/sqlair/blob/main/package_test.go) that baked in the bulk insert changes.
> 
> I think the more common approach that I regularly want to use bulk insert in sqlair for is the following:
> 
> ```
> type S struct {
>     a string `db:"col1"`
>     b string `db:"col2"`
>     c string `db:"col3"`
>     d string `db:"col4"`
> }
> 
> s := sqlair.MustPrepare(
>     "INSERT INTO t (col1, col2, col3, col4) VALUES ($S.*)",
>     S{},
> )
> 
> db.Query(
>     s,
>     []S{
>         {a: "a1", b: "b1", c: "c1", d: "d1"}, 
>         {a: "a2", b: "b2", c: "c2", d: "d2"},
>     },
> )
> ```
> 
> Name 	Value
> sqlair_0 	"a1"
> sqlair_1 	"b1"
> sqlair_2 	"c1"
> sqlair_3 	"d1"
> sqlair_4 	"a2"
> sqlair_5 	"b2"
> sqlair_6 	"c2"
> sqlair_7 	"d2"

Good point, I'd been meaning to add something to package_test but had forgotten. I've added a test now that has a simple case and a more complex case of a bulk insert.

internal/expr/bindinputs.go
Aflynn50: I've refactored this method into a new `queryBuilder`. Then used the visitor pattern with the `typedExpr` to call the add methods on it.

internal/expr/bindinputs.go
```diff
@@ -151,10 +106,121 @@ func newInsertColumn(input typeinfo.Input, column string, explicit bool) insertC
 	}
 }
 
+// bindInputs generates and verifies the query parameters corresponding to the
```
manadart, line 161: Just above here on line 95, there's a typo "explicity".

internal/expr/bindinputs.go
```diff
@@ -126,13 +65,29 @@ func newOutputColumn(tableName string, columnName string, output typeinfo.Output
 	return outputColumn{column: tableName + "." + columnName, output: output}
 }
 
+type typedExpr interface {
```
manadart: I'd locate this at the top, before it's first usage.

internal/expr/bindinputs.go
```diff
@@ -126,13 +65,29 @@ func newOutputColumn(tableName string, columnName string, output typeinfo.Output
...
 // typedInputExpr stores information about a Go value to use as a standalone query
 // input.
 type typedInputExpr struct {
 	input typeinfo.Input
 }
 
-// insertColumn stores information about a single column of a row in an INSERT
+// addToQueryBuilder adds the typed input expressions to the query builder.
+func (te *typedInputExpr) addToQueryBuilder(qb *queryBuilder) error {
+	return qb.addInput(te)
```
manadart: This is not really _visitor_, because `queryBuilder` holds all the logic, and if you add another `typedExpr` implementation that it needs to handle, it grows.

That said, this _is_ managing complexity after a fashion, though you've tightly coupled the builder and the expressions.

manadart: One way to do it would be for `queryBuilder` to implement an interface that exposes what the visitors need, then push the logic back into the visitors. Looking at:
```go
func (qb *queryBuilder) addInput(te *typedInputExpr) error {
	params, err := te.input.LocateParams(qb.typeToValue)
	if err != nil {
		return err
	}
	if params.Omit {
		return omitEmptyInputError(te.input.Desc())
	}
	if params.Bulk {
		return fmt.Errorf("cannot use bulk inputs outside an insert statement")
	}

	qb.argUsed[params.ArgTypeUsed] = true
	qb.sqlBuilder.writeInput(qb.inputCount, len(params.Vals))
	for _, val := range params.Vals {
		namedInput := sql.Named("sqlair_"+strconv.Itoa(qb.inputCount), val)
		qb.namedInputs = append(qb.namedInputs, namedInput)
		qb.inputCount++
	}
	return nil
}
```
The first half is validation that could (and should) be on `typedInputExpr`.
The rest could be done by coordinating methods like `setArgTypeUsed`, `writeInput`, `addNamedInput` or some such.

manadart (approved):
This is worth landing to get the feature. 

I would advocate thinking about how to refactor the `queryBuilder` interactions. I'll ruminate on this myself.

internal/expr/querybuilder.go
```diff
@@ -122,16 +114,28 @@ func (qb *queryBuilder) checkAllArgsUsed(typeToValue typeinfo.TypeToValue) error
 	return nil
 }
 
+// inputAssigner assigns input numbers to input expressions.
+type inputAssigner struct {
+	inputCount int
+}
```
letFunny: To me this looks like a code smell, that probably there is too much abstraction for the given complexity. In my opinion, what you had at https://github.com/canonical/sqlair/pull/145/commits/afade243d0d0a68049dccfe4167e009272f92f9a was great, it moved the logic to encapsulate it and made the code clearer and more self-contained.

We now seem to be marching towards more encapsulation and decoupling, which is a worthwhile goal don't get me wrong. However, to me the results look a little bit too complex. Maybe we could wait until we need to add something else to the code and then do the refactor once we have new concrete use-cases.

Aflynn50: As discussed in person, we'll agree to disagree. In my opinion this is not adding complexity, it is making explicit complexity that is already there. The `inputAssigner` allows the `bindInputs` method on the `insertColumn` to take care of assigning the input numbers for the `boundInsertColumn` to use. Before it was being done by the caller based on the return values which (in a very minor way) broke the encapsulation of the the `boundColumn`.

This also makes changes to the way inputs are assigned a lot easier. We just change the `inputAssigner` struct and then follow through the compiler errors where `assignInputs` is called.

PR #149, opened 2024-05-07
feat: add a cache for driver prepared statements

Aflynn50:
When the same SQLair query is executed multiple times against the database it is recompiled by the database every time. This is inefficient. 

It is possible to instead prepare a SQL query against the database driver and cache it in SQLair to allow for reuse of the query.

With this change, when a SQLair statement is executed directly against a database it is first prepared and then executed, the prepared statement is put in the cache for reuse. If the statement is executed in a transaction, the SQLair statement is looked up in the cache but if it is not found it is not prepared on the driver. This is due to the issues in #117.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -13,24 +13,37 @@ import (
 	"github.com/canonical/sqlair/internal/expr"
 )
 
-// statementCache caches the driver prepared sql.Stmt objects associated with
-// each sqlair.Statement. A sqlair.Statement can correspond to multiple sql.Stmt
-// objects prepared on different databases. Entries in the cache are therefore
-// indexed by the sqlair.Statement cache ID and the sqlair.DB cache ID.
+// statementCache caches driver prepared sql.Stmt objects associated with
+// sqlair.Statement objects. The driver prepared sql.Stmt objects corresponding
+// to a single sqlair.Statement can be prepared on different database and have
+// different generated SQL (depending on query arguments).
+//
+// Entries in the cache are indexed by the sqlair.Statement cache ID and the
+// sqlair.DB cache ID. Each entry contains the sql.Stmt and the SQL string that
+// was used to create it.
 //
 // A finalizer is set on sqlair.Statement objects to close the associated
-// sql.Stmt objects. Similarly, a finalizer is set on sqlair.DB objects to close
-// all sql.Stmt objects prepared on the DB, close the DB, and remove the DB
-// cache ID from the cache.
+// sql.Stmt objects and remove them from the cache. Similarly, a finalizer is
+// set on sqlair.DB objects to close all sql.Stmt objects prepared on the DB,
+// close the DB, and remove the DB cache ID from the cache.
+//
+// Only a single driver prepared sql.Stmt is cached for each sqlair.DB/
+// sqlair.Statement pair. If the sqlair.Statement is re-prepared with different
+// generated SQL then the previous sql.Stmt is evicted from the cache. A
+// finalizer is set on the evicted sql.Stmt to ensure it is closed once all
+// references die.
 //
 // The mutex must be locked when accessing either the stmtDBCache or the
 // dbStmtCache.
 type statementCache struct {
-	// stmtDBCache stores sql.Stmt objects addressed via the cache ID of the
-	// sqlair.Statement they built from and the sqlair.DB they are prepared on.
-	stmtDBCache map[uint64]map[uint64]*sql.Stmt
+	// stmtDBCache stores driverStmt objects containing a sql.Stmt and the sql
+	// used to generate it. These are addressed via the cache ID of the
+	// corresponding sqlair.Statement and the cache ID of the sqlair.DB they are
+	// prepared against.
+	stmtDBCache map[uint64]map[uint64]*driverStmt
 
-	// dbStmtCache indicates when a sqlair.Statement has been prepared on a particular sqlair.DB.
+	// dbStmtCache indicates when a sqlair.Statement has been prepared on a
+	// particular sqlair.DB.
 	dbStmtCache map[uint64]map[uint64]bool
 
 	// stmtIDCount and dbIDCount are monotonically increasing counters used to
@@ -41,6 +54,12 @@ type statementCache struct {
 	mutex sync.RWMutex
 }
 
+// driverStmt represents a SQL statement prepared against a database driver.
+type driverStmt struct {
+	stmt *sql.Stmt
+	sql  string
+}
+
 var once sync.Once
 var singleStmtCache *statementCache
 
@@ -48,7 +67,7 @@ var singleStmtCache *statementCache
 func newStatementCache() *statementCache {
 	once.Do(func() {
 		singleStmtCache = &statementCache{
-			stmtDBCache: map[uint64]map[uint64]*sql.Stmt{},
+			stmtDBCache: map[uint64]map[uint64]*driverStmt{},
 			dbStmtCache: map[uint64]map[uint64]bool{},
 		}
 	})
@@ -61,7 +80,7 @@ func newStatementCache() *statementCache {
 func (sc *statementCache) newStatement(te *expr.TypeBoundExpr) *Statement {
 	cacheID := atomic.AddUint64(&sc.stmtIDCount, 1)
 	sc.mutex.Lock()
-	sc.stmtDBCache[cacheID] = map[uint64]*sql.Stmt{}
+	sc.stmtDBCache[cacheID] = map[uint64]*driverStmt{}
 	sc.mutex.Unlock()
 	s := &Statement{te: te, cacheID: cacheID}
 	// This finalizer is run after the Statement is garbage collected.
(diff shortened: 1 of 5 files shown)


cache.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// storeStmt stores a prepared *sql.Stmt in the cache.
+func (sc *statementCache) storeStmt(db *DB, s *Statement, sqlstmt *sql.Stmt) {
+	sc.mutex.Lock()
+	// Check if sqlstmt has been inserted by another process else already.
+	sqlstmtAlt, ok := sc.stmtDBCache[s.cacheID][db.cacheID]
+	if ok {
+		sqlstmt.Close()
+		sqlstmt = sqlstmtAlt
```
SimonRichardson: How are you testing this? 

If you want `sqlstmtAlt` to point to `sqlstmt` then you either need to write to the pointer `*sqlstmt` or return the `sql.Stmt`.

Aflynn50: Ah good point, that's just closing the original statement and doing nothing else. I've refactored this function to actually prepare the statement as well to simplify this a bit.

This is not being tested at the moment, the difficult thing is, this case is only triggered by a race. In general we haven't been testing for race issues and relying on inspection, but this was still very much missed.

We could add in more infrastructure to the test driver to test this sort of thing, adding artificial waits and so on, but for now I'm not that keen to put in the work to add it unless we start coming across issues with this elsewhere.

SimonRichardson: That's better. You're still doing work to then throw it away if another goroutine got there first. Although this is likely to be an edge case, I wonder if this is still the best approach, or if another strategy can be used?

Aflynn50: imo this is very much an edge case, it has to be the same statement being used separately in two different places, i.e. two instances of exactly the same SQL query happening at the same time. Plus this code is only going to be hit if its the first time they are being prepared, any subsequent times the lookup will find the sql.Stmt in the cache and this function won't even be called.

An obvious alternative would just be to lock the cache for the whole preparation as well but that has a far higher cost since its a global cache and it will block all queries whilst the cache is locked and it will happen fairly often since running a statement for the first time on the database is a relatively common action.

sqlair.go
```diff
@@ -453,6 +477,21 @@ func (tx *TX) Query(ctx context.Context, s *Statement, inputArgs ...any) *Query
...
+			// The txstmt is closed by database/sql when the transaction is
+			// commited or rolled back.
+			txstmt := tx.sqltx.Stmt(sqlstmt)
+			if pq.HasOutputs() {
+				rows, err = txstmt.QueryContext(innerCtx, pq.Params()...)
+			} else {
+				result, err = txstmt.ExecContext(innerCtx, pq.Params()...)
+			}
+			return rows, result, err
+		}
```
manadart, line 493: Under the original https://github.com/canonical/sqlair/pull/81 we were re-preparing each time because the statement was closed with the transaction.

In attempting to fix this under https://github.com/canonical/sqlair/pull/117, it was noted that preparing against the transaction's database required an extra connection to be acquired, which opened the possibility of deadlock for situations where [SetMaxOpenConns](https://pkg.go.dev/database/sql#DB.SetMaxOpenConns) had been called with a some number > 0.

Why don't we check the `DBStats` here and do this preparation if `MaxOpenConnections` is zero? Juju would then be able to benefit from the statement caching.

cache.go
```diff
@@ -0,0 +1,169 @@
...
+	"runtime"
+	"sync"
+	"sync/atomic"
+
+	"github.com/canonical/sqlair/internal/expr"
+)
+
+// statementCache caches driver prepared sql.Stmt objects associated with
+// sqlair.Statement objects. The driver prepared sql.Stmt objects corresponding
+// to a single sqlair.Statement can be prepared on different database and have
```
manadart: "different databases"

sqlair.go
```diff
@@ -35,9 +35,16 @@ type S []any
 var ErrNoRows = sql.ErrNoRows
 var ErrTXDone = sql.ErrTxDone
 
+// stmtCache stores the driver prepared statements associated to the SQLair
```
manadart, line 38: I would change all these "driver prepared statements" to "driver-prepared statements".

cache.go
```diff
@@ -0,0 +1,169 @@
...
+	// This finalizer is run after the DB is garbage collected.
+	runtime.SetFinalizer(db, sc.removeAndCloseDBFunc)
+	return db
+}
+
+// lookupStmt checks if a Statement has been prepared on the db driver with the
+// given primedSQL. If it has, the driver prepared sql.Stmt is returned.
+func (sc *statementCache) lookupStmt(db *DB, s *Statement, primedSQL string) (stmt *sql.Stmt, ok bool) {
+	// The Statement cache ID is only removed from stmtDBCache when the
+	// finalizer is run. The Statements cache ID must be in the stmtDBCache
```
manadart: "Statement's".

manadart (commented):
This is faithful to the original approved by Gustavo, but I'd like to see if we can get the benefit in (transaction-only) Juju.

sqlair.go
Aflynn50: We did consider that, but it feels very unsafe in regard to concurrency. For one, the SetMaxOpenConns could be set between us reading the DBstats and trying to open the connection. Secondly we do not control the `sql.db`. When the `sqlair.DB` is created we just use the `sql.db` passed to it. It is possible for another reference of that to be held separately and for someone to call `MaxOpenConns` on that without our knowledge.

In my opinion, without SQLair fully controlling the DB (i.e. having our own `Open` and not having a `PlainDB` method), we're still open to some obscure deadlock scenarios.

manadart (approved):
Let's land this.

I added this to domain/secret/state/state_test.go in Juju:
```go
func (s *stateSuite) BenchmarkSomeStatements(c *gc.C) {
	st := newSecretState(c, s.TxnRunnerFactory())

	s.setupUnits(c, "mysql")

	sp := []domainsecret.UpsertSecretParams{{
		Description: ptr("my secretMetadata"),
		Label:       ptr("my label"),
		Data:        coresecrets.SecretData{"foo": "bar"},
		AutoPrune:   ptr(true),
	}, {
		Description: ptr("my secretMetadata2"),
		Label:       ptr("my label2"),
		Data:        coresecrets.SecretData{"foo": "bar2"},
	}}
	uri := []*coresecrets.URI{
		coresecrets.NewURI(),
		coresecrets.NewURI(),
	}

	ctx := context.Background()
	err := st.CreateUserSecret(ctx, 1, uri[0], sp[0])
	c.Assert(err, jc.ErrorIsNil)
	err = st.CreateCharmApplicationSecret(ctx, 1, uri[1], "mysql", sp[1])
	c.Assert(err, jc.ErrorIsNil)

	for i := 0; i < c.N; i++ {
		_, _, err := st.ListCharmSecrets(ctx, domainsecret.ApplicationOwners{"mysql"}, domainsecret.NilUnitOwners)
		c.Assert(err, jc.ErrorIsNil)

		_, err = st.ListUserSecretsToDrain(ctx)
		c.Assert(err, jc.ErrorIsNil)

		_, err = st.GetSecretGrants(ctx, uri[0], coresecrets.RoleView)
		c.Assert(err, jc.ErrorIsNil)

		_, _, err = st.GetSecretValue(ctx, uri[1], 1)
		c.Assert(err, jc.ErrorIsNil)
	}
}
```

Then ran this suite with `-check.b`.

Without your patch:
```
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2874646 ns/op
OK: 1 passed
--- PASS: TestPackage (2.63s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2766026 ns/op
OK: 1 passed
--- PASS: TestPackage (2.71s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2725160 ns/op
OK: 1 passed
--- PASS: TestPackage (2.62s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2735487 ns/op
OK: 1 passed
--- PASS: TestPackage (2.57s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2842151 ns/op
OK: 1 passed
--- PASS: TestPackage (2.61s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2698437 ns/op
OK: 1 passed
--- PASS: TestPackage (2.62s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2713350 ns/op
OK: 1 passed
--- PASS: TestPackage (2.56s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2481026 ns/op
OK: 1 passed
--- PASS: TestPackage (2.46s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2760748 ns/op
OK: 1 passed
--- PASS: TestPackage (2.64s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2714200 ns/op
OK: 1 passed
--- PASS: TestPackage (2.53s)
PASS
ok      github.com/juju/juju/domain/secret/state        25.953s
```

With your patch:
```
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2773115 ns/op
OK: 1 passed
--- PASS: TestPackage (2.68s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2869826 ns/op
OK: 1 passed
--- PASS: TestPackage (2.68s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2448388 ns/op
OK: 1 passed
--- PASS: TestPackage (2.38s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2509037 ns/op
OK: 1 passed
--- PASS: TestPackage (2.44s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2587446 ns/op
OK: 1 passed
--- PASS: TestPackage (2.50s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2687746 ns/op
OK: 1 passed
--- PASS: TestPackage (2.64s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2490027 ns/op
OK: 1 passed
--- PASS: TestPackage (2.47s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements        1000           2558516 ns/op
OK: 1 passed
--- PASS: TestPackage (3.75s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2945531 ns/op
OK: 1 passed
--- PASS: TestPackage (3.06s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2338716 ns/op
OK: 1 passed
--- PASS: TestPackage (2.62s)
PASS
ok      github.com/juju/juju/domain/secret/state        27.238s
```

With your patch, plus preparing against the transaction's DB:
```
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2898581 ns/op
OK: 1 passed
--- PASS: TestPackage (2.72s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements        1000           2540134 ns/op
OK: 1 passed
--- PASS: TestPackage (3.74s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements        1000           2485494 ns/op
OK: 1 passed
--- PASS: TestPackage (3.66s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           3061992 ns/op
OK: 1 passed
--- PASS: TestPackage (2.70s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2893728 ns/op
OK: 1 passed
--- PASS: TestPackage (2.69s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2699059 ns/op
OK: 1 passed
--- PASS: TestPackage (2.54s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2736241 ns/op
OK: 1 passed
--- PASS: TestPackage (2.58s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           3147431 ns/op
OK: 1 passed
--- PASS: TestPackage (2.80s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements        1000           2554111 ns/op
OK: 1 passed
--- PASS: TestPackage (3.68s)
=== RUN   TestPackage
PASS: state_test.go:1005: stateSuite.BenchmarkSomeStatements         500           2281511 ns/op
OK: 1 passed
--- PASS: TestPackage (2.27s)
PASS
ok      github.com/juju/juju/domain/secret/state        29.405s
```

We're not round-tripping to the DB leader over the network of course, but it also doesn't appear that there's any particular help or hurt (though times do dilate slightly).

Aflynn50:
Not too surprising that the times go up slightly. Because of the issue with transactions that meant we couldn't cache their statements Juju does not benefit at all from this caching.

PR #153, opened 2024-06-12
fix: add error for unused types in prepare and refactor bindtypes/arginfo 

Aflynn50:
When doing `sqlair.Prepare`, SQLair will not throw an error if you pass extra types that are not used in the query. This refactor is motivated by the need to add this error.

The `BindTypes` function used `ArgInfo` from the `typeinfo` package to get information about the arguments. The `ArgInfo` object has methods that allow `BindTypes` to pass simply the name of the type needed and they will return all the relevent information about the type or an error message if it is not found. The `typeinfo` package is only concerned with getting information about types, and does not do any storing of state.

A simple (but wrong) way to track which types have been used during prepare would be to add state to this `ArgInfo` object, and when `BindTypes` was finished, it could be queried to see if all types had been used. However, this would add state to `ArgInfo`

Another way to do this would be to keep track of the arguments used in `BindTypes` directly, however, the level of abstraction here is wrong, several layers have to be unwrapped to get the type information.

The solution I have landed upon is adding an extra layer in-between `ArgInfo` and the `bindTypes` functions on the typed expressions. This follows a similar pattern to the query builder in the `BindInputs` stage. The `ArgInfo` object arguably had too much responsibility considering that it was designed only to be concerned with types. Therefore, it was natural to transfer some of this to the new `typedExprBulider` layer which is now responsible for finding the arguments in the map returned from `GenerateArgInfo` and also with the keeping track of outputs and arguments used.

### DIFF
--- a/internal/expr/bindtypes.go
+++ b/internal/expr/bindtypes.go
@@ -55,11 +55,7 @@ func (pe *ParsedExpr) BindTypes(args ...any) (tbe *TypeBoundExpr, err error) {
 		}
 	}
 
-	if err := teb.checkAllArgsUsed(argInfo); err != nil {
-		return nil, err
-	}
-
-	return &TypeBoundExpr{typedExprs: teb.typedExprs}, nil
+	return teb.Build()
 }
 
 // expression represents a parsed node of the SQLair query's AST.
@@ -67,7 +63,8 @@ type expression interface {
 	// String returns a text representation for debugging and testing purposes.
 	String() string
 
-	// bindTypes binds the types to the expression to generate a typedExpr.
+	// bindTypes binds the types to the expression to generate a typedExpr and
+	// adds it to the typedExprBuilder.
 	bindTypes(*typedExprBuilder) error
 }
 
@@ -82,8 +79,7 @@ func (b *bypass) String() string {
 	return "Bypass[" + b.chunk + "]"
 }
 
-// bindTypes returns the bypass part itself since it contains no references to
-// types.
+// bindTypes adds the bypass part, unchanged, to the typedExprBuilder.
 func (b *bypass) bindTypes(teb *typedExprBuilder) error {
 	teb.AddBypass(b)
 	return nil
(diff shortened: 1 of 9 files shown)


internal/typeinfo/arginfo.go
```diff
@@ -13,18 +13,10 @@ import (
...
-// ArgInfo is used to access type information about SQLair input and output
-// arguments. Methods on ArgInfo can be used to generate input and output value
-// locators.
-//
-// ArgInfo should only be accessed using it methods, not used directly as a
-// map.
-type ArgInfo map[string]arg
-
 // GenerateArgInfo takes sample instantiations of argument types and uses
 // reflection to generate an ArgInfo containing the types.
```
manadart: It's now generating a map of them, which needs to be explained by a change to this comment.

internal/typeinfo/arginfo.go
```diff
@@ -55,175 +47,116 @@ func GenerateArgInfo(typeSamples []any) (ArgInfo, error) {
...
 
-// Kind looks up the type name and returns its kind.
-func (argInfo ArgInfo) Kind(typeName string) (reflect.Kind, error) {
-	arg, ok := argInfo[typeName]
-	if !ok {
-		return 0, nameNotFoundError(argInfo, typeName)
-	}
-	return arg.typ().Kind(), nil
+// ArgInfo exposes useful information about SQLair input/output argument types.
+type ArgInfo interface {
```
manadart: Put this at the top, before its first use.

internal/expr/bindtypes.go
```diff
@@ -93,8 +84,9 @@ func (b *bypass) String() string {
 
 // bindTypes returns the bypass part itself since it contains no references to
```
manadart: Here and other places, you need to alter the comments around the fact that there are no returns, rather accumulations in the `typedExprBuilder`.

manadart (commented):
Couple of nits to address, but this is a solid change.

Also, you've an unused function `checkDuplicate` in _internal/typeinfo/validate.go_ that you might as well remove with this change.

internal/expr/bindtypes.go
Aflynn50: Added them.

Aflynn50:
I've just added a `Build` method to the `typedExprBuilder`. It occurred to me that its a bit weird that the final building of it is the only part we do manually in `BindTypes`.

manadart (approved):
Thanks. Good change.

PR #157, opened 2024-07-12
fix: format in read the docs yaml

Aflynn50:
There was an issue the the read the docs yaml that was only made apparent once the build process started. This fixes it.

The option under post_checkout had to be a list not a string.


tmihoc (approved):
Good catch!

PR #158, opened 2024-07-12
fix: prepend locations in .readthedocs.yaml with with docs directory

Aflynn50:
One of the locations in the RTD yaml was incorrect since we are using a docs directory.


tmihoc (approved):
Another good catch!

PR #163, opened 2024-09-17
docs: move syntax docs to RTD reference

Aflynn50:
The syntax docs were originally part of the top level doc.go. The syntax
should be moved to the read the docs because it can be better expressed
there and we can have a single source of reference.

The doc.go file can now be deleted as it is obsolete. On pkg.go.dev the
Github README is displayed there which links to the RTD docs.

Also in this PR:
docs: delete explanation section
There is nothing currently planned to go in this section. It can be
added back if needed.


tmihoc (approved):
LGTM, as discussed. Thanks!

PR #165, opened 2024-09-20
docs: SQLair tutorial updates

Aflynn50:
Update the tutorial to make it more procedural and possible to follow along.

### DIFF
--- a/docs/tutorial/tutorial.md
+++ b/docs/tutorial/tutorial.md
@@ -1,305 +1,495 @@
 (tutorial)=
-# SQLair tutorial
-SQLair is a package for Go that provides type mapping between structs/maps and SQL databases by allowing references to the Go types within the SQL query itself.
+# Get started with SQLair
 
-This tutorial will take you through the basics of writing queries with SQLair and give you a taste of what is possible.
+Imagine you are writing a Go program that handles your customers and their
+orders. The business logic is contained in the Go program but the customer and
+order data is stored in a SQL database. In this tutorial, you'll see how easy
+SQLair makes it to map between the database and the Go program.
+Let's get started!
 
-As a simple example, given Go structs `Employee` and `Team`, instead of the SQL query:
 
+## Set up
+For this tutorial you will need Go version 1.18. To use SQLair you will need and
+a Go project that sets up a database. To create the project run:
+```bash
+mkdir tutorial 
+cd tutorial
+go mod init sqlair-tutorial
 ```
-SELECT id, team_id, name
-FROM employee
-WHERE team_id = ?
-```
-
-With SQLair you could write:
-
-```
-SELECT &Employee.*
-FROM employee
-WHERE team_id = $Location.team_id
-```
-
-SQLair adds special *expressions* to your SQL statements. These expressions allow you to use Go types directly in the query to reference the source of the parameters and the destination of the results. The purpose of a query in the context of your program can be worked out from a single glance.
-
-SQLair uses the [`database/sql`](https://pkg.go.dev/database/sql) package to interact with the raw database and provides a convenience layer on top. This convenience layer provides a very different API to `database/sql`.
-
-In this introduction we will use the example of a company's database containing employees and the teams they belong to. A full example code listing is given at the end of this post.
-
-## Getting Started
-### Tagging structs
-
-To use a struct type with SQLair the first step is to tag the fields your structs with column names. The `db` tag is used to indicate which database column a struct field corresponds to.
-
-For example:
-
+To set up the database, in the tutorial folder, create a new `main.go` file,
+open it in your favourite editor and copy in the code below:
 ```go
-type Employee struct {
-	Name   string `db:"name"`
-	ID     int    `db:"id"`
-	TeamID int    `db:"team_id"`
-}
-```
+package main
 
-All struct fields that correspond to a column and are used with SQLair should be tagged with the `db` tag. If a field of the struct does not correspond to a database column it can be left untagged, and it will be ignored by SQLair.
+import (
+    // database/sql is used to open the database and set up the schema.
+    "database/sql"
+    "fmt"
 
-It is important to note that SQLair *needs* the struct fields to be public in to order read from them and write to them.
-
-### Writing the SQL
-#### SQLair Prepare
-
-A SQLair statement is built by preparing a query string. The `sqlair.Prepare` function turns a query with SQLair expressions into a `sqlair.Statement`. These statements are not tied to any database and can be created at the start of the program for use later on.
-
-*Note:* `sqlair.Prepare` does not prepare the statement on the database. This is done automatically behind the scenes when the query is built on a `DB`.
-
-```go
-stmt, err := sqlair.Prepare(`
-		SELECT &Employee.*
-		FROM person
-		WHERE team = $Manager.team_id`,
-		Employee{}, Manager{},
+    // go-sqlite3 provides the SQLite driver. Importing it registers the driver 
+    // so that sql.Open can find it.
+    // SQLair can be used with any database/driver that works with the 
+    // database/sql package.
+    _ "github.com/mattn/go-sqlite3"
 )
-```
 
-The `Prepare` function needs a sample of every struct mentioned in the query. It uses reflection information from the struct to generate the columns and check that the expressions make sense. These type samples are only used for reflection information, their content is disregarded.
+// CreateDB opens a SQLite database and adds a customer and order schema.
+func CreateDB() (*sql.DB, error) {
+    // sql.Open will create an in process, in memory database that will disappear 
+    // when the program finishes running.
+    sqlDB, err := sql.Open(
+        "sqlite3", 
+        "file:sqlair-tutorial.db?cache=shared&mode=memory",
+    )
+    if err != nil {
+        return nil, fmt.Errorf("opening SQLite database: %w", err)
+    }
 
-There is also a function `sqlair.MustPrepare` that does not return an error and panics if it encounters one.
+    // Use database/sql to create the database schema.
+    _, err = sqlDB.Exec(`
+    CREATE TABLE customers (
+        id integer PRIMARY KEY,
+        name text,
+        address text,
+        city text, 
+        postal_code text,
+        country text
+    );
 
-#### Input and output expressions
-
-In the SQLair expressions, the characters `$` and `&` are used to specify input and outputs respectively. The dollar `$` specifies a struct to fetch an argument from and the ampersand `&` specifies a struct to read query results into.
-
-##### Input expressions
-
-SQLair Input expressions replace the question mark placeholders in the SQL statement. An input expression is made up of the struct type name and a column name taken from the `db` tag of a struct field. For example:
-
-```
-UPDATE employee 
-SET name = $Empolyee.name
-WHERE id = $Employee.id
-```
-
-The expression `$Employee.name` tells SQLair that when we create the query we will give it an `Employee` struct and that the `Name` field is passed as a query argument.
-
-*Note:* we use the column name from the `db` tag, *not* the struct field name after the type.
-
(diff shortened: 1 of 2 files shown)


docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
+(tutorial)=
+# SQLair tutorial
+
+Imagine you are writing a Go program that handles your customers and their
+orders. The business logic is contained in the Go program but the customer and
+order data is stored in a SQL database. In this tutorial, we'll see how easy
+SQLair makes it to map the database between the database and the Go program.
+Let's get started!
```
tmihoc: > map the database between the database and the

Sounds wrong or confusing.

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+
+
+```{note}
+This tutorial will take around 30 minutes to complete. If you want a quicker
+start see {ref}`prepare-and-run`.
+```
```
tmihoc: If that doc really acts as a quicker start, something is wrong there. In short, I wouldn't link there and also make sure that that doc cannot act as a quickstart.

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+
+## Setting up the environment
```
tmihoc, line 16: I'd rephrase as "Set up your test environment"

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+
+To start with create a new folder called tutorial and initialise a new Go project.
```
tmihoc, line 18:
```suggestion
To start with, create a new folder called `tutorial` and initialise a new Go project.
```

tmihoc: In fact, I'd simplify even further:

```suggestion
The first thing you'll need is a Go project. Let's create one:
```

Then I'd add line by line comments in the code block, just to help someone who's really clueless catch up.

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
+(tutorial)=
+# SQLair tutorial
```
tmihoc:
```suggestion
# Get started with SQLair
```

docs/tutorial/new-tutorial.md
tmihoc:
```suggestion
## Set up your test environment
```

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+## Setting up the environment
+
+To start with create a new folder called tutorial and initialise a new Go project.
+```bash
+mkdir tutorial 
+cd tutorial
+go mod init sqlair-tutorial
+```
```
tmihoc: Not sure how to create a suggestion for multiple lines, but here's what I'd do:

```suggestion
The first thing you'll need is a Go project. Let's create one:
```bash
# Create a new directory:
mkdir tutorial 

# Enter the directory:
cd tutorial

# Initialise a new Go project:
go mod init sqlair-tutorial
```
```

PS Based on the intro of the tutorial, it sounds like what the user really needs is a Go program and a database. So I'm not sure why we're just getting them to initialise the project here.

File: docs/tutorial/new-tutorial.md [hunk 7]
```diff
@@ -0,0 +1,384 @@
...
+
+We first need to set up the backbone of our project, the database. The easiest
```
tmihoc: This tutorial "we" sounds like a good idea but it's usually either awkward (almost always) or confusing (e.g., when you need to distinguish between "here's the output I got -- yours might be similar"). I'd mostly stick to "you".

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+way to create a SQL database in Go is to use [SQLite](https://www.sqlite.org/)
+with the [Go driver from mattn](https://github.com/mattn/go-sqlite3) though
+SQLair works with any database driver that can interface with
+[database/sql](https://pkg.go.dev/database/sql). With SQLite, we can create an in
+process, in memory database that can be populated within the program and that
+will disappear when the program finishes running.
+
+We also want to set the database up with our customer and order schema. We will
+have two tables. One for customers and the other for orders.
+
```
tmihoc: This whole paragraph is too much talking at the user and it's also too much detail about something that is not at issue. I'd simplify ("You'll also need a SQL database to store your customer and order data. Let's go ahead and create one:", then in-code comments to clarify the steps, e.g., that there are many ways to create a SQL database but, since our program is in Go, we'll want Go tooling). If you still feel the need to highlight something else (e.g., that SQLair works with any database driver that can interface with... -- though I'd also wonder why we're not highlighting that bit in the homepage), put that in a note (which, if you want, you can make exciting by prefixing it with "Did you know? ...").

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+To create the database, create a new `main.go` file in the project, open it in
+your favourite editor and copy in the code below.
```
tmihoc, line 37: Again, not sure how to do multi-line suggestions, but here what's I'd do: 

```suggestion
To create the database, in your Go project create a `main.go` file; open it in your favourite editor; then copy in the code below:
```

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+    err := tutorial()
+    if err != nil {
+        fmt.Printf("ERROR: %s\n", err)
+    }
+    fmt.Println("Finished without errors!")
+}
+```
+The program above will open the SQLite database, add our schema to it, then wrap
+it as a SQLair DB. SQLair does not create the underlying database, only provide
+an interface for getting and setting the values within it.
```
tmihoc: I'd do this whole thing as below (changed information order, updated punctuation, added comment on the outcome before the code, demoted by-the-way info to a note, and fixed some typos):

```{suggestion}
To create the database, in your Go project create a `main.go` file; open it in your favourite editor; then copy in the code below. 

<code block>

This will open the SQLite database, add your schema to it, then wrap it as a SQLair DB. 

```{note}
 SQLair does not create the underlying database -- it only provides an interface for getting and setting the values within it. 
```
```

File: docs/tutorial/new-tutorial.md [hunk 11]
```diff
@@ -0,0 +1,384 @@
...
+Now, to test the program, install the dependencies and run:
+```bash
+go mod tidy
+go run .
+```
+You should get the output: `"Finished without errors!"`.
+
+# Defining the data types
+
+Before we start processing our customer and order data we need to define go
```
tmihoc: Try to avoid this sort of "Before we start..." sentences. The user doesn't gain much for them and you're just wasting their attention on meta-talk.

docs/tutorial/new-tutorial.md
Aflynn50: That doc just shows you how to prepare and run a query, which may be all they want to do. I guess its not really a quick start, just a proof of concept kind of thing. I'll remove the link.

docs/tutorial/new-tutorial.md
```diff
@@ -0,0 +1,384 @@
...
+structs that we are going to use to represent them in our program. These should
+map onto the rows of the tables in the database.
+
+Add the following types to the `main.go` file in your tutorial project.
```
tmihoc: Try to put the "where things must happen" before the "what must happen". So:

> In your project's `main.go` file add the types below:

docs/tutorial/new-tutorial.md
Aflynn50: Cool, will change that.

docs/tutorial/new-tutorial.md
Aflynn50: I've compressed that first paragraph to just mention the database and driver we are using. I think this is pretty important backgroud info since it affects the syntax of the SQL that people would write. I've moved the comment about in memory/in process into the code.

The point about it working with any driver should maybe be on the homepage, but I guess its kindof implicit.

docs/tutorial/new-tutorial.md
Aflynn50: Makes sense, I can pretty much halve the size of this paragraph and still say the same thing.

tmihoc (changes requested):
As discussed during our meeting:

(1) The tutorial should have the following main sections:

* Set things up
This should say something like "You'll need Go, a SQL database created with Go, Go structs tagged for the database [not sure about this one but it feels like a natural way to put it], and SQLair. 
* Write data to your database
* Read data from your database
The read all bit can be done in the sense of a homework challenge, with a link to the relevant reference and an expand block with what you have now as the answer.
* Tear things down
This should include deleting the project, the library, uninstalling Go (even if you only mention these things verbally or link to external stuff).

Note: I wouldn't bother with any subsections. 

(2) Something the structure suggested just now already addresses: I'd keep each section active. (That will also help avoid the temptation to slide into how-to guide or reference mode.) The sections should follow a pattern with a little bit of main text which should also highlight what needs to be done and what that will achieve + code with comments that help people understand the details. 

(3) If it needs to link to anything, the tutorial should only link in the sense of "See more" (an out-of-text, end-of-chapter "Further reading" sort of thing). Whenever it links to a different website, the link should be prefixed with that (e.g., [Go | Struct tags]...). 

I think if we make these changes the tutorial will be really strong.

tmihoc (changes requested):
Summary of our in-person chat:

(1) Make the titles Verb rather than Verb-ing based.

(2) Change the setup and teardown titles to Set things up, Tear things down.

(3) In the setup section, keep the pieces where you use sql and sqlDB (creating the SQL database) but move the part where you wrap that DB into a SQLair db to the next section where you prepare the Go types -- those two topics are in the same umbrella (prep work, but not the setup kind but rather the kind that helps you get into the SQLair logic.

(4) In the setup bit, be more explicit about the bits that matter and cut out anything that does not: "You'll need Go version ..., a Go program that does... and its associated SQL database that stores... First, ensure you have the right version of Go: .. . Next, create a directory ... and in it create a file with your Go program that also creates the SQL database: ... You're all set -- time to get cracking!

I'll keep thinking about the titles for the sections between Set things up and Tear things down. (It's a bit like `juju bootstrap` -- in the tutorial these days we don't highlight it too much; rather, what we emphasize is "this is what your deploy phase looks like".)

docs/tutorial/tutorial.md
```diff
@@ -1,305 +1,495 @@
...
+> See more: [`Query.GetAll`](https://pkg.go.dev/github.com/canonical/sqlair#Query.GetAll)
 
-If you have any more in depth questions or issues please use see our [GitHub](https://github.com/canonical/sqlair). Contributions are always welcome! Hopefully this has been of some use, enjoy using the library.
+## Cleanup
+Getting rid of the tutorial is as simple as deleting the directory! No further
+steps required.
+```bash
+cd ..
+rm -r tutorial
+```
```
tmihoc, line 490: What about uninstalling the SQLair library?

Aflynn50: Thats a good point, and as far as I can tell its surprisingly difficult. There's a stack overflow post [here](https://stackoverflow.com/questions/13792254/removing-packages-installed-with-go-get) about it, but the top answer (using `go get package@none`) no longer works because in the latest version they made `go get` only work if you're already in a project.

It seems you can do this:
```
go clean -cache -modcache
```
Which I think removes unused packages but when I ran it on my machine it came back with a weird message about linking. I'm not sure I'd want to put it in the tutorial.

In general with Go I think people just don't really worry about this sort of thing, but I agree it would be nice to have a way to do this.

tmihoc (approved):
Look as discussed, thanks!

PR #166, opened 2024-09-24
docs: update reference

Aflynn50:
Update the input and output reference to include the formal syntax under each form of expression. Also delete the redundant input/output how to guides


docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
...
 SQLair expressions can be either input expressions or output expressions. The
-input expressions specify how SQLair will extract the query arguments from the
+input expressions specify how SQLair will read the query arguments from the
 values passed to the `Query` method. Input expressions replace SQL input
 placeholders (such as `?` or `$1`).
 
 There are several different types of input expression though all use a dollar
 sign ($) in front of the types.
 
+Syntax for the different input expressions is given in [Backus Naur
```
tmihoc: The* syntax

docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
...
+form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form). 
+In the syntax definitions, the following symbols have the given meanings:
```
tmihoc: the symbols have the meaning below:*

docs/reference/output-expression-syntax.md
```diff
@@ -1,20 +1,35 @@
...
 SQLair expressions can be either input expressions or output expressions. The
-output expressions specify how SQLair will scan the query results into the
+output expressions specify how SQLair will write the query results into the
 provided structs/maps. Output expressions are designed to replace the column
 selection part of a SQL query.
 
 There are several types of output expression though all use an ampersand (`&`)
 in front of the types.
 
+Syntax for the different types of output expression is given in [Backus Naur
```
tmihoc: The* syntax

tmihoc: expressions*

docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
 # Input expression syntax
 
 SQLair expressions can be either input expressions or output expressions. The
```
tmihoc: This sentence is undermining the goal announced in the title, which was to focus on input expressions. It takes us one step back. I'd remove it / move it to a higher level that links down to input and output.

docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
...
-input expressions specify how SQLair will extract the query arguments from the
+input expressions specify how SQLair will read the query arguments from the
```
tmihoc: will read x from y into z -- spelling everything out will make things a lot clearer / faster to consume.

docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
...
+form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form). 
+In the syntax definitions, the following symbols have the given meanings:
+- `<column-name>` - Any valid SQL column name.
+- `<table-name>` - Any valid SQL table name.
+- `<struct-name>` - Any valid Golang struct name.
+- `<map-name>` - Any valid Golang map name.
+- `<slice-name>` - Any valid Golang slice name.
+
+
 ## Input a query parameter
```
tmihoc: The titles are still how-to-guide-y. Makes me wonder about the structure of the doc.

docs/reference/input-expression-syntax.md
```diff
@@ -2,25 +2,38 @@
...
-$<type-name>.<tag/key>
+```bnf
+<input> ::= "$" <type-name> "." <column-name>
+
+<type-name> ::= <struct-name> | <map-name>
+<column> ::= <column-name> | <table-name> "." <column-name>
 ```
 The struct or map can be passed to the `Query` function and the specified
 value will be extracted, and passed to the driver as a query argument.
 For example in the snippet of SQL:
```
tmihoc: This is a snippet of SQLair, not SQL, right?

tmihoc (changes requested):
Summary of our in-person chat:

(1) It's good we're getting rid of this content in the how-to guides.

(2) It's good we're getting rid of the formal BNF schema in the reference -- you're right, without definitions for each expression it's useless. Please do still make the titles nominal, to better fit the Reference style (so, go from a verbal structure to a nominal structure, e.g., from "Output a single column" to "Output expression for a single column" (I'd favor this over briefer but less explicit titles such as "Single column").

Thanks for your patience with my questions on this!

docs/reference/input-expression-syntax.md
Aflynn50: I think changing it to that would be misleading, saying "the symbols" implies these are all the symbols. These are really just the symbols that are defined outside of the grammer. 

I've changed it to: "the following symbols have the given meanings:" but I know this isn't idea either. I'm not really sure how to write it.

docs/reference/input-expression-syntax.md
Aflynn50: Changed it to match that form, though it doesn't quite "read into" anything so much as pass to the database

docs/reference/input-expression-syntax.md
Aflynn50: Its SQL with a SQLair expression, I've changed it to just say "in this snippet" to avoid confusion.

tmihoc (approved):
Looks as discussed, thanks!

PR #167, opened 2024-09-25
docs: how to query

Aflynn50:
Add an article on how to query, this replaces the prepare and run article that was here before.


docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+    City string `db:"city"`
+}
+```
+
+> See more: {ref}`types` 
+
+## Write your query string using the types
+
+SQLair queries are regular SQL queries with special input and output expressions
+inserted which allow you to reference the Go types directly from the SQL.
```
tmihoc: I'd cut out "inserted" -- unnecessary and interrupts the flow.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
+(query)=
+# How to query the database
+
+## Create/Make your types usable by SQLair
```
tmihoc: I think we agreed to number the section titles.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+The first step in building your query is to decide what you would like to put
+into/get out of the database. Identify the types in your Go program that hold
+this information or create them if they don't already exist.
+
+Make sure that the types are in the right format to work with SQLair.
+
+For example:
+```go
```
tmihoc, line 10: Here and below: The "Make sure..." and "For example" bits logically belong in the same first paragraph, so I'd put them there. But I also understand if you want more white space.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+    FROM employee
+    WHERE location_id = $Location.id
+`
+```
+
+> See more {ref}`input-expression-syntax`, {ref}`output-expression-syntax`.
+
+## Prepare your query string to get a Statement
+
+Now you have your query string, you need to pass it to `sqlair.Prepare` along
```
tmihoc, line 46: `sqlair.Prepare()` -- I suggest using the brackets to denote a function.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+with all the types mentioned in the input and output expressions. 
+
```
tmihoc, line 48: You should state the recipe explicitly, including the parameters -- ctx, query, and the types mentioned in...

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+SQLair uses the type information to generate the SQL and verify the input/output
+expressions. If there is an issue with the actual SQL in the query, this will
+not catch it. The returned `Statement` holds the parsed and verified query.
+
+
+For example:
+```go
+stmt, err := sqlair.Prepare(ctx, query, Employee{}, Location{})
```
tmihoc, line 57: `LocationID{}`?

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+if err != nil {
+    return err
+}
+```
+
+> See more [`sqlair.Preapre`](https://pkg.go.dev/github.com/canonical/sqlair#Prepare)
```
tmihoc: `sqlair.Prepare()` (with typo fixed and parentheses)

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Execute the Statement on the database
```
tmihoc: Uncapitalize "statement".

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+
+Execute the statement on a `DB` or a `TX` using the `Query` method on either.
```
tmihoc: Here and everywhere else: I'd adopt a formula that is explicit about the goal and the recipe: 

> To execute the statement on a database or a transaction, on your `DB` or `TX` object run the `Query()` method followed by the `Statement` and all the input arguments ... For example... This will return a `Query` object...

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+> See more:
+[`DB.Query`](https://pkg.go.dev/github.com/canonical/sqlair#DB.Query),
+[`TX.Query`](https://pkg.go.dev/github.com/canonical/sqlair#TX.Query),
+[`sqlair.Query`](https://pkg.go.dev/github.com/canonical/sqlair#Query)
+
+### Get one row
+To get only the first row returned from the database, use `Query.Get`. Pointers to
+all the output variables mentioned in the query should be passed to `Query.Get`.
```
tmihoc: Restate the second sentence so it's active (as in, don't tell me what should be done in the abstract, tell me what to do now -- this is a how-to guide). E.g.,

>  use `Query.Get(), passing as attribute pointers to all the output variables mentioned...`

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+
+err := db.Query(ctx, stmt, location).Get(&employee)
+if err != nil {
+    return err
+}
+
+// employee now contains the first employee returned from the database.
+```
+
+See more: [`Query.Get`](https://pkg.go.dev/github.com/canonical/sqlair#Query.Get)
```
tmihoc: Add `>` in front of See more.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Get all the rows
+To get all the rows returned from the database, use `Query.GetAll`. Pointers to
+slices of all the output variables in the query should be passed to
```
tmihoc: As before: Restate actively.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+
+err := db.Query(ctx, stmt, location).GetAll(&employees)
+if err != nil {
+    return err
+}
+
+// employees now contains all the employees returned from the database.
+```
+
+See more:
```
tmihoc: Add >

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+for iter.Next() {
+    var employee Employee
+    err := iter.Get(&employee)
+    if err != nil {
+        return err
+    }  
+}
+```
+
+See more:
```
tmihoc: Add >

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+[`Query.Iter`](https://pkg.go.dev/github.com/canonical/sqlair#Query.Iter),
```
tmihoc, line 160: Here and everywhere else: Add parentheses after the name of a method (if it makes sense).

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+    LocationID: 17
+}
+err := tx.Query(ctx, stmt, employee).Run()
+if err != nil {
+    return err
+}
+// employee has been inserted into the database.
+```
+
+See more:
```
tmihoc: Add >

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+[`Query.Run`](https://pkg.go.dev/github.com/canonical/sqlair#Query.Run)
+
+
+## (Optional) Get the query outcome
+
+The query outcome contains metadata about the execution of a query. Currently,
```
tmihoc, line 199: Move this after the recipe.

docs/howto/query.md
```diff
@@ -0,0 +1,227 @@
...
+it only contains a `sql.Result` object which contains information returned from
+the driver such as the number of rows affected. It may contain more in the
+future.
+
+To get the outcome, a pointer to a `sqlair.Outcome` object can be provided as
```
tmihoc: "can be" -- Can I get the outcome in any other way? (Don't tell me what can happen -- just give me the recipe.)

docs/howto/types.md
```diff
@@ -1,13 +1,12 @@
...
 supports inputs from structs maps and slices and can output into structs
 and maps. 
 
-In general, all types used with SQLair need to be named, otherwise there is no
-way to reference them in the query. The rest of the preparation of the type
-depends on what kind it is.
+All types used with SQLair need to be named so that they can be referenced in
+the query.
 
 ## How use a struct with SQLair
```
tmihoc: How to*

docs/howto/types.md
```diff
@@ -34,6 +33,25 @@ or output expression.
 If a struct contains an embedded struct then SQLair will treat the fields of the
 embedded structs as if they were fields in the parent struct.
 
+### The "omitempty" tag
```
tmihoc: The title doesn't fit in a how-to. Also, in the main text we call `omitempty` a keyword -- the tag is `db`. This also makes me wonder: Are there any other special keywords or is this it? Basically, what's the tl;dr on reserved names here? Is it `db` for tags and `omitempty` for keywords, and that's it?

tmihoc (changes requested):
Looking great but left some comments.

docs/howto/query.md
Aflynn50: I'm not too sure about this, I think in python people do that a lot, but in Go less so. To me that syntax would specifically imply it was a function with no arguments that returned nothing. I could list out what its arguments and return types but then it would be too long, and that belongs in reference anyway.

docs/howto/query.md
Aflynn50: For prepare, its just the query string and the types. The context is just for the actual `Query` method because that calls through to the DB so may take longer and need cancelling.

docs/howto/query.md
Aflynn50: Just `Location` here, its the name of the struct defined at the top.

docs/howto/query.md
Aflynn50: Sounds good, though I didn't want to put an example here because it should always be run immediately with one of the methods listed below. Putting an example just doing `q := db.Query(ctx, stmt, args...)` might encourage the wrong thing.

docs/howto/query.md
Aflynn50: Changed to be more active

docs/howto/types.md
Aflynn50: On reflection I think this doc should be turned into a reference. If you wanted to use SQLair you would just want to look up how to format your specific type rather than be guided through all the possibilities.

If its ok with you I think I will move this to reference and make it more reference like.

Aflynn50: I've moved it over, let me know what you think

docs/howto/query.md
tmihoc: OK. By all means, let's do what's standard in Go.

docs/howto/query.md
```diff
@@ -0,0 +1,225 @@
...
+if err != nil {
+    return err
+}
+```
+
+> See more [`sqlair.Prepare`](https://pkg.go.dev/github.com/canonical/sqlair#Prepare)
+
+## 4. Execute the statement on the database
+
+Execute the statement on a `DB` or a `TX` using the `Query` method passing it
```
tmihoc: To execute the statement on a..., use the Query method, passing as parameters...

docs/howto/query.md
```diff
@@ -0,0 +1,225 @@
...
+[`Query.Run`](https://pkg.go.dev/github.com/canonical/sqlair#Query.Run)
+
+
+## 5. (Optional) Get the query outcome
+
+To get the outcome, provide a pointer to a `sqlair.Outcome` as the first
+argument to any of the `Get` methods. This will fill the `sqlair.Outcome` with
+information about the outcome of the query.
+
+The query outcome contains metadata about the execution of a query. Currently,
```
tmihoc, line 199: This feels like material for a note (BTW sort of content).

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
+(types)=
```
tmihoc, line 1: Moving this to reference feels like a great idea.

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
...
+# Type format
+
+SQLair supports structs maps and slices in input expressions and structs and
```
tmihoc: structs,* maps,* and

tmihoc: I'd also structure in two bullet points:

SQLair supports the following types:

- in input expressions:
    - structs 
    - maps
    - slices
 - in output expressions:
    - structs
    - maps

The nested lists are optional (as in, you can give them in-line in the host list item).

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
...
+maps in output expressions. 
+
+All types used with SQLair must be named so they can be referenced in the query.
+
+## Structs
+
+To use a struct with SQLair, all struct fields that correspond to a database
```
tmihoc: To be usable by SQLair*

(Adjusting the style to something more reference-y.)

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
...
+    DoB     string  `db:"date_of_birth"`
+    Weight  int
+    height  int
+}
+```
+In this example the `Weight` and `height` fields will be ignored by SQLair.
+
+### The "omitempty" keyword
+
+The `omitempty` keyword tells SQLair to omit a column specified in a struct tag
```
tmihoc: I'd start with "In a struct tag, ..." -- a zoom-in approach generally makes things easier to parse (you start with something people already know, namely the struct and its tag, and then continue with the new info about this specific struct tag keyword).

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+### sqlair.M
+For convenience, SQLair provides a named map type
+[`sqlair.M`](https://pkg.go.dev/github.com/canonical/sqlair#M) which has the
+type `map[string]any`.
+
+## Slices
+
+Named slices can only be used as inputs and will expand into a comma separated
```
tmihoc, line 78: comma-separated* (I know some people don't dash on purpose, e.g., in physics journals where doing that would end up in dashes all over the place, but here I'd say it's not too much and feels appropriate).

docs/reference/types.md
tmihoc: I'd retitle to "Supported Go types and their format" and structure into:

## Supported Go types

SQLair supports the following types:

- in [input expressions]:
    - structs 
    - maps
    - slices
 - in [output expressions]:
    - structs
    - maps

[Note: If you want, you can link structs, maps, and slices to their Go definition too. To indicate to the user that the link is taking them outside of the present website, you can use a little caret, e.g., [structs^]...]

## Type format 

[everything you have now -- just nest the sections]

docs/reference/types.md
```diff
@@ -0,0 +1,81 @@
...
+    ID      int     `db:"id_number"`
+    DoB     string  `db:"date_of_birth"`
+    Weight  int
+    height  int
+}
+```
+In this example the `Weight` and `height` fields will be ignored by SQLair.
+
+### The "omitempty" keyword
+
```
tmihoc, line 40: Are there any other special keywords?

docs/howto/query.md
Aflynn50: Nice, good point. Notes always make this sort of thing look more professional too

docs/reference/types.md
Aflynn50: No, its just that one, should I mention that its the only one?

docs/reference/types.md
Aflynn50: Nice, done that

tmihoc (approved):

PR #192, opened 2026-02-06
fix: implements a behaviour change to zero slice values

tlm:
This PR changes the contract that GetAll has with the caller around how slices are populated for queries. Prior to this change GetAll would append all query values to the slice supplied resulting in existing slice elements being retained.

This poses a problem when failed transactions are retried because should a call have previously succeeded but the transaction fails for some other reason then the slice will get the same values again or at least the changed subset.

I am unsure if this behaviour was on purpose in the original implementation but it does fight the standard SQL package where values are always overwritten in a query.

If a previous caller to this func was relying on this behaviour then it makes sense that an AppendAll variant could also be provided.

Due to the change in behaviour of this commit it MUST be rolled out in a version bump of Sqlair. I have also added regression testing to the package to catch this going forward.

### DIFF
--- a/sqlair.go
+++ b/sqlair.go
@@ -317,6 +317,9 @@ func (o *Outcome) Result() sql.Result {
 // A pointer to an empty [Outcome] struct may be provided as the first output
 // variable to get information about query execution.
 //
+// A provided slice that already contains one or more values will have its
+// length reset to 0 before scanning into the slice.
+//
 // [ErrNoRows] will be returned if no rows are found.
 func (q *Query) GetAll(sliceArgs ...any) (err error) {
 	if q.err != nil {
@@ -348,6 +351,9 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 		if sliceVal.Kind() != reflect.Slice {
 			return fmt.Errorf("need pointer to slice, got pointer to %s", sliceVal.Kind())
 		}
+		// Set the length of the slice value back to zero maintaing any existing
+		// allocated capacity.
+		sliceVal.SetLen(0)
 		sliceVals = append(sliceVals, sliceVal)
 	}
 
--- a/sqlair.go
+++ b/sqlair.go
@@ -351,8 +351,8 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 		if sliceVal.Kind() != reflect.Slice {
 			return fmt.Errorf("need pointer to slice, got pointer to %s", sliceVal.Kind())
 		}
-		// Set the length of the slice value back to zero maintaing any existing
-		// allocated capacity.
+		// Set the length of the slice value back to zero maintaining any
+		// existing capacity.
 		sliceVal.SetLen(0)
 		sliceVals = append(sliceVals, sliceVal)
 	}
--- a/docs/howto/tx.md
+++ b/docs/howto/tx.md
@@ -54,5 +54,5 @@ if err != nil {
 :class: tip
 [TX.Commit](https://pkg.go.dev/github.com/canonical/sqlair#TX.Commit),
 [TX.Rollback](https://pkg.go.dev/github.com/canonical/sqlair#TX.Rollback),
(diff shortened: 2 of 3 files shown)


tlm:
It may be slightly better to set the length of the slice at the end of the func when we assign back to it. That way we can say for any failure case the existing slice is unmodified. A better contract I feel.

SimoneDutto:
Hello, sorry to jump in but i thought it was an interesting pr to review.

Wouldn't it be better to just return an error if the slice is not empty?
```
if v.Len() > 0 {
	fmt.Println("you shouldn't do this")
}
```

So people upgrading to a newer version of sqlair will immediately find the places where they were using (willingly or not) this behavior.

gfouillet:
> Hello, sorry to jump in but i thought it was an interesting pr to review.
> 
> Wouldn't it be better to just return an error if the slice is not empty?
> 
> ```
> if v.Len() > 0 {
> 	fmt.Println("you shouldn't do this")
> }
> ```
> 
> So people upgrading to a newer version of sqlair will immediately find the places where they were using (willingly or not) this behavior.

The issue with this approach is that the error will be triggered in some case of transaction failure, so it would be quite undeterministic. Instead of having inconsistant result you will have inconstant failure, which is not better :smile_cat: 

To me, resetting the slice length is the right approach, both in term of memory consumption and correctness.

sqlair.go
```diff
@@ -348,6 +351,9 @@ func (q *Query) GetAll(sliceArgs ...any) (err error) {
 		if sliceVal.Kind() != reflect.Slice {
 			return fmt.Errorf("need pointer to slice, got pointer to %s", sliceVal.Kind())
 		}
+		// Set the length of the slice value back to zero maintaing any existing
```
gfouillet: **todo:**
```suggestion
		// Set the length of the slice value back to zero maintain any existing
```

gfouillet: **thought:** I am not quite sure the comment is right out of context though.

The fact that setting the length to zero may reduce the need of an allocation between two retry, it is not the *intent* of this line.

The intent is to flush any result from previous retries to be pulled in next one.

gfouillet (commented):
Great spot, thanks.

I believe we should have two more tests cases:

* A case where the dbVals in the second query is longer than the expected result (`dbVals = append(dbVals, Person{})` before second call
* A case where the dbVals in the second query is shorter than the expected result (`dbVals = dbVals[1:]`). For this one it would be worthy to also check the content of the retrieved values.

Both test case may help to find edge case in term of reusing memory slot (even if there is no bug in the implementation at the moment)

sqlair.go
manadart: "maintaining" is the correct word here.

manadart (approved):
Thanks for this. There is already minor tweak that I landed, which is yet to be rolled out. We'll bump and get a new version in to the 4.0 branch.

tlm:
@gfouillet interesting approach by introducing the error. I do feel that introducing an error is still a behaviour change and requires a big version bump. If we are travelling down that road may as well just deal with it.

sqlair.go
tlm: Fixed up.

tlm:
> Hello, sorry to jump in but i thought it was an interesting pr to review.
> 
> Wouldn't it be better to just return an error if the slice is not empty?
> 
> ```
> if v.Len() > 0 {
> 	fmt.Println("you shouldn't do this")
> }
> ```
> 
> So people upgrading to a newer version of sqlair will immediately find the places where they were using (willingly or not) this behavior.

As per the PR description this change needs to occur in a major version bump because of a behaviour change. Your approach is interesting but it is also a behaviour change resulting in the same version bump. @gfouillet comments then come into play.

If there is legitimate use case for wanting the behaviour we can add an 1AppendAll1 version in a minor bump to the lib.

tlm:
@manadart performed a drive by docs fix for CI. Just fixing a broken link for ErrTxDone to now point to stdlib as the errro was never defined in this package. See last commit in PR.

